PR #3, opened 2019-11-14
Removing Picard MarkDuplicates step

adthrasher:
Addresses feedback on the RFC from Kyle regarding duplicate marking. Removing that step and updating the downstream steps to use the STAR aligned BAM directly.


claymcleod:
@adthrasher see last two commits

PR #5, opened 2019-11-19
A little winter cleaning

claymcleod:
This pull serves a few purposes:

* Adds a `bootstrap-reference.wdl` workflow to create the reference files we use. This may need to be refined later down the line.
* Creates subfolders for `standard-rnaseq` and `reference` since we now have n=2 workflow categories.
* Update all underlying workflows to reflect the change.


claymcleod:
@jsunny23 @adthrasher I'm going to go ahead and merge in. Jobin has some updates to apply on top of this for outputting a STARDB zip, which is the only improvement remaining.

PR #7, opened 2020-01-06
Documentation

adthrasher:
Adding WDL-style documentation in the parameter_meta and meta fields.


adthrasher:
@claymcleod - This is the WDL style for adding documentation. Though there doesn't seem to be good support for the multi-author case. The meta fields don't support an array value.

claymcleod:
Looks good to me.

PR #9, opened 2020-01-22
Sorting

adthrasher:
Moving the BAM sort step out of STAR and implement using Picard's SortSam. This should reduce the memory requirement for STAR and also the overall cost as a high memory machine won't be necessary for the entire length of the alignment+sort.


claymcleod:
This appears to have passed validation, so I'm going to merge in so I can run some tests tonight.

PR #10, opened 2020-01-24
improve: infer strandedness for counts

claymcleod:


claymcleod:
@adthrasher @jsunny23 FYI.

PR #11, opened 2020-02-05
Add memory arg to picard.validate_bam

a-frantz:


claymcleod:
Oh! Actually, you need to set the `memory` meta var.

PR #12, opened 2020-02-06
feat: add a `miniwdl check` Github Action

a-frantz:


a-frantz:
The failing check is due to syntax error are already present in the repo. `rnaseq-star-db-build.wdl` imports `tools/star.wdl` twice. This current implementation will also fail with files that contain whitespace, but figure that's safe.

.github/workflows/main.yml
```diff
@@ -0,0 +1,26 @@
+name: CI
+
+on: [push]
+
+jobs:
+  check:
```
claymcleod: Can we rename this to "miniwdl check"?

PR #13, opened 2020-02-07
feat: add import_syntax_check to Github Actions

a-frantz:


a-frantz:
Alright the test is working now. Do we want to add a badge?

.github/workflows/main.yml
```diff
@@ -23,4 +23,17 @@ jobs:
...
+        done
+  
+  import_syntax_check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Check import syntax
+      run: |
+        files=$(find . -name '*.wdl')
+        bad_lines=$(awk '/import/' $files | awk '!/master/')
```
claymcleod: I think you probably want to search for the full URL string (not just `master`), just to be safe.

PR #16, opened 2020-02-24
Adthrasher/dnanexus changes

adthrasher:
Making changes to rna-seq v2 to work with DNAnexus. 
Adding interactive t-SNE workflow.

### DIFF
--- a/tools/ngsderive.wdl
+++ b/tools/ngsderive.wdl
@@ -37,3 +37,33 @@ task infer_strand {
         File strandedness_file = stdout()
     }
 }
+
+task read_length {
+    input {
+        File bam
+        File bai
+        Int max_retries = 1
+        Int memory_gb = 5
+    }
+
+    Float bam_size = size(bam, "GiB")
+    Int disk_size = ceil(((bam_size) * 2) + 10)
+ 
+    command {
+        mv ~{bai} ~{bam}.bai 
+        ngsderive readlen ~{bam} | awk 'NR > 1' | cut -d$'\t' -f5
+    }
+
+    runtime {
+        disk: disk_size + " GB"
+        memory: memory_gb + " GB"
+        docker: 'stjudecloud/bioinformatics-base:bleeding-edge'
+        maxRetries: max_retries
+    }
+
+    output {
+        String read_length = read_string(stdout())
+        File read_length_file = stdout()
+    }
+
+}
--- a/tools/ngsderive.wdl
+++ b/tools/ngsderive.wdl
@@ -12,6 +12,7 @@ task infer_strand {
         File bai
         File gtf
         Int max_retries = 1
+	Int memory_gb = 5
     }
 
     Float bam_size = size(bam, "GiB")
@@ -26,6 +27,7 @@ task infer_strand {
 
     runtime {
         disk: disk_size + " GB"
+        memory: memory_gb + " GB"
         docker: 'stjudecloud/bioinformatics-base:bleeding-edge'
         maxRetries: max_retries
     }
--- a/tools/ngsderive.wdl
+++ b/tools/ngsderive.wdl
@@ -12,7 +12,7 @@ task infer_strand {
         File bai
         File gtf
         Int max_retries = 1
-	Int memory_gb = 5
+        Int memory_gb = 5
     }
 
     Float bam_size = size(bam, "GiB")
--- a/tools/tar.wdl
+++ b/tools/tar.wdl
@@ -1,0 +1,50 @@
+## Description: 
+##
+## This WDL tool wraps any functionality related to tar.
+
+version 1.0
+
+task tar_print_version {
+    command {
+        tar --version
+    }
+
+    runtime {
+        docker: 'stjudecloud/bioinformatics-base:bleeding-edge'
+    }
+
+    output {
+        String out = read_string(stdout())
+    }
+}
+
+task untar {
+    input {
+        File infile
+        Int max_retries = 1
+    }
+
+    runtime {
+        docker: 'stjudecloud/bioinformatics-base:bleeding-edge'
+        maxRetries: max_retries
+    }
+
+    command {
+        mkdir tar_output
+        tar  --no-same-owner -xf ${infile} -C tar_output
+    }
+
+    output {
+        Array[File] outfiles = glob("tar_output/*")
+    }
+
+    meta {
+        author: "Andrew Thrasher"
+        email: "andrew.thrasher@stjude.org"
+        description: "This WDL tool unzips a tar archive" 
+    }
+
+    parameter_meta {
+        infile: "Archive in tar format to be extracted"
+    }
+}
--- a/tools/tsne.wdl
+++ b/tools/tsne.wdl
@@ -1,0 +1,97 @@
+version 1.0
+
+task plot {
+    input {
+        Array[File] counts
+        Array[String]? inputs
+        Array[File]? input_counts
+        File blacklist
+        File covariates
+        File gencode_gtf
+        String outfile = "tsne.html"
+        Int max_retries = 1
+        Int memory_gb = 5
+        Int disk_size = 15
+    }
+
+    command {
+        itsne-main -b ${blacklist} -g ${gencode_gtf} -c ${covariates} -o ${outfile} \
+            ${true='--input-sample ' false='' defined(inputs)}${sep='--input-sample ' inputs} \
+            ${sep=' ' counts} ${sep=' ' input_counts}
+    }
+  
+
+    runtime {
+        disk: disk_size + " GB"
+        memory: memory_gb + " GB"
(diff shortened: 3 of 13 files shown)


tools/ngsderive.wdl
```diff
@@ -35,3 +37,33 @@ task infer_strand {
...
+        Int max_retries = 1
+        Int memory_gb = 5
+    }
+
+    Float bam_size = size(bam, "GiB")
+    Int disk_size = ceil(((bam_size) * 2) + 10)
+ 
+    command {
+        mv ~{bai} ~{bam}.bai 
+        ngsderive readlen ~{bam} | awk 'NR > 1' | cut -d$'\t' -f5
```
jsunny23: One thing that Andrew F. updated on the QC branch was parsing the output from ngsderive in a way that doesn't fail if other things are written to stdout. For example, the cluster may print job information on stdout and this gets read by the workflow and will cause a failure if it tries to use that output.

One way is to write the output of ngsderive to a file and then use that file instead of stdout(). 
```suggestion
        ngsderive readlen ~{bam} | awk 'NR > 1' | cut -d$'\t' -f5 > outfile.txt
```

tools/tar.wdl
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    runtime {
+        docker: 'stjudecloud/bioinformatics-base:bleeding-edge'
+        maxRetries: max_retries
+    }
+
+    command {
+        mkdir tar_output
+        tar  --no-same-owner -xf ${infile} -C tar_output
```
jsunny23:
```suggestion
        tar --no-same-owner -xf ${infile} -C tar_output
```

tools/tar.wdl
```diff
@@ -0,0 +1,50 @@
...
+
+version 1.0
+
+task tar_print_version {
+    command {
+        tar --version
+    }
+
+    runtime {
+        docker: 'stjudecloud/bioinformatics-base:bleeding-edge'
```
jsunny23: Just a suggestion that we could probably just use ubuntu 18.04 images for these tasks which is what `bioinformatics-base` is based on. 

```suggestion
        docker: 'ubuntu:18.04'
```

tools/util.wdl
```diff
@@ -68,3 +68,44 @@ task prepare_read_groups_for_star {
...
+        String? suffix
+        Int max_retries = 1
+    }
+
+    command {
+        basename ${in_file} "${suffix}" > stdout.txt
+    }
+
+    runtime {
+        docker: 'stjudecloud/bioinformatics-base:bleeding-edge'
```
jsunny23:
```suggestion
        docker: 'ubuntu:18.04'
```

workflows/interactive-tsne/interactive-tsne.wdl
```diff
@@ -0,0 +1,123 @@
...
+##
+## THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
+## BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+## NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+## DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+## OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+version 1.0
+
+import "../rnaseq/rnaseq-standard.wdl" as rnav2
```
jsunny23: Should be absolute imports - that will also fix some of the errors on `miniwdl check`

tools/util.wdl
```diff
@@ -68,3 +68,44 @@ task prepare_read_groups_for_star {
         read_groups: "The read group portion of a BAM header as a string"
     }
 }
+
+task file_basename {
```
claymcleod: I almost feel like this isn't worth making its own task. I didn't look where it is used, but it it possible to just call this inline with the other commands in tasks where it is needed?

tools/tsne.wdl
```diff
@@ -0,0 +1,116 @@
...
+        File html = outfile
+    }
+
+    meta {
+        author: "Andrew Thrasher"
+        email: "andrew.thrasher@stjude.org"
+    }
+
+    parameter_meta {
+        covariates: "A tab delimited file with Sample, Diagnosis, Strandedness, LibraryType, ReadLength. Valid values for strandedness: [Stranded, Unstranded]. Valid values for LibraryType: [mRNA, totalRNA]. ReadLength should be specified similar to the form PE100bp."
```
claymcleod: Maybe point out "SE" or "PE", whatever is expected for single-end specifically?

tools/tsne.wdl
```diff
@@ -0,0 +1,116 @@
...
+    }
+}
+
+task validate_tissue_type {
```
claymcleod: Worth it being its own task or can be combined in another to save requisitioning a node?

tools/util.wdl
```diff
@@ -68,3 +68,44 @@ task prepare_read_groups_for_star {
...
+        basename ${in_file} "${suffix}" > stdout.txt
+    }
+
+    runtime {
+        docker: 'stjudecloud/bioinformatics-base:bleeding-edge'
+        maxRetries: max_retries
+    }
+
+    output { 
+        String out = read_string("stdout.txt")
```
claymcleod: Potentially,

```suggestion
        String out = read_string(stdout())
```

combined with not writing to file?

tools/util.wdl
```diff
@@ -68,3 +68,44 @@ task prepare_read_groups_for_star {
...
+        basename ~{in_file} | awk -F '.' '{print $1}' > stdout.txt
+    >>>
+
+    runtime {
+        docker: 'stjudecloud/bioinformatics-base:bleeding-edge'
+        maxRetries: max_retries
+    }
+
+    output { 
+        String out = read_string("stdout.txt")
```
claymcleod: Same thought as above for not writing to file, just calling `stdout()` directly.

workflows/interactive-tsne/interactive-tsne.wdl
```diff
@@ -0,0 +1,123 @@
...
+        Array[File] in_bams
+        File gencode_gtf
+        File? stardb_tar_gz
+        File? reference_counts
+        File? covariates_file
+        String tissue_type
+        String? output_filename
+    }
+    
+    String blacklist_url = "https://athrashe.blob.core.windows.net/tsne-test/gene.blacklist.tsv"
```
claymcleod: We should create an official St. Jude Cloud reference files bucket for this. Also, does it make sense to pull this out into the inputs? Not sure.

workflows/interactive-tsne/interactive-tsne.wdl
```diff
@@ -0,0 +1,123 @@
...
+    String outfile = select_first([output_filename, "output.html"])
+    
+    if (! defined(stardb_tar_gz)){
+        String star_url = "https://athrashe.blob.core.windows.net/tsne-test/STARDB.tar.gz"
```
claymcleod: Same here, should be official St. Jude Cloud reference files. And I think we might should version them so: `https://stjudecloud.blob.core.windows.net/reference/RNA-seq/v2/STARDB.tar.gz`

tools/tsne.wdl
adthrasher: Ultimately, I think it only matters that the labels are consistent. The underlying code doesn't do any parsing of the field, it just uses them as categorical labels and does batch correction. The python code just combines the labels into a single label that the R code then uses directly. `    self.df["Covariates"] = self.df["Strandedness"] + "_" + self.df[                                                                                                                 
        "LibraryType"] + "_" + self.df["ReadLength"]     `

tools/tsne.wdl
adthrasher: The natural place would be to check it in the plotting section, but I want to avoid running rna-seq v2 for the input samples if the wrong tissue type is specified.

tools/util.wdl
adthrasher: This was in response to the issue we were discussing with LSF. Jobin suggested this change above and I do think it is worth being explicit about redirecting the output, because it is easy to imagine another backend specification that writes batch system output to the stdout and combines them.

workflows/interactive-tsne/interactive-tsne.wdl
adthrasher: Agreed, there should be an official bucket. One issue with inputs is the cluttered display in DNAnexus. It is already fairly busy.

PR #17, opened 2020-02-25
qc workflow

a-frantz:


claymcleod:
Once you rename the base container to `stjudecloud/conda-base`, it will fix another issue I see, which is that currently the `sjcloud-base` docker image would need to be built locally rather than it being fetchable from Docker hub.

Which raises the question, should we try publishing these using Github Packages? I think that could be cool to try here, especially since the long docker names required by using GH packages don't have to been seen by the end user.

a-frantz:
I thought that was always the plan, but I wasn't aware of Github packages. I thought the only way to avoid building locally was to publish to Dockerhub.

However these still haven't been tested in the workflow. There might be bugs we haven't caught yet. (I'm specifically worried about predefining entrypoints. Not sure how WDL will handle that.)

My thinking is we first test with GH packages, make sure there aren't any kinks, and eventually publish to Dockerhub. I see these tools as potentially being useful outside the context of this repo. Simple containerization of individual tools. Like the initial goal of bioinformatics-base (which I'm planning on debugging. Shouldn't be hard to get `ngsderive` installed correctly.)

claymcleod:
Yeah I agree with those points. Also worth considering though is the clean experience of them showing up in this repo since they _are_ primarily written for this. I think it could be a really interesting thing. 

It occurs to me that it might be as easy as syncing them with docker hub using the CI as well 🤷‍♂

claymcleod:
Would be good to make the output file from `ngsderive strandedness` look _exactly_ as it does without postprocessing (just like I did for `readlen` and `instrument`). This is because a future version of MultiQC is going to parse these directly (https://github.com/ewels/MultiQC/pull/1104). I think it would still be good to postprocess the file into a distilled value in `String strandedness` as that String is used in the RNA-seq pipeline when a strandedness is not provided.

Also, I think it would be good if we could throw an error if the derived strandedness did not equal the provided strandedness.

docker/fastqc/1.0.0-alpha/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM stjudecloud/conda-base:1.0.0-alpha AS builder
+
+RUN conda create -n fastqc \
+        fastqc==0.11.8 \
+        -y \
+    && conda clean --all
+
+FROM debian:10-slim
+COPY --from=builder /opt/conda/envs/fastqc/bin/ /opt/conda/envs/fastqc/bin/
```
claymcleod: Could this be one `COPY` for all of the `fastqc` dir to make the Dockerfile cleaner?

docker/qualimap/1.0.0-alpha/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM stjudecloud/conda-base:1.0.0-alpha AS builder
+
+RUN conda create -n qualimap \
+        qualimap==2.2.2c \
+        -y \
+    && conda clean --all
+
+FROM debian:10-slim
+COPY --from=builder /opt/conda/envs/qualimap/bin/ /opt/conda/envs/qualimap/bin/
```
claymcleod: Same thought here.

docker/fastqc/1.0.0-alpha/Dockerfile
a-frantz: There a bunch of directories in there that aren't needed. I haven't checked their size, so might not actually impact the final docker image size significantly. But this is the most space efficient method.

claymcleod: Okay, sounds okay to me then!

docker/fastq_screen/1.0.0-alpha/fastq_screen.conf
```diff
@@ -0,0 +1,127 @@
+# This is a configuration file for fastq_screen
```
claymcleod: Can we clean up this file by removing all of the comments?

a-frantz:
@claymcleod @jsunny23 can you guys take a look at my changes to `rnaseq-standard`? I removed all the steps that are in the QC workflow, but I want to check that the split makes sense. I also plan on writing another workflow which just wraps standard and QC.

claymcleod:
Hm, I know this is what we discussed, but I really don't love the way c2771a4 lands. May take some more thought as to how to cleanly do this but I don't have ideas atm.

a-frantz:
I'll think about it some more, but this was the most obvious way to:
1) make `parse_input` a lightweight task
2) prevent any tasks from *requiring* a wait signal
3) imo, I think it's readable and obvious why we're doing it this way, which my previous solution did not do. Could use some comments to make it explicitly clear if it's not clear enough as is.

tools/fastq_screen.wdl
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    output {
+        File db = tar_filename
+    }
+
+    meta {
+        author: "Clay McLeod"
+        email: "clay.mcleod@STJUDE.org"
+        description: ""
```
claymcleod: Could you please add your name and fill in this description?

tools/fastq_screen.wdl
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    parameter_meta {}
```
claymcleod: Delete or add details.

tools/fastq_screen.wdl
```diff
@@ -0,0 +1,73 @@
...
+
+    String output_basename = basename(read1, ".fastq") + "_screen"
+    String db_name = basename(db)
+
+    command {
+        cp ~{db} /tmp
+        tar -xsf /tmp/~{db_name} -C /tmp/
+
+        format_arg=''
+        if [ "~{format}" = "illumina1.3" ]; then
```
claymcleod:
```suggestion
        if [[ "~{format}" = "illumina1.3" ]]; then
```

tools/fastqc.wdl
```diff
@@ -12,28 +12,29 @@ task fastqc {
         String prefix = basename(bam, ".bam")
         Int memory_gb = 4
         Int max_retries = 1
+        String wait_var = ""
```
claymcleod: Definitely don't like this. Can't think of a better option at the moment though.

workflows/qc/quality-check-standard.wdl
```diff
@@ -0,0 +1,155 @@
...
+        String input_fq_format
+    }
+
+    command {
+        if [ "~{input_experiment}" != "WGS" ] && [ "~{input_experiment}" != "WES" ] && [ "~{input_experiment}" != "RNA" ]; then
+            >&2 echo "experiment input must be 'WGS', 'WES', or 'RNA'"
+            exit 1
+        fi
+
+        if [ -n "~{input_strand}" ] && [ "~{input_strand}" != "stranded-reverse" ] && [ "~{input_strand}" != "stranded-forward" ] && [ "~{input_strand}" != "unstranded" ]; then
```
claymcleod: Let's make all of these capitalized (`Stranded-Reverse`) to mirror what we do with `ngsderive`.

tools/htseq.wdl
```diff
@@ -9,17 +9,17 @@ task count {
     input {
         File bam
         File gtf
-        String strand
-        String inferred
+        String provided_strand
```
claymcleod: Nice 👍

tools/multiqc.wdl
```diff
@@ -9,47 +9,58 @@ task multiqc {
...
     command {
-        echo ${sorted_bam} > file_list.txt
-        echo ${validate_sam_string} > validate_sam.txt
+        echo ~{sorted_bam} > file_list.txt
+        echo ~{validate_sam_string} > validate_sam.txt
         echo validate_sam.txt >> file_list.txt
 
-        for file in ${sep=' ' qualimap_bamqc} ; do
+        mkdir qualimap_bamqc/
+        tar -xzf ~{qualimap_bamqc} --strip-components=1 -C qualimap_bamqc/;
```
claymcleod: I think this still will present problems in the multiqc report where it shows as two line items in the summary table.

tools/ngsderive.wdl
```diff
@@ -14,24 +14,78 @@ task infer_strand {
...
         disk: disk_size + " GB"
-        docker: 'stjudecloud/bioinformatics-base:bleeding-edge'
+        docker: 'stjudecloud/ngsderive:1.0.0-alpha'
         maxRetries: max_retries
     }
 
     output {
-        String strandedness = read_string(stdout())
-        File strandedness_file = stdout()
+        String strandedness = read_string("strandedness.txt")
```
claymcleod: Have we tried an RNA-seq sample without a provided strandedness to make sure it works? Please ensure you do before you resolve this comment.

tools/qc.wdl
```diff
@@ -49,23 +49,23 @@ task parse_infer_experiment {
     }
```
claymcleod: Can we look at this and maybe remove any of the references to `rseqc` tools since we don't use that anymore? There may also be a docker container, not sure. This doesn't have to happen for this PR, just something I don't want to forget to clean up.

tools/qualimap.wdl
```diff
@@ -52,45 +56,53 @@ task rnaseq {
...
     }
 
-    String stranded = if (strand != "") then 
-                        if (strand == "reverse") then "strand-specific-reverse" else
-                        if (strand == "yes") then "strand-specific-forward" else
-                        if (strand == "no") then "non-strand-specific"
+    String out_directory = select_first([prefix, "qualimap_rnaseq_results"])
+    String out_tar_gz_file = out_directory + ".tar.gz"
+    String stranded = if (provided_strand != "") then 
+                        if (provided_strand == "stranded-reverse") then "strand-specific-reverse" else
```
claymcleod: Make sure all of these use caps (`Stranded-Reverse`)

tools/qualimap.wdl
```diff
@@ -52,45 +56,53 @@ task rnaseq {
...
     Int disk_size = select_first([disk_size_gb, ceil(((bam_size + gencode_gtf_size) * 12) + 10)])
  
     command {
-        qualimap rnaseq -bam ${bam} -gtf ${gencode_gtf} -outdir ${outdir} -oc qualimap_counts.txt -p ${stranded} -pe --java-mem-size=${java_heap_size}G
+        qualimap rnaseq -bam ~{bam} \
+                        -gtf ~{gencode_gtf} \
+                        -outdir ~{out_directory} \
+                        -oc qualimap_counts.txt \
+                        -p ~{stranded} \
+                        -pe \
```
claymcleod: I think we need to make this a global workflow option, some RNA-seq data is single end. But we can default ours all to `true`.

tools/qualimap.wdl
```diff
@@ -102,6 +114,6 @@ task rnaseq {
     parameter_meta {
         bam: "Input BAM format file to generate coverage for"
         gencode_gtf: "A GTF format features file containing Gencode features"
-        strand: "Strand information for RNA-seq experiments. Options: [strand-specific-forward, strand-specific-reverse, non-strand-specific]"
+        provided_strand: "Strand information for RNA-seq experiments. Options: [strand-specific-forward, strand-specific-reverse, non-strand-specific]"
```
claymcleod: Capitalize.

tools/rseqc.wdl
```diff
@@ -12,19 +12,20 @@ task infer_experiment {
         Int? map_qual
```
claymcleod: Likely can delete this whole file.

tools/samtools.wdl
```diff
@@ -95,23 +97,59 @@ task split {
...
+        File bam
+        Int max_retries = 1
+        String wait_var = ""
+    }
+
+    Float bam_size = size(bam, "GiB")
+    Int disk_size = ceil((bam_size * 2) + 10)
+
+    command {
+        samtools view -H ~{bam} | grep "@RG" > stdout.txt
```
claymcleod:
```suggestion
        samtools view -H ~{bam} | grep "^@RG" > stdout.txt
```

tools/samtools.wdl
```diff
@@ -95,23 +97,59 @@ task split {
...
+    }
+
+    meta {
+        author: "Andrew Thrasher, Andrew Frantz"
+        email: "andrew.thrasher@stjude.org, andrew.frantz@stjude.org"
+        description: "This WDL tool is a utility to get read group information from a BAM file and write it out to as a string" 
+    }
+
+    parameter_meta {
+        bam: "Input BAM format file to generate coverage for"
```
claymcleod: Input file to get read groups for?

workflows/qc/quality-check-standard.wdl
```diff
@@ -0,0 +1,155 @@
...
+import "https://raw.githubusercontent.com/stjudecloud/workflows/rfcs/qc-workflow/tools/md5sum.wdl"
+import "https://raw.githubusercontent.com/stjudecloud/workflows/rfcs/qc-workflow/tools/multiqc.wdl" as mqc
+
+workflow quality_check {
+    input {
+        File gencode_gtf
+        File fastq_screen_db
+        String fastq_format = "sanger"
+        String experiment
+        File bam
```
claymcleod: Put BAM in the first position here.

workflows/qc/quality-check-standard.wdl
```diff
@@ -0,0 +1,155 @@
...
+task parse_input {
+    input {
+        String input_experiment
+        String input_strand
+        File input_fq_db
+        String input_fq_format
+    }
+
+    command {
+        if [ "~{input_experiment}" != "WGS" ] && [ "~{input_experiment}" != "WES" ] && [ "~{input_experiment}" != "RNA" ]; then
```
claymcleod: For experiment, let's do `RNA` -> `RNA-seq`

claymcleod:
Other thoughts:

* For each comment, ensure you finish the thread with what we decide and mark resolved. This will really help me as I review future changes to know what was resolved and what is outstanding.
* I don't recall seeing `input.json`s for the qc workflows. Let's make sure we have those.

docker/STAR/1.0.0-alpha/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM stjudecloud/conda-base:1.0.0-alpha AS builder
```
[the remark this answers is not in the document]
a-frantz: Agreed. I'm planning on that being the last commit before merging. That and pointing all the imports back to master.

.github/workflows/docker-branch.yml
```diff
@@ -0,0 +1,46 @@
...
+          for tag in docker/"$dir"/*; do
+            tag=$(basename "$tag")
+            # check tag is SemVer compliant
+            if ! [[ $tag =~ ^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*)?(\+[0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*)?$ ]]; then
+              >&2 echo "Tag subdirectory $tag for tool $dir is not SemVer compliant."
+              exit 1
+            fi;
+          done
+        done
+  docker_build_branch:
```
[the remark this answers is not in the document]
a-frantz: I think you're right. On the master CI I don't build if it's not SemVer compliant, would make sense to make that consistent on branches as well.

.github/workflows/lint-check.yml
```diff
@@ -0,0 +1,41 @@
...
+            awk '!/https:\/\/raw.githubusercontent.com\/stjudecloud\/workflows\/master/')
+          if [ -n "$bad_lines" ]; then
+            >&2 echo "Must import files from the master branch on Github."
+            >&2 echo "The following lines in file $file are bad:"
+            >&2 echo "$bad_lines"
+            EXITCODE=1
+          fi
+        done
+        exit $EXITCODE
+  docker_pull_check:
```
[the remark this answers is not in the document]
a-frantz: I rewrote it and added some better logging. Want to take another look at it?

a-frantz:
Should an input for the qc workflow be the `.bai`? It's already being generated by `rnaseq-standard`, and we currently regenerate it for qc *and* provide it as an output. The user will just have two `bai`s when they aggregate the results.

tools/ngsderive.wdl
a-frantz: Yes I have. This works.

tools/multiqc.wdl
a-frantz: Screen Shot 2020-03-12 at 11 18 52 AM
You're right. I'm looking at the MultiQC docs to figure out the right way to do this.

a-frantz: Do you have any idea why it fails to get the sample name from `bamqc` but succeeds for `qualimap rnaseq`?

tools/qc.wdl
a-frantz: Just went ahead and removed all references to `rseqc`

tools/fastqc.wdl
a-frantz: I implemented the wait on `picard validate` for the QC pipeline, but we don't validate the input bam in `rnaseq-standard`, so I left it using `wait_var`. It's only 2 tasks using it. It might make sense to validate the input bam before we start splitting it into fastqs. And if we did that, we could completely remove all `wait_var`s

tools/multiqc.wdl
a-frantz: I wrote a solution, but I don't love it:
```
# MultiQC should be able to strip any extensions on the BAM name
# down to the sample name, and will use that sample name for results
qualimap_bamqc_dir=$(basename ~{sorted_bam} '.bam')
mkdir "$qualimap_bamqc_dir"
tar -xzf ~{qualimap_bamqc} --strip-components=1 -C "$qualimap_bamqc_dir"/;
echo "$qualimap_bamqc_dir"/genome_results.txt >> file_list.txt
for file in `find "$qualimap_bamqc_dir"/raw_data_qualimapReport/`; do
    echo $file >> file_list.txt
done
```
I left `qualimap_rnaseq` alone, because MultiQC could parse the sample name from the file contents.

docker/STAR/1.0.0-alpha/sort_star_input.py
```diff
@@ -0,0 +1,88 @@
+import argparse
+
+
+def sort_fastqs(read_one_fastqs, read_two_fastqs):
+    read1_names = [fastq.split('/')[-1] for fastq in read_one_fastqs]
+    read2_names = [fastq.split('/')[-1] for fastq in read_two_fastqs]
+    
+    for i in range(1, len(read1_names)):
```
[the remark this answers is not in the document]
a-frantz: What I'm doing is reordering the full paths, `read_<one/two>_fastqs`, based on the alphanumeric sorting of the basenames, `read<1/2>_names`.

Using `sorted()` on the paths will be wrong, because it will sort based on the cromwell directories, not the read name. And using `sorted()` on the basenames won't allow me to map those index changes back to the file paths.

`enumerate()` won't work because I'm iterating over `len() - 1`. I don't want to iterate over every element.

This whole thing absolutely needs better documentation. I wasn't even sure if it would solve the problem with the MLL samples when I wrote it, so didn't bother documenting till I knew that what I thought the problem was, actually was the problem. Might have thrown out this whole script.

docker/STAR/1.0.0-alpha/sort_star_input.py
```diff
@@ -0,0 +1,88 @@
...
+        read2_names[j+1] = key
+        read_two_fastqs[j+1] = key_pair
+      
+    return (read_one_fastqs, read_two_fastqs)
+
+def sort_read_groups(read_groups_string):
+    read_groups = [rg for rg in read_groups_string.split(' , ')]
+    ids = [flags.split(' ')[0].split(':')[1] for flags in read_groups]
+
+    for i in range(1, len(ids)):
```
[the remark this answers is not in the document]
a-frantz: Yes, I can write a function to do all the sorting based on two lists. Lots of redundant code here. Wrote it quick and dirty first.

docker/fastq_screen/1.0.0-alpha/fastq_screen.conf
claymcleod: I meant any line that starts with `#` or a blank line should be removed. E.g.

```
##############
## Databases #
##############
#########
## Human - sequences available from
## ftp://ftp.ensembl.org/pub/current/fasta/homo_sapiens/dna/
DATABASE	Human	/tmp/fastq-screen-db/Human/Homo_sapiens.GRCh38

#########
## Mouse - sequence available from
## ftp://ftp.ensembl.org/pub/current/fasta/mus_musculus/dna/
DATABASE	Mouse	/tmp/fastq-screen-db/Mouse/Mus_musculus.GRCm38
```

would become

```
DATABASE	Human	/tmp/fastq-screen-db/Human/Homo_sapiens.GRCh38
DATABASE	Mouse	/tmp/fastq-screen-db/Mouse/Mus_musculus.GRCm38
```

You might keep one comment at the top saying something like:

```
## Data available from the following sources:
##
## ftp://ftp.ensembl.org/pub/current/fasta/homo_sapiens/dna/
## ftp://ftp.ensembl.org/pub/current/fasta/mus_musculus/dna/
## ...etc...
```

tools/fastq_screen.wdl
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    output {
+        File db = tar_filename
+    }
+
+    meta {
+        author: "Clay McLeod, Andrew Frantz"
+        email: "clay.mcleod@STJUDE.org, andrew.frantz@stjude.org"
+        description: "This WDL tool downloads the Fast Q Screen database and archives it."
```
claymcleod:
```suggestion
        description: "This WDL tool downloads the FastQ Screen database and archives it."
```

tools/fastq_screen.wdl
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    output {
+        Array[File] out_files = glob("~{output_basename}*")
+    }
+
+    meta {
+        author: "Andrew Frantz"
+        email: "andrew.frantz@stjude.org"
+        description: "This WDL tool runs Fast Q Screen on a pair of fastqs."
```
claymcleod:
```suggestion
        description: "This WDL tool runs FastQ Screen on a pair of fastqs."
```

tools/qualimap.wdl
```diff
@@ -102,6 +116,6 @@ task rnaseq {
...
```
claymcleod: Does this need to be updated to the new capitalized values?

a-frantz: Yes it does. Looks like that parameter info was already out of date before the change, which is why I missed it. Good catch.

a-frantz: Which actually, I've noticed a lot of meta info was already out of date. I'll need to include something for that in WDLdoc. I'll go through all the meta info and fix anything that's out of date before merging. Do we care about requiring everything to have meta data? Or only be concerned with things that are wrong?

claymcleod: I think it should be considered a warning. Maybe wdldoc can have a `--strict` option.

workflows/qc/inputs/quality-check-standard-inputs.json
```diff
@@ -0,0 +1,10 @@
+{
+    "quality_check.bam": "/cromwell/inputs/sample.bam",
+    "quality_check.gencode_gtf": "/cromwell/inputs/gencode.v31.annotation.gtf (only required for RNA-seq data)",
+    "quality_check.experiment": "<WGS/WES/RNA-seq>",
+    "quality_check.strand": "<Stranded-Reverse/Stranded-Forward/Unstranded> (optional, if not specified, we will try to guess using ngsderive)",
+    "quality_check.fastq_screen_db": "/cromwell/inputs/fastq-screen-db.tar.gz",
+    "quality_check.fastq_format": "<sanger/illumina1.3> (default is sanger)",
```
claymcleod:
```suggestion
    "quality_check.fastq_format": "<sanger/illumina1.3> (default = sanger)",
```

claymcleod:
https://github.com/stjudecloud/workflows/blob/rfcs/qc-workflow/workflows/qc/quality-check-standard.wdl#L45 this should default to `true`

tools/qc.wdl
```diff
@@ -4,57 +4,40 @@
...
-            elif (( $(echo "$(echo "~{in}" | tail -n 2 | head -n 1 | sed  's/^.*: //') > 0.5" | bc -l ) ))
-            then 
-               echo 'strand-specific-forward' > stdout.txt
-            else
-               echo 'non-strand-specific' > stdout.txt
-            fi
-        else
-           echo "infer_experiment failed to determine type" > /dev/stderr
-           exit 1 
+           if [ "$(echo "~{in}" | grep -c "ERROR")" -gt 0 ]
```
claymcleod: I think this can be cleaner. Just set the string to grep.

```
if [ "~{strict}" == "true" ]; then
  GREP_PATTERN="ERROR"
else
  GREP_PATTERN="(ERROR|WARNING)"
fi

if [ "$(echo "~{in}" | grep -Ec "$GREP_PATTERN")" -gt 0 ]
```

claymcleod: Also, does that last line actually work with double quotes embedded in double quotes?

tools/qualimap.wdl
```diff
@@ -9,32 +9,36 @@ task bamqc {
...
-        String prefix = basename(bam, ".bam")
+        String prefix = basename(bam, ".bam") + "_qualimap_bamqc_results"
         Int max_retries = 1
         Int memory_gb = 8
         Int? disk_size_gb
     }
-    Int java_heap_size = ceil(memory_gb * 0.9)
 
+    Int java_heap_size = ceil(memory_gb * 0.9)
+    String out_directory = select_first([prefix, "qualimap_bamqc_results"])
```
claymcleod: Instead of stripping the components in the multiqc step and writing to a folder, why don't we just use whatever top-level folder named was tar-ed up in this step (and don't even deal with creating a new folder in multiqc). That way, at least we isolate the problem of picking a name to here. We could even just set the prefix as `basename(bam, ".bam")` by default, which accomplishes your solution in the Slack channel

tools/qualimap.wdl
```diff
@@ -51,46 +55,56 @@ task bamqc {
 task rnaseq {
     input {
         File bam
-        File gencode_gtf
-        String outdir = "qualimap_rnaseq"
+        File? gencode_gtf
+        String prefix = basename(bam, ".bam") + "_qualimap_rnaseq_results"
```
claymcleod: Same as comment above.

workflows/qc/quality-check-standard.wdl
```diff
@@ -0,0 +1,166 @@
...
+    call ngsderive.instrument as ngsderive_instrument { input: bam=validate_bam.validated_bam, max_retries=max_retries }
+    call ngsderive.read_length as ngsderive_read_length { input: bam=validate_bam.validated_bam, max_retries=max_retries }
+    call qualimap.bamqc as qualimap_bamqc { input: bam=validate_bam.validated_bam, max_retries=max_retries }
+
+    call samtools.subsample as samtools_subsample { input: bam=validate_bam.validated_bam, max_retries=max_retries }
+    call picard.bam_to_fastq as b2fq { input: bam=samtools_subsample.sampled_bam, max_retries=max_retries }
+    call fq.fqlint { input: read1=b2fq.read1, read2=b2fq.read2, max_retries=max_retries }
+    call fq_screen.fastq_screen as fastq_screen { input: read1=b2fq.read1, read2=b2fq.read2, db=fastq_screen_db, format=fastq_format, max_retries=max_retries }
+    
+    if (experiment == "RNA-seq") {
```
claymcleod: Let's call this `RNA-Seq`. Sorry about this, we just changed capitalization.

workflows/rnaseq/rnaseq-standard.wdl
```diff
@@ -77,38 +81,33 @@ workflow rnaseq_standard {
...
+    call qc.parse_validate_bam { input: in=validate_bam.out, max_retries=max_retries }
     call ngsderive.infer_strand as ngsderive_strandedness { input: bam=picard_sort.sorted_bam, bai=samtools_index.bai, gtf=gencode_gtf, max_retries=max_retries }
-    call qualimap.bamqc as qualimap_bamqc { input: bam=picard_sort.sorted_bam, max_retries=max_retries }
-    call qualimap.rnaseq as qualimap_rnaseq { input: bam=picard_sort.sorted_bam, gencode_gtf=gencode_gtf, strand=strand, inferred=ngsderive_strandedness.strandedness, max_retries=max_retries }
-    call htseq.count as htseq_count { input: bam=picard_sort.sorted_bam, gtf=gencode_gtf, strand=strand, inferred=ngsderive_strandedness.strandedness, max_retries=max_retries }
-    call samtools.flagstat as samtools_flagstat { input: bam=picard_sort.sorted_bam, max_retries=max_retries }
+    call htseq.count as htseq_count { input: bam=picard_sort.sorted_bam, gtf=gencode_gtf, provided_strand=provided_strand, inferred_strand=ngsderive_strandedness.strandedness, max_retries=max_retries }
     call md5sum.compute_checksum { input: infile=picard_sort.sorted_bam, max_retries=max_retries }
```
claymcleod: We can remove this here, it's going to be computed as the first step in the QC pipeline.

claymcleod:
Andrew, once you make these changes, can you please run the RNA-Seq v2 pipeline on all the Clinical Pilot samples? Let's check the concordance.

tools/fastq_screen.wdl
```diff
@@ -0,0 +1,77 @@
...
+
+    String output_basename = basename(read1, "R1.fastq") + "screen"
+    String db_name = basename(db)
+
+    command {
+        cp ~{db} /tmp
+        tar -xsf /tmp/~{db_name} -C /tmp/
+
+        format_arg=''
+        if [[ "~{format}" = "illumina1.3" ]]; then
```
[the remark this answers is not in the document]
a-frantz: The user's input is already checked in `parse_input`, so at this point we know it's either empty (which defaults to `sanger`) or the correct spelling of `illumina1.3`.

docker/STAR/1.0.0-alpha/sort_star_input.py
```diff
@@ -0,0 +1,153 @@
...
+arguments out to files which can be supplied to STAR via the `cat` command.
+"""
+
+import argparse
+import os
+from typing import List, Tuple
+
+
+def sort_lists(
+    target_list: List[any],
```
jsunny23:
```suggestion
    target_list: List[Any],
```
You would also need to import `Any` from `typing`.

or, if you know the types that you expect
```suggestion
    target_list: List[Union[int, str]],
```

workflows/rnaseq/rnaseq-standard.wdl
a-frantz: @adthrasher 's TSNE workflow relies on gene counts being computed in this workflow. However, it wouldn't be hard to modify that workflow to just run `htseq-count` on its own. Which route should we go? @claymcleod

claymcleod: I'm referring to the md5sum here, not htseq count

tools/qc.wdl
a-frantz: @claymcleod @jsunny23 Was going to make that suggested change, realized essentially every piece of validate sam was broken. However the double quotes is fine, which is what I was expecting to be the problem. Basically the output was never being saved correctly, but should be good now.

docker/STAR/1.0.0-alpha/sort_star_input.py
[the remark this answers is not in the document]
claymcleod: What about something like this? I would this this is more canonical/simple.

```python
filepaths = [
    "/Users/cmcleod/baz",
    "/Users/afrantz/foo",
    "/Users/jsunny/bar"
]

basenames = [
    "baz",
    "foo",
    "bar"
]

sorted_filepaths, sorted_basenames = zip(*sorted(zip(filepaths, basenames), key = lambda item: item[1]))
print(f"Sorted filepaths: {sorted_filepaths}")
print(f"Sorted basenames: {sorted_basenames}")
```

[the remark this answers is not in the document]
a-frantz: IMO that's much less clear. Someone just casually looking at the code would need a much better understanding of Python to follow that one liner as opposed to the insertion sort function I wrote. And this whole script takes less than a second to run, so IMO any performance differences are negligible.

However that is leveraging Python builtins, which is more professional. It admittedly felt silly rewriting insertion sort 😂.

I think since performance is moot it comes down to which is more "presentable" code. I can see it going either way, readability vs professionalism; not exactly sure who the audience for this code is.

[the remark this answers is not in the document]
a-frantz: After taking a look at that code in context I like it more than my implementation. Going to commit your suggested change here.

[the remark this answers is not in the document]
claymcleod: Sounds good! In this case, I'd argue the shorter code actually makes it more understandable. This might be a prime place for a comment for those not familiar argument unpacking.

tools/picard.wdl
```diff
@@ -56,7 +56,7 @@ task validate_bam {
         Boolean summary_mode = false
         Boolean index_validation_stringency_less_exhaustive = false
         Boolean ignore_warnings = true
-        String output_filename = basename(bam, ".bam") + ".validation_report.txt"
+        String output_filename = basename(bam, ".bam") + ".ValidateSamFile.txt"
```
claymcleod, line 59: Why this change? Is this what `multiqc` expects? @a-frantz

a-frantz: Yes. There were a bunch of our results not showing up in MultiQC, so I went through everything and checked what it was expecting. I'm testing it now, but the docs say that should be the file name extension.

claymcleod: Good deal.

PR #18, opened 2020-03-04
Overall Improvements to the workflows

jsunny23:
Appreciate any comments or suggestions to improve the workflows based on running these workflows in practice. This is still a WIP right now, but I wanted to get as many improvements in front of people to get their feedback. 

**Changes made:**

- LSF config file
  - [x] removed singularity requirement (based on our LSF cluster configuration)
  - [x] Default redirect to `stdout.lsf` and `stderr.lsf` files instead of `stdout` and `stderr`. This should fix steps that parse stdout. 

- [x] Added a period to the final output names so that prefix does not need to include a period. It seems like something external users may not figure out until they run the workflow. It also makes scripting a bit easier as the `output_prefix` is exactly the sample name.  

- [x] Moved `prepare_read_groups_for_star` step to be part of `get_read_groups`. This step usually took ~5 seconds to complete and didn't warrant provisioning a new node with docker requirement. 

- [x] Moved `parse_validate_bam` step to be part of `validate_bam`. This step usually took ~1 second to complete. 

- [x] Renamed `multiqc_zip` to `multiqc_tar_gz` to better represent the actual file and to keep it consistent with stardb.

- [x] Increased MAX_OUTPUT from Picard ValidateSamFile to better check the errors and warnings from the file.

### DIFF
--- a/conf/lsf.conf
+++ b/conf/lsf.conf
@@ -13,13 +13,11 @@ backend {
         runtime-attributes = """
         Int cpu = 1
         Int hosts = 1
-        Float? memory_mb = 3800
+        Float? memory_mb = 4096
         String lsf_queue = "standard"
         String? lsf_job_group
         String? docker 
         """
-
-        #submit = """bsub -J ${job_name} -cwd ${cwd} -o ${out} -e ${err} -R rusage[mem=${memory_mb + "MB"}] -R "span[hosts=1]" -n ${cpu} ${"-q " + lsf_queue}  /usr/bin/env bash ${script}"""
 
         submit = """
           bsub \
@@ -39,11 +37,11 @@ backend {
             -q ${lsf_queue} \
             -n ${cpu} \
             ${"-g " + lsf_job_group} \
-            -R "select[singularity] rusage[mem=${memory_mb}] span[hosts=${hosts}]" \
+            -R "rusage[mem=${memory_mb}] span[hosts=${hosts}]" \
             -J ${job_name} \
             -cwd ${cwd} \
-            -o ${cwd}/execution/stdout \
-            -e ${cwd}/execution/stderr \
+            -o ${cwd}/execution/stdout.lsf \
+            -e ${cwd}/execution/stderr.lsf \
             "singularity exec --bind ${cwd}:${docker_cwd} docker://${docker} ${job_shell} ${script}"
         """
 
--- a/conf/lsf.conf
+++ b/conf/lsf.conf
@@ -4,6 +4,25 @@ call-caching {
   enabled = true
 }
 
+database {
+  profile = "slick.jdbc.HsqldbProfile$"
+  db {
+    driver = "org.hsqldb.jdbcDriver"
+    url = """
+    jdbc:hsqldb:file:cromwell-executions/cromwell-db/cromwell-db;
+    shutdown=false;
+    hsqldb.default_table_type=cached;hsqldb.tx=mvcc;
+    hsqldb.result_max_memory_rows=10000;
+    hsqldb.large_data=true;
+    hsqldb.applog=1;
+    hsqldb.lob_compressed=true;
+    hsqldb.script_format=3
+    """
+    connectionTimeout = 120000
+    numThreads = 1
+   }
+}
+
 backend {
   default = LSF
   providers {
--- a/conf/lsf.conf
+++ b/conf/lsf.conf
@@ -40,33 +40,33 @@ backend {
 
         submit = """
           bsub \
-            -q ${lsf_queue} \
-            -n ${cpu} \
-            ${"-g " + lsf_job_group} \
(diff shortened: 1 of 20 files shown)


tools/picard.wdl
```diff
@@ -52,6 +52,7 @@ task mark_duplicates {
 task validate_bam {
     input {
         File bam
+        Boolean strict = true
```
adthrasher: The name strict seems counterintuitive based on how it is used. If strict mode, we only exit on errors. If not strict, then we exit on errors or warnings.

conf/lsf.conf
```diff
@@ -39,11 +37,11 @@ backend {
             -q ${lsf_queue} \
             -n ${cpu} \
             ${"-g " + lsf_job_group} \
-            -R "select[singularity] rusage[mem=${memory_mb}] span[hosts=${hosts}]" \
```
adthrasher: Is singularity available on all of our cluster nodes now?

tools/star.wdl
```diff
@@ -124,8 +124,8 @@ task alignment {
     }
 
     output {
-        File star_log = output_prefix + "Log.final.out"
-        File star_bam = output_prefix + "Aligned.out.bam"
+        File star_log = output_prefix + ".Log.final.out"
```
adthrasher: Once this is merged, I'll need to update the t-SNE workflow to supply a prefix without the dot.

conf/lsf.conf
jsunny23: Yeah, all the `rhel7` nodes have singularity. Currently, only test nodes still use the singularity tag, so removing the requirement allows it to run on any `rhel7` node.

conf/lsf.conf
```diff
@@ -39,11 +37,11 @@ backend {
...
+            -R "rusage[mem=${memory_mb}] span[hosts=${hosts}]" \
```
adthrasher: We may want to consider adjusting the memory parameter. Generally, we specify total memory in our tasks, but in our LSF environment, the memory is per core. So if a task specifies C cores and N GB of RAM, LSF will interpret that as C * N GB of memory.

tools/star.wdl
a-frantz: You cloud also just cherrypick the commit. Might be easier and you can do it now instead of waiting and potentially forgetting. Just my two cents.

claymcleod:
Is it worth merging this in an taking your last TODO as another task?

conf/lsf.conf
jsunny23: Agreed, changed this line in `lsf.conf` to 
`-R "rusage[mem=${round(round(memory_mb)/cpu)}] span[hosts=${hosts}]"`
which should handle the memory per core.

jsunny23:
Yes, I think I can merge this soon. I just want to do a test run to make sure that I haven't changed any functionality.

conf/lsf.conf
a-frantz: `~{}` syntax is preferred to `${}`. For two reasons, 1) `${}` isn't supported at all in `<<<...>>>` task definitions, 2) is a personal opinion, I think using `$` can make it unclear on a quick glance whether it's a Bash variable or a WDL variable. I suggest changing all instances of `${}` to `~{}`.

tools/picard.wdl
```diff
@@ -64,6 +65,19 @@ task validate_bam {
         picard -Xmx${java_heap_size}g ValidateSamFile I=${bam} \
             IGNORE=INVALID_PLATFORM_VALUE \
             IGNORE=MISSING_PLATFORM_VALUE > stdout.txt
+
+        if [ "${strict}" == "true" ]
```
a-frantz: Same as above about `~{}`

tools/util.wdl
```diff
@@ -14,9 +14,13 @@ task get_read_groups {
     Float bam_size = size(bam, "GiB")
     Int disk_size = ceil((bam_size * 2) + 10)
 
-    command {
-        samtools view -H ${bam} | grep "@RG" > stdout.txt
-    }
+    command <<<
+        samtools view -H ${bam} | grep "@RG" \
```
a-frantz: Same as above

conf/lsf.conf
jsunny23: Agreed. I changed the syntax to use ~{} instead of ${}.

tools/picard.wdl
jsunny23: Agreed. I've changed it to `ignore_warnings` to make it clearer that that flag can be set/unset to report errors only or errors/warnings.

claymcleod:
🎉

tools/star.wdl
jsunny23: @adthrasher This is merged in now FYI.

PR #19, opened 2020-03-07
improvement: tar qualimap outputs

claymcleod:
Fixes:

* Improved `samtools subsample` to not error when give a BAM with too few reads.
* Fixed all of the outputs not being saved from `qualimap` by both tarring up the output and fixing the docker container with the missing fonts.


claymcleod:
@a-frantz this is ready for your review and you can merge in. A few things are still outstanding:

* Currently, the multiqc report is messed up and reports the fields on two different lines. I'm thinking this is due to me untarring the file to `qualimap_bamqc/`. The solution would be to instead write this to a folder with the sample name, but it's not currently being passed in. It seems related to the results folder organization we talked about earlier, so I'm going to leave that piece to you.
* I still don't see the novel/partially known/known splice junctions graph showing up in multiqc. Maybe we should look at the qualimap rnaseq module and see if it is even checked for in multiqc?

a-frantz:
@claymcleod what is the `test.dat` file in the root? It's some binary. Comitted by accident?

a-frantz:
@claymcleod waiting for your review. `import_syntax_check` and `miniwdl_check` are failing because I pointed them back to the rfcs branch instead of this one in preparation for the merge.

.github/workflows/lint-check.yml
```diff
@@ -9,13 +9,18 @@ jobs:
...
-        if [ -n "$bad_lines" ]; then
-          echo "$bad_lines"
-          exit 1
-        fi
+        EXITCODE=0
+        for file in $(find . -name '*.wdl'); do
+          bad_lines=$(awk '/import/' "$file" | \
+            awk '!/https:\/\/raw.githubusercontent.com\/stjudecloud\/workflows\/master/')
+          if [ -n "$bad_lines" ]; then
+            >&2 echo "Must import files from the master branch on gihub."
```
claymcleod: Github*

PR #22, opened 2020-05-15
Changing STAR alignment to use either the number of cores if specified or all available cores

adthrasher:
In cloud environments, STAR defaulted to 1 core despite the instance having multiple cores to satisfy the memory requirement. This changes the threading of STAR to use all available cores, unless the number of CPUs was explicitly passed.

### DIFF
--- a/tools/deeptools.wdl
+++ b/tools/deeptools.wdl
@@ -13,7 +13,7 @@ task bamCoverage {
         Int max_retries = 1
         Int memory_gb = 5
         Int ncpu = 1
-        Boolean use_ncpu = false
+        Boolean detect_nproc = false
     }
 
     Float bam_size = size(bam, "GiB")
@@ -23,7 +23,7 @@ task bamCoverage {
         set -euo pipefail
         
         n_cores=~{ncpu}
-        if [ ${true='true' false='' use_ncpu} ]
+        if [ ${true='true' false='' detect_nproc} ]
         then
             n_cores="max""
         fi
--- a/tools/deeptools.wdl
+++ b/tools/deeptools.wdl
@@ -25,7 +25,7 @@ task bamCoverage {
         n_cores=~{ncpu}
         if [ ${true='true' false='' detect_nproc} ]
         then
-            n_cores="max""
+            n_cores="max"
         fi
 
         if [ ! -e ~{bam}.bai ]
--- a/tools/deeptools.wdl
+++ b/tools/deeptools.wdl
@@ -12,6 +12,8 @@ task bamCoverage {
         String prefix = basename(bam, ".bam")
         Int max_retries = 1
         Int memory_gb = 5
+        Int ncpu = 1
+        Boolean use_ncpu = false
     }
 
     Float bam_size = size(bam, "GiB")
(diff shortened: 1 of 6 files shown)


a-frantz:
Will this cause problems on clusters? Where potentially only one core was allocated, but `nproc` can report something like 64?

adthrasher:
@a-frantz - Yes, it will. I think we should address with documentation that you need to explicitly set core parameters for managed HPC environments.

a-frantz:
Does the core parameter get ignored/messed up on either cloud environments or HPCs? Wouldn't it be better to just leave it up to the user how many to provision?

adthrasher:
Clay and I noticed that in DNAnexus alignment was using 1 core of an 8 core machine. The number of cores isn't explicitly set, so they are determined by the minimum machine that satisfies the memory requirement. In that case, there is no reason to leave 7 cores idle. We're not explicitly requesting 8, but if we have them, we want to use them.

a-frantz:
IMO it would be better to document "on a cloud environment request <blank> cores because they will be allocated regardless" than to require HPC users to always set it or fail. Either way, one environment is going to need to set it. The result of not setting it on the cloud is sub-optimal performance, the result of not setting it on a HPC is potential failure.

Another option is adding a `detect_nproc` flag defaulting to false, and suggesting it be set to true for cloud environments.

claymcleod:
I like the `detect_nproc` idea.

One thing we are trying to optimize for is ease of porting the workflow to the cloud using `dx-wdl`. Any modifications after that step make the process much more error-prone. I think up until this point, we've just been relying on the default parameters to be set appropriately to achieve that simplicity.

@adthrasher how much would this complicate the build process by `dx-wdl` if we wanted to specify the default parameter for the cloud environment? Is it just an option to the build call?

adthrasher:
@claymcleod - We could add a parameter and set it by default. It has to be a workflow level option as dxWDL doesn't support specifying task level options (unlike other backends).

a-frantz:
Is DNAnexus planning to change the inability to set task level parameters? If our default resources aren't enough for a step is there just no way to run that sample? Because right now, rarely the `qualimap` steps run out of memory, and I've just been upping the RAM requirement on a per sample basis. Is that not possible when built with dx-wdl?

adthrasher:
I think it is an implementation issue with dxWDL. There is discussion here: https://github.com/dnanexus/dxWDL/blob/master/doc/MissingCallArguments.md. You can set the instance type on a per-stage basis at runtime, so for memory that works fine, but in the case of STAR we were explicitly limiting the number of threads to the input number of cores.

adthrasher:
@a-frantz - I've converted to an optional parameter that defaults to false. This parameter determines whether "nproc" is used on the ncpu argument.

tools/deeptools.wdl
```diff
@@ -20,15 +22,22 @@ task bamCoverage {
     command {
         set -euo pipefail
         
+        n_cores=~{ncpu}
+        if [ ${true='true' false='' detect_nproc} ]
+        then
+            n_cores="max""
```
claymcleod, line 28:
```suggestion
            n_cores="max"
```

workflows/rnaseq/rnaseq-standard.wdl
```diff
@@ -56,6 +57,7 @@ workflow rnaseq_standard {
         strand: "empty, 'Stranded-Reverse', 'Stranded-Forward', or 'Unstranded'. If missing, will be inferred"
         output_prefix: "Prefix for output files"
         max_retries: "Number of times to retry failed steps"
+        detect_nproc: "Determine number of cores and use all for multi-core steps"
```
claymcleod, line 60:
```suggestion
        detect_nproc: "Use all available cores for multi-core steps"
```

PR #24, opened 2020-07-07
Replacing lsf.conf with known good version using ${} syntax. 

adthrasher:
Adding check-job-alive script to return proper information from LSF to Cromwell about job status.

### DIFF
--- a/conf/lsf.conf
+++ b/conf/lsf.conf
@@ -1,7 +1,7 @@
 include required(classpath("application"))
 
 call-caching {
-  enabled = true
+  enabled = false
 }
 
 database {
@@ -17,6 +17,7 @@ database {
     hsqldb.applog=1;
     hsqldb.lob_compressed=true;
     hsqldb.script_format=3
+    hsqldb.lock_file=false
     """
     connectionTimeout = 120000
     numThreads = 1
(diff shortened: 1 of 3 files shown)


conf/lsf.conf
```diff
@@ -32,41 +33,46 @@ backend {
...
         Int cpu = 1
         Int hosts = 1
-        Float? memory_mb = 4096
-        String lsf_queue = "standard"
+        Int? memory_mb = 4096
+        String lsf_queue = "compbio"
         String? lsf_job_group
-        String? docker 
+        String? docker
+        # Float memory_per_cpu = memory_mb / cpu
```
jsunny23, line 40: Remove commented line.

conf/lsf.conf
```diff
@@ -32,41 +33,46 @@ backend {
...
         """
 
+        #submit = """bsub -J ${job_name} -cwd ${cwd} -o ${out} -e ${err} -R rusage[mem=${memory_mb + "MB"}] -R "span[hosts=1]" -n ${cpu} ${"-q " + lsf_queue}  /usr/bin/env bash ${script}"""
```
jsunny23, line 43: This isn't used either.

conf/lsf.conf
```diff
@@ -32,41 +33,46 @@ backend {
...
+            -e ${cwd}/execution/stderr.lsf \
+            "singularity exec --bind ${cwd}:${docker_cwd} docker://${docker} ${job_shell} ${script}"
         """
 
 
-        kill = "bkill ~{job_id}"
-        check-alive = "bjobs ~{job_id}"
+        kill = "bkill ${job_id}"
+        check-alive = "bjobs ${job_id}"
+        # check-alive = "check-job-alive ${job_id}"
```
jsunny23, line 74:
```suggestion
        check-alive = "check-job-alive ${job_id}"
```

Maybe we need to make a comment that the `check-job-alive` script must be in the user's path.

a-frantz: Yes that needs to be documented somewhere. Right now `check-job-alive` is the only file in `bin/` and it's purpose is very unclear.

adthrasher: For internal usage, we have https://wiki.stjude.org/display/devops/WDL+integration+with+LSF where I intend to add documentation about `check-job-alive`. For external users, do you have a preference on where to document that? An inline comment seems easy to overlook.

a-frantz: I'm not sure where it would be best for external users right now. It could go in the README, but it's very out of date at the moment and there's nowhere in the current doc it makes sense to put information about LSF configuration. Updating the whole thing needs to be done at some point, and when that's done it would make sense to include a section on configuration.

jsunny23: I think for now an inline comment is okay but I agree it will probably be overlooked. If they don't have `check-job-alive`, I believe cromwell will fail pretty early in the run so that should hopefully guide people to looking for some documentation about it. We should probably make updates to the README later and/or make Github wiki pages if there's a lot of things we want to document.

adthrasher: I inlined the comment, but also added bin as a section in the README.

a-frantz: I think this is good till we update the README/make a wiki 👍

PR #25, opened 2020-07-09
Only run FastQ Screen on WGS and WES

a-frantz:
FastQ Screen isn't meant to be run on RNA data


workflows/qc/quality-check-standard.wdl
```diff
@@ -101,12 +97,16 @@ workflow quality_check {
                 flagstat_file=samtools_flagstat.outfile,
                 fastqc_files=fastqc.out_files,
                 qualimap_bamqc=qualimap_bamqc.results,
-                fastq_screen=fastq_screen.out_files,
                 qualimap_rnaseq=qualimap_rnaseq.results,
                 max_retries=max_retries
         }
     }
     if (experiment == "WGS" || experiment == "WES") {
+        File fastq_screen_db_defined = select_first([fastq_screen_db, "No DB"])
```
jsunny23, line 105: I'm a bit unsure about this: I feel like there should be a better way to coerce an optional input into a required input but after looking through the GitHub Issues, most of the discussion is aimed at development spec and I don't know if 1.0 has a way to do this.

tools/fastq_screen.wdl
```diff
@@ -50,7 +50,7 @@ task fastq_screen {
     Float db_size = size(db, "GiB")
     Int disk_size = ceil(db_size * 2)
 
-    String output_basename = basename(read1, "R1.fastq")
+    String sample_basename = basename(read1, "_R1.fastq")
```
claymcleod, line 53: Any assumptions about sample naming should be bubbled up to workflow options so that others can change it based on their format.

workflows/qc/quality-check-standard.wdl
a-frantz: I don't like it, but it's the best 1.0 way I could find.

Unfortunately the expression gets evaluated regardless of the evaluation of the if clause, which is why I had to include the `"no GTF/DB"` strings. Otherwise it was throwing errors that it couldn't select nothing, even though the expression logically shouldn't have been evaluated.

tools/fastq_screen.wdl
a-frantz: `_R1`/`_R2` are currently hardcoded suffixes in `picard.bam_to_fastq`, and I think any way to optionally set the suffixes trimmed here is going to be messy and unused. I don't think a use case for calling this without `picard.bam_to_fastq` first exists.

The best way I can think of to handle this with mutable suffixes would be to allow a workflow option for changing `_R`, and then applying `1/2` after the user supplied suffix (Otherwise there would need to be an option for a read 1 suffix and an option for a read 2 suffix). That option would also need to be passed to `picard.bam_to_fastq`, and potentially other tools making this assumption (I haven't checked if there are any).

Personally I think that's introducing an unnecessary complication. It will make the code messier, and introduce workflow/tool options for the user to review that will likely never be used. Are there any use cases you can think of where this would be useful?

workflows/qc/quality-check-standard.wdl
adthrasher: To my knowledge, this is the best current way to handle this.

tools/fastq_screen.wdl
adthrasher: You could make sample name an optional parameter. Then do something like `select_first(sample_name, inferred_name)`.

PR #26, opened 2020-07-20
Gzip fastqs

a-frantz:
What kind of validation do we want before we merge this? It runs without errors, but we are changing the STAR command. Do we want an R^2 test of some sort for RNA-Seq V2?

Also I made all the command sections use `<<<...>>>`, instead of having some use `{...}`. I like it for consistency's sake, but if people prefer to continue using `{...}` where possible I can revert that commit.

### DIFF
--- a/tools/deeptools.wdl
+++ b/tools/deeptools.wdl
@@ -19,7 +19,7 @@ task bamCoverage {
     Float bam_size = size(bam, "GiB")
     Int disk_size = ceil((bam_size * 4) + 10)
  
-    command {
+    command <<<
         set -euo pipefail
         
         n_cores=~{ncpu}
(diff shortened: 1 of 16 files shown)


claymcleod:
Ideally, yeah. @jsunny23 are the end-to-end tests ready yet?

jsunny23:
The end-to-end tests aren't ready yet as it still uses some hardcoded values and I'm running it on a private branch till I obfuscate some of the implementation details. I can pull in this branch and run the validations, but I need some more time before I can add the CI to this repo.

a-frantz (commented):
This is a 404 error. Should've changed `master/` not `workflows/`.

jsunny23:
Yup, thanks for the catch!

jsunny23 (approved):
I ran the CI on the branch and it didn't have any errors. The r^2 was 1 since the counts generated were the same, so this looks good to me. 

Please revert my changes to the import paths as well before you merge.

PR #27, opened 2020-08-10
Expression updates

adthrasher:
Make htseq memory configurable. Output Picard ValidateSam errors.

### DIFF
--- a/tools/htseq.wdl
+++ b/tools/htseq.wdl
@@ -12,7 +12,7 @@ task count {
         String provided_strand
         String inferred_strand
         String outfile = basename(bam, ".bam") + ".counts.txt"
-        Int memory_gb = 20
+        Int added_memory_gb = 20
         Int max_retries = 1
     }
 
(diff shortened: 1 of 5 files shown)


tools/htseq.wdl
```diff
@@ -29,6 +29,7 @@ task count {
                                          # in this case
 
     Float bam_size = size(bam, "GiB")
+    Float mem_size = bam_size + memory_gb
```
a-frantz, line 32: You should add an entry for `memory_gb` to the parameter meta to make it clear it's being added to the bam size, instead of an absolute value. Maybe rename the parameter to something like `added_memory_gb` to make that abundantly clear. Not sure about the renaming, but something to consider.

workflows/rnaseq_expression_classification/rnaseq_expression_classification_from_bams.wdl
```diff
@@ -109,7 +109,8 @@ workflow rnaseq_expression_classification_from_bams {
 
     parameter_meta {
         in_bams: {
-            help: "Provide bams to run for comparison"
+            help: "Provide bams to run for comparison",
+            patterns: ["*.bam"]
```
a-frantz, line 113: Think using `patterns` is going to make the `wdldoc` section ugly. Don't need to change it on your end, I can add handling for it in a new release of `wdldoc`. Just in the meantime, won't look great in the docs.

PR #30, opened 2020-11-02
fix: changes for dxWDL compatibility

a-frantz:


jsunny23 (approved):
Just curious, do you know how much memory overhead dxWDL needs approximately (like for qualimap)?

a-frantz:
No I'm not sure, but the previous value of 8gb for memory was only suitable for RNA samples. 32gb is more reasonable for WGS. That change should've been made for the workflow in any environment.

adthrasher (approved):

PR #31, opened 2020-11-04
Qc refactor

a-frantz:


a-frantz:
Are you guys ok renaming the 'strand' variables to `strandedness`? IMO it's more readable/explicit.

Also how well tested were all our Bash `if` tests that use in-line WDL parsing? `miniwdl check` has trouble with them, so I updated the samtools `nproc` tests to something it could parse.

But looking over the full project miniwdl check there are quit a few more. Not sure if it those are false errors on miniwdl's part, or if we messed up the in-line WDL var evaluations... They seem ok according to the spec, but I think miniwdl is pretty thorough and we've found undocumented WDL parsing quirks before.

adthrasher (approved):

adthrasher:
I'm fine with renaming `strand` -> `strandedness` that is more explicit. I tested the `nproc` statements originally and it worked, but moving the parsing of WDL out of the task makes sense.

jsunny23 (approved):

jsunny23:
I don't really mind either `strand` or `strandedness`, so if one's clearer, I think it's fine to update.

claymcleod:
All sounds good to me.

PR #32, opened 2020-12-18
Adding workflow from Fastqs for RNA-Seq V2.

adthrasher:
Add workflow to run the RNA-Seq V2 pipeline from Fastq files rather than starting from a BAM.


workflows/rnaseq/rnaseq-standard-fastq.wdl
```diff
@@ -0,0 +1,123 @@
+## # RNA-Seq Standard
+##
+## This WDL workflow runs the STAR RNA-seq alignment workflow for St. Jude Cloud.
+## The workflow takes an input BAM file and splits it into fastq files for each read in the pair.
```
a-frantz, line 4: The header should be re-written

workflows/rnaseq/rnaseq-standard-fastq.wdl
```diff
@@ -0,0 +1,123 @@
...
+import "https://raw.githubusercontent.com/stjudecloud/workflows/master/tools/deeptools.wdl"
+
+workflow rnaseq_standard {
+    input {
+        File gencode_gtf
+        Array[File] read_one_fastqs
+        Array[File] read_two_fastqs
+        File stardb_tar_gz
+        String strandedness = ""
+        String output_prefix = "sample"
```
a-frantz, line 49: Should this basename a fastq? Not a big fan of using `sample`, but a basename will likely be junk... Maybe it should be a required input?

workflows/rnaseq/rnaseq-standard-fastq.wdl
```diff
@@ -0,0 +1,123 @@
...
+    parameter_meta {
+        gencode_gtf: "GTF file provided by Gencode"
+        read_one_fastqs: "Input Fastq format file(s) with 1st read in pair to align"
+        read_two_fastqs: "Input Fastq format file(s) with 2nd read in pair to align"
+        stardb_tar_gz: "Database of reference files for the STAR aligner. Can be generated by `rnaseq-star-db-build.wdl`"
+        strandedness: "empty, 'Stranded-Reverse', 'Stranded-Forward', or 'Unstranded'. If missing, will be inferred"
+        output_prefix: "Prefix for output files"
+        max_retries: "Number of times to retry failed steps"
+        detect_nproc: "Use all available cores for multi-core steps"
+    }
```
a-frantz, line 65: missing meta for `read_groups`

a-frantz (changes requested):

workflows/rnaseq/rnaseq-standard-fastq.wdl
a-frantz: The STAR step will actually fail without a well-formed `read_groups`. My `sort_star_input.py` script isn't skippable/doesn't allow an empty `read_groups` argument. I can go ahead and fix that if you want.

workflows/rnaseq/rnaseq-standard-fastq.wdl
adthrasher: I considered that. I think it should probably just be a required parameter. The extension isn't necessarily consistent.

workflows/rnaseq/rnaseq-standard-fastq.wdl
adthrasher: Yes, thanks. I removed it for testing as the fastqs don't have read group information. I'll add it back.

workflows/rnaseq/rnaseq-standard-fastq.wdl
```diff
@@ -1,11 +1,9 @@
-## # RNA-Seq Standard
+## # RNA-Seq Standard from FastQ
 ##
-## This WDL workflow runs the STAR RNA-seq alignment workflow for St. Jude Cloud.
-## The workflow takes an input BAM file and splits it into fastq files for each read in the pair. 
+## This WDL workflow runs the STAR RNA-seq alignment workflow for St. Jude Cloud from fastq input..
```
a-frantz, line 3:
```suggestion
## This WDL workflow runs the STAR RNA-seq alignment workflow for St. Jude Cloud from fastq input.
```

workflows/rnaseq/rnaseq-standard-fastq.wdl
```diff
@@ -1,11 +1,9 @@
...
+## The workflow takes an paired input fastq files for each read in the pair.
```
a-frantz, line 4:
```suggestion
## The workflow takes paired input fastq files. 
```

workflows/rnaseq/rnaseq-standard-fastq.wdl
```diff
@@ -46,20 +44,21 @@ workflow rnaseq_standard {
         Array[File] read_two_fastqs
         File stardb_tar_gz
         String strandedness = ""
-        String output_prefix = "sample"
-        String read_groups = ""
+        String output_prefix
+        String read_groups
```
a-frantz, line 48:
```suggestion
        String? read_groups
```

adthrasher: I think `read_groups` should be a required parameter. Even if the user supplies "unknown", at least it would create a BAM with the correct number of read group records.

a-frantz (changes requested):

workflows/rnaseq/rnaseq-standard-fastq.wdl
a-frantz: O. I'm not familiar with what happens when STAR isn't supplied that parameter. Does it guess or is the concept of "read groups" lost? I modified the STAR step to just leave out `--outSAMattrRGline` if `read_groups` isn't provided.

adthrasher: I suppose that is OK then. I don't like producing BAMs with no @RG records, but maybe allowing that is OK and in practice, we won't do that.

workflows/rnaseq/rnaseq-standard-fastq.wdl
```diff
@@ -0,0 +1,123 @@
...
+        String output_prefix
+        String? read_groups
+        Int max_retries = 1
+        Boolean detect_nproc = false
+        Boolean validate_input = true
+    }
+
+    parameter_meta {
+        gencode_gtf: "GTF file provided by Gencode"
+        read_one_fastqs: "Input Fastq format file(s) with 1st read in pair to align. One file per read group."
```
a-frantz, line 56:
```suggestion
```
I messed up the suggestion comment. This line is included twice

a-frantz (approved):
Pending a test a run with and without `read_groups` (I'm curious what "without" looks like), this all looks good.

a-frantz:
Do we have any interest in supporting single end alignment? `read_two_fastqs` is currently required at the top level (and the STAR task). Which raises a question of what happens if a user supplies a single end BAM to the full workflow.

adthrasher:
I don't think so. The same assumption is true in the BAM-based workflow, I believe. I'm not sure how useful single-end RNA-Seq would be. Our DNA/bwa pipeline should support single-end sequencing though.

a-frantz:
I think if no one's asked for it, we don't need to implement it. But WARDEN allows single-end input, (and the test/example data is SE) so that leads me to believe there are valid use cases. But if there's no demand for it in this tool, that's fine.

claymcleod:
> I think if no one's asked for it, we don't need to implement it. But WARDEN allows single-end STAR input, (and the test/example data is SE) so that leads me to believe there are valid use cases. But if there's no demand for it in this tool, that's fine.

I agree with this approach, let's wait until someone asks for it.

a-frantz:
@adthrasher have we done any test runs yet?

adthrasher:
@a-frantz - Yes, I ran 177 samples from Dana Farber through from fastq.

a-frantz:
> @a-frantz - Yes, I ran 177 samples from Dana Farber through from fastq.

Was that with or without the `read_groups` parameter? Have we ensured it works as expected with both?

adthrasher:
@a-frantz - Turns out, I have a use case for single-end read support. So I've added that. The data I'm running also has no read group information, so I'm testing that feature as well.

tools/star.wdl
```diff
@@ -180,6 +180,6 @@ task alignment {
         read_one_fastqs: "An array of FastQ files containing read one information"
         read_two_fastqs: "An array of FastQ files containing read two information in the same order as the read one FastQ"
         stardb_tar_gz: "A gzipped TAR file containing the STAR reference files"
-        read_groups: "A string containing the read group information to output in the BAM file"
+        read_groups: "A string containing the read group information to output in the BAM file. Read group fields should be space delimited. Read groups should be comma separated. The ID for each read group must match the basename of a fastq file (up to the first period). Required fields: ID, PU, SM, PL, LB"
```
a-frantz, line 183: Why did this become required again? Also what's the significance of `ID, PU, SM, PL, LB`? That's not the format my sorting script is expecting or what's passed in the regular `rnaseq_standard`.

adthrasher: Picard ValidateSam errors if read groups are not present. Downstream analysis tools such as GATK also require it. So rather than handling special cases, I'd rather the pipeline just require read group information (even if it is just a bunch of "unknown" values. The read group format requires `ID, PU, SM, PL, LB` fields

tools/star.wdl
```diff
@@ -160,6 +180,6 @@ task alignment {
         read_one_fastqs: "An array of FastQ files containing read one information"
         read_two_fastqs: "An array of FastQ files containing read two information in the same order as the read one FastQ"
         stardb_tar_gz: "A gzipped TAR file containing the STAR reference files"
-        read_groups: "A string containing the read group information to output in the BAM file"
+        read_groups: "A string containing the read group information to output in the BAM file. Read group fields (Required fields: ID, LB, PL, PU, & SM) should be space delimited. Read groups should be comma separated, with a space on each side (e.g. ' , '). The ID for each read group must match the basename of a fastq file (up to the first period). Expected form: `ID:rg1 PU:flowcell1.lane1 SM:sample1 PL:illumina LB:sample1_lib1 , ID:rg2 PU:flowcell1.lane2 SM:sample1 PL:illumina LB:sample1_lib1`"
```
a-frantz, line 183: These are only required fields when part of RNA-Seq V2. At the task level, the only requirement is `ID:<fastq-basename matching string>[ , ID:<other fastq basename>...]`. 

I still think having the example in both will be good, but I'd switch it from `Expected form:` to `Example:` here, as the requirements are looser.

workflows/rnaseq/rnaseq-standard-fastq.wdl
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    parameter_meta {
+        gencode_gtf: "GTF file provided by Gencode"
+        read_one_fastqs: "Input Fastq format file(s) with 1st read in pair to align"
+        read_two_fastqs: "Input Fastq format file(s) with 2nd read in pair to align"
+        stardb_tar_gz: "Database of reference files for the STAR aligner. Can be generated by `rnaseq-star-db-build.wdl`"
+        strandedness: "empty, 'Stranded-Reverse', 'Stranded-Forward', or 'Unstranded'. If missing, will be inferred"
+        output_prefix: "Prefix for output files"
+        read_groups: "A space-delimited read group record for each read group. Exactly one fastq filename must match each read group ID from `read_one_fastqs` and `read_two_fastqs`. Expected form: `ID:rg1 PU:flowcell1.lane1 SM:sample1 PL:illumina LB:sample1_lib1 , ID:rg2 PU:flowcell1.lane2 SM:sample1 PL:illumina LB:sample1_lib1`"
```
a-frantz, line 61: Move the "required fields" line from the task description to here. Also bring in the line about space-comma (` , `) delimiters.

workflows/rnaseq/rnaseq-standard-fastq.wdl
```diff
@@ -0,0 +1,122 @@
...
+import "https://raw.githubusercontent.com/stjudecloud/workflows/master/tools/htseq.wdl"
+import "https://raw.githubusercontent.com/stjudecloud/workflows/master/tools/samtools.wdl"
+import "https://raw.githubusercontent.com/stjudecloud/workflows/master/tools/util.wdl"
+import "https://raw.githubusercontent.com/stjudecloud/workflows/master/tools/deeptools.wdl"
+
+workflow rnaseq_standard_fastq {
+    input {
+        File gencode_gtf
+        Array[File] read_one_fastqs
+        Array[File] read_two_fastqs
```
a-frantz, line 44: Sorry if we went over this before, can we not support SE in RNA-Seq V2? We've enable it for STAR, can it work with the other tools? The header of this file would need to updated if we enable it, FYI

a-frantz (changes requested):

tools/star.wdl
adthrasher: Picard ValidateSam requires all of those fields, so it will pass the sort step, but will error out on the Picard validate step.

a-frantz: But it's possible (probably only with an external user workflow) to be running the STAR task without the `ValidateSam` step. IMO we should only document what's needed for the specific task, and move those inter-connected requirements to the workflow level. We've already made it optional which would error on `ValidateSam`.

Sidenote: is it a Picard warning or error? Don't we have it strict on warnings by default, but configurable to ignore them? If a user doesn't want to fill in proper dummy values, can they disable an option to get past the validation step?

a-frantz, line 183: IMO this is how it should read:
```suggestion
        read_groups: "A string containing the read group information to output in the BAM file. If including multiple read group fields per-read group, they should be space delimited. Read groups should be comma separated, with a space on each side (e.g. ' , '). The ID field must come first for each read group and must match the basename of a fastq file (up to the first period). Example: `ID:rg1 PU:flowcell1.lane1 SM:sample1 PL:illumina LB:sample1_lib1 , ID:rg2 PU:flowcell1.lane2 SM:sample1 PL:illumina LB:sample1_lib1`"
```

adthrasher: Picard errors on missing read group fields e.g. `MISSING_PLATFORM_VALUE`. 

I'll swap the language to "Example" as you point out, this can be run without the downstream validation.

workflows/rnaseq/rnaseq-standard-fastq.wdl
adthrasher: ngsderive strandedness requires paired-end data, though we could conceivably update that in the future. htseq-count works with both SE and PE data. So does deeptools bamcoverage. fqlint would need to be skipped.

a-frantz: Ok I agree we can leave it required for now. We can make those changes when the use-case comes up.

a-frantz (approved):

PR #34, opened 2020-12-22
feat: add subsampling option

a-frantz:
Untested and only added to rnaseq-standard for now.
@jsunny23 this could help make the CI faster if you use a decent size test BAM.


workflows/rnaseq/rnaseq-standard.wdl
```diff
@@ -57,26 +58,46 @@ workflow rnaseq_standard {
...
+
+        call util.get_read_groups as get_subsampled_read_groups { input: bam=subsample.sampled_bam, max_retries=max_retries }
+        call b2fq.bam_to_fastqs as subsampled_bam_to_fastqs { input: bam=subsample.sampled_bam, max_retries=max_retries, detect_nproc=detect_nproc }
+    }
+    if (subsample_n_reads < 1) {
+        call util.get_read_groups { input: bam=input_bam, max_retries=max_retries }
+        call b2fq.bam_to_fastqs { input: bam=input_bam, max_retries=max_retries, detect_nproc=detect_nproc }
+    }
+
+    String read_groups = select_first([get_read_groups.out, get_subsampled_read_groups.out])
```
adthrasher, line 90: Could we do select_first on  the subsampled bam to avoid having two different calls to get read groups and bam_to_fastqs? Essentially select_first([subsample.sampled_bam, input_bam]), assuming it would wait for the result of subsampling rather than just using the input_bam value.

a-frantz: Yes that's a much cleaner method. Good catch

a-frantz:
The runs haven't finished, but my tests with and without subsampling have made it to alignment, so don't anticipate issues.

Didn't test tSNE itself, but I'm satisfied with a miniwdl check for those changes, IMO

adthrasher (approved):

workflows/rnaseq/rnaseq-standard.wdl
```diff
@@ -57,19 +58,32 @@ workflow rnaseq_standard {
         stardb_tar_gz: "Database of reference files for the STAR aligner. Can be generated by `rnaseq-star-db-build.wdl`"
         strandedness: "empty, 'Stranded-Reverse', 'Stranded-Forward', or 'Unstranded'. If missing, will be inferred"
         output_prefix: "Prefix for output files"
+        subsample_n_reads: "Only process a random sampling of `n` reads. `-1` for processing entire input BAM."
```
claymcleod, line 61: More accurately, any number `>0`

claymcleod (approved):
Looks good, have you tested this? Should get with Alex to see how the results look on, say, the Clinical Pilot dataset.

adthrasher:
> Looks good, have you tested this? Should get with Alex to see how the results look on, say, the Clinical Pilot dataset.

We are asking the clinical team to test it ad hoc. David suggested that Maria try various cutoffs. I suggest that we just use their result.

a-frantz:
> Looks good, have you tested this? Should get with Alex to see how the results look on, say, the Clinical Pilot dataset.

It gets through subsampling and alignment without a problem. The validity of the results is going to depend on the cutoff, which will ultimately be up to the user.

We'll have to wait and see what the clinical team comes up with in terms of reasonable values we can suggest in the docs.

a-frantz:
Something I just noticed, my tests with and without subsampling took the same amount of time (~day and a half) to run due to waiting on nodes for the alignment step. I think people were reporting difficulty getting nodes last week, but in the future if the point of this option is to cut run time down users will need to scale back the default requirements of the alignment step so unnecessary node acquisition doesn't become a block (or a cost overrun in the case of asking for too large an instance in the cloud).

adthrasher:
> Something I just noticed, my tests with and without subsampling took the same amount of time (~day and a half) to run due to waiting on nodes for the alignment step. I think people were reporting difficulty getting nodes last week, but in the future if the point of this option is to cut run time down users will need to scale back the default requirements of the alignment step so unnecessary node acquisition doesn't become a block (or a cost overrun in the case of asking for too large an instance in the cloud).

I think the alignment step is dominated by the STAR memory requirements related to the genome size. My experience is that even using very small bams, STAR requires significant memory to store the reference data structure.

adthrasher:
Now that we've seen the results of the clinical team's testing, I think we should be good to merge and move forward with rolling this out.

PR #37, opened 2021-02-01
Add xenograft cleansing support for RNA-Seq V2

adthrasher:
Adding optional step to run XenoCP on sample after the initial alignment

### DIFF
--- a/workflows/rnaseq/rnaseq-standard-fastq.wdl
+++ b/workflows/rnaseq/rnaseq-standard-fastq.wdl
@@ -94,7 +94,7 @@ workflow rnaseq_standard_fastq {
 
     if (cleanse_xenograft){
         File contam_db = select_first([contaminant_stardb_tar_gz, ""])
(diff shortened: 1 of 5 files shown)


workflows/rnaseq/rnaseq-standard-fastq.wdl
```diff
@@ -49,6 +50,8 @@ workflow rnaseq_standard_fastq {
         Int max_retries = 1
         Boolean detect_nproc = false
         Boolean validate_input = true
+        Boolean cleanse_xenograft = false
+        File? contaminant_stardb_tar_gz
```
a-frantz, line 54: I went to check if `rnaseq-star-db-build.wdl` would be compatible with building this, and looks like it is, except right now the workflow assumes gencode.v31 and GRCh38_no_alt filenames. Not sure this PR is the right place to address it, but I'd like that to be generalized. You need to do task level overrides to change the filenames, which isn't intuitive (and potentially breaks? I think I remember the WDL spec discouraging overriding defaults provided at the `call` level)

a-frantz (approved):

a-frantz:
The `docker-branch` error was fixed on `master` and can be ignored. This looks good besides that

PR #38, opened 2021-02-05
refactor: use samtools quickcheck to do initial validation

a-frantz:
For large BAMs `picard validate` can take multiple hours, preventing any other tools from running. This adds `samtools quickcheck` to do a precursory validation before other tools run as opposed to waiting for the complete `picard validate`.

### DIFF
--- a/tools/samtools.wdl
+++ b/tools/samtools.wdl
@@ -24,6 +24,10 @@ task quickcheck {
         maxRetries: max_retries
     }
 
+    output {
+        File checked_bam = bam
+    }
+
     meta {
         author: "Andrew Thrasher, Andrew Frantz"
         email: "andrew.thrasher@stjude.org, andrew.frantz@stjude.org"
--- a/tools/multiqc.wdl
+++ b/tools/multiqc.wdl
@@ -86,7 +86,6 @@ task multiqc {
     }
 
     parameter_meta {
-        bam: "The BAM referenced by other input reports"
         validate_sam_file: "A file output from Picard's ValidateSam tool"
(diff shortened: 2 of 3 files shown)


tools/samtools.wdl
```diff
@@ -24,6 +24,10 @@ task quickcheck {
         maxRetries: max_retries
     }
 
+    output {
+        File checked_bam = bam
```
adthrasher, line 28: I see that you're now depending on the output of this for downstream steps. I think I prefer to allow the downstream steps to run on the input BAM and just have quickcheck fail if it detects issues. It does mean potential for wasted downstream computation in the case of an error, but in other cases we get a head start. Granted quickcheck should be quick, so it may not matter either way.

adthrasher: In the RNA-Seq V2 and bam_to_fastq pipelines, we don't use the output of quickcheck or Picard ValidateSam, we just run it s a failable step to catch errors.

a-frantz: The rational is that half a dozen jobs get launched at once, some that can run for multiple hours. If a user inputs a bad BAM, that adds up to a lot of wasted computation they might be paying for if they don't kill the jobs.

In RNA-Seq V2 and bam_to_fastq, the amount of potentially wasted computation is much smaller, so it isn't much of a concern.

That was the rationale for initially requiring a wait on `picard ValidateSam`, but that can take a few hours to run on large BAMs. The validation before launching a bunch of jobs isn't worth the huge increase in runtime.

I think `samtools quickcheck` is a nice middle ground.

adthrasher (approved):

PR #40, opened 2021-02-09
feat: add sequencErr to QC

a-frantz:


a-frantz:
In light of how large the output files are, I'm not sure we want to pull this into master without making it optional (default `False`?). Does anyone disagree or have other thoughts?

a-frantz:
@adthrasher @claymcleod this is passing my tests, ready to be pulled in with approval.

Except @claymcleod the Docker build CI still isn't working. Have you added the GH secret for it? I've just been pushing the SequencErr images manually.

docker/sequencErr/1.0.0/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM registry.access.redhat.com/ubi7/python-36 AS builder
```
a-frantz, line 1: It seems `seqeuncerr` doesn't build correctly on Debian systems, which is why I'm using Red Hat instead of `conda-base` here. I told Pandurang and he said they've only tested on the cluster. But it works with Red Hat as the base image.

adthrasher (approved):

docker/sequencErr/1.0.0/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+ARG GITHUB_TOKEN
+RUN git clone --depth 1 --branch v2.0.7 https://${GITHUB_TOKEN}@github.com/stjude/sequencErr.git
```
adthrasher, line 4: I noticed this is failing. Are we missing the GITHUB_TOKEN secret?

a-frantz: I don't think Clay has added it yet. I've pushed the images manually.

adthrasher: @claymcleod - Can you add GITHUB_TOKEN to this repo so that it can clone the SequenErr project?

PR #41, opened 2021-02-18
refactor(fastq_screen): remove need for specifically named DB

a-frantz:


a-frantz:
This is really just a style change. Because of how FastQ Screen needs to be configured, we need to know the path to the reference DBs ahead of time. Before we required a specifically named parent directory (after being untarred). This PR removes the necessity for a specifically named parent directory by enforcing no parent directory. I think it's (slightly) less error-prone. As long as the user obtains the DB through our `make-qc-reference.wdl` it will work.

Regardless, the DB paths are hardcoded so a user made DB directory will most likely fail. The FastQ Screen config system is terrible.

adthrasher (approved):

PR #42, opened 2021-02-18
refactor: remove specifications that GTF must be from Gencode

a-frantz:


a-frantz:
Ensembl GTF went through QC and rnaseq-standard without problems (except for the mismatched references causing a small issue in `ngsderive`). I don't see a reason this repo should be specifying Gencode. We've made it clear in our other repos/docs that **we** use/recommend Gencode, but IMO this repo should be agnostic.

PR #48, opened 2021-05-27
docs: update README and workflow/tool headers

a-frantz:


a-frantz:
I decided not to version any of the `build_<reference>` workflows, or the `calc-gene-lengths` workflow. I versioned everything else, but @adthrasher I wasn't sure what to put for the classification workflow. And I wasn't sure if there was a specification for that lying around somewhere.

I also said everything was "In Production" except the ESTIMATE workflow. Not sure if that's what we want to say for all of them.

adthrasher:
We should probably deprecate/remove the classification workflow. I doubt that it still works given the changes that have been made to the underlying tools.

README.md
```diff
@@ -43,19 +43,25 @@ The repository is laid out as follows:
 
 * `bin` - Scripts used by Cromwell configuration settings. Add this to `$PATH` prior to using configurations  in `conf` with Cromwell.
 * `conf` - Cromwell configuration files created for various environments that we use across our team. Feel free to use/fork/suggest improvements.
-* `docker` - Dockerfiles used in our workflows. All docker images are published to [Docker Hub](https://hub.docker.com/u/stjudecloud).
-  * At the time of writing, there is only one Docker image heavily in use, which is our [bioinformatics base image](./docker/bioinformatics-base/Dockerfile).
+* `docker` - Dockerfiles used in our workflows. All docker images are published to [Docker Hub](https://hub.docker.com/u/stjudecloud) as a part of our CI and are versioned.
```
adthrasher, line 46: The department guidance is moving to use GitHub packages rather than Docker hub. We may want to consider doing the same.

a-frantz: Is there a reason for preferring GitHub packages? Because we initially used GitHub packages till we learned releases couldn't be deleted or overwritten. The way our CI works, we create branch specific images which would be permanent in GH, but removable in Docker hub.

adthrasher: One of the primary motivating factors is the pull rate limitation. In Docker hub, the limit is applied to the user, not the repository. In GitHub packages, the transfers are limited at the repository level and are unlimited for public repos.

a-frantz: That makes a lot of sense. I'll add switching to GH packages on Trello

a-frantz:
> We should probably deprecate/remove the classification workflow. I doubt that it still works given the changes that have been made to the underlying tools.

Did a WDL implementation migrate to another repo or is it just deprecated?

Is there any intention of making it functional again? If not, I'd vote for removing the whole directory. If we might bring it up to date at some point, a deprecation notice in the README might be better.

README.md
```diff
@@ -25,7 +25,7 @@ expect that the workflows will work just as well using other runners.
 The easiest way to get started is to install [bioconda][bioconda] and the run the following commands:
 
 ```bash
-conda create -n workflows-dev -c conda-forge cromwell==47 -y
+conda create -n workflows-dev -c conda-forge cromwell miniwdl -y
```
jsunny23, line 28: Is there a reason to recommend installing both cromwell and miniwdl?

a-frantz: It's more of a suggestion. I find if you're doing any development you want to be able to run `miniwdl check`.

adthrasher:
> > We should probably deprecate/remove the classification workflow. I doubt that it still works given the changes that have been made to the underlying tools.
> 
> Did a WDL implementation migrate to another repo or is it just deprecated?
> 
> Is there any intention of making it functional again? If not, I'd vote for removing the whole directory. If we might bring it up to date at some point, a deprecation notice in the README might be better.

We ended up going with a native DX app and I currently have no intention to make the WDL functional again. It is fairly specific to the DX environment. So we should remove the entire directory.

README.md
jsunny23: Personally, I wouldn't include it since we don't reference `miniwdl` anywhere else in the repository. But I'm fine to leave it in if we're planning on adding documentation around using `miniwdl` to check workflows.

a-frantz:
I removed our statement about versioning, and set RNA-Seq Standard at v2.0.0 and QC at v1.0.0. I think this is suitable to be on `master` while we figure out the specifics of how we'll version things in the future.

PR #50, opened 2021-06-09
fix: Remove read_string() outputs

a-frantz:


a-frantz:
@adthrasher can you comment on #48 about whether we should keep the RNA-Seq expression workflow? The read_string() removal breaks how the `file_prefix` is used, and if we're about to remove the workflow I don't want to spend the time rewriting that usage. It's a non-trivial fix because WDL doesn't offer an easy way to convert `Array[File]` to `Array[String]`.

PR #52, opened 2021-06-30
fix: Picard ValidateSamFile refactor

a-frantz:


a-frantz:
Turns out there was a bug in MultiQC that would've crashed any samples that made it through with Picard failures. It was fixed in the latest patch release, but even if this new behavior was enabled earlier, we wouldn't have been able to pass any samples with errors.

tools/picard.wdl
```diff
@@ -81,22 +78,25 @@ task validate_bam {
...
             ~{stringency_arg} \
-            MAX_OUTPUT=100000 \
+            ~{ignore_prefix}~{sep=' IGNORE=' ignore_list} \
+            MAX_OUTPUT=~{max_errors} \
             > ~{output_filename}
 
+        rc=$?
+        if [ $rc -le -1 ] || [ $rc -ge 4 ]; then
+            exit $rc
+        fi
```
adthrasher, line 90: Not sure if you found the docs, but here are the documented return codes: 

> Return codes depend on the errors/warnings discovered:
> -1 failed to complete execution 0 ran successfully 1 warnings but no errors 2 errors and warnings 3 errors but no warnings

https://gatk.broadinstitute.org/hc/en-us/articles/360037594251-ValidateSamFile-Picard-

a-frantz: I was looking at a different set of docs that didn't have that (because naturally they have multiple differing sets of docs), but I found that information through [code diving](https://github.com/broadinstitute/picard/blob/master/src/main/java/picard/sam/ValidateSamFile.java#L113). I also saw it return `255` during one of my tests.

adthrasher (approved):

PR #55, opened 2021-11-05
feat: add QC summary

a-frantz:


adthrasher (changes requested):
Please add a meta section to the `qc_summary` task. I think that will help any external users understand why/what we are cherry-picking from the multiqc report.

a-frantz:
> Please add a meta section to the `qc_summary` task. I think that will help any external users understand why/what we are cherry-picking from the multiqc report.

Is my latest commit what you had in mind? I wanted to avoid mentioning how we use it internally (i.e. "we parse these results and store them in a database." I think that would be the only comment of its sort in the repository)

adthrasher:
It is. I might add that we are pulling stats from qualimap and ngsderive currently,  just for completeness.

adthrasher (approved):

a-frantz:
@rahul0620

PR #57, opened 2021-11-23
feat: bgzip check

a-frantz:


tools/util.wdl
```diff
@@ -260,4 +260,24 @@ task qc_summary {
...
+    input {
+        File bam
+        Int max_retries = 1
+    }
+
+    Float bam_size = size(bam, "GiB")
+    Int disk_size = ceil(bam_size + 10)
+
+    command {
+        bgzip -t ~{bam}
```
adthrasher, line 275: I’m assuming that bgzip -t exits non-zero if there is an issue. Do we want the pipeline to fail or should it just report the issue? Also do we want to capture the bgzip error message? I’m not sure we can do anything with it other than regenerate the BAM.

a-frantz: We definitely want it to fail, because if there's a problem with the compression QualiMap and FastQC both hang instead of failing. It really confused me when I first ran into this bad sample, so an explicit failure is preferred.

Since we want it to fail, there's no better way to report the bgzip error other than just saying "check the stderr logs".

Ideally this task will fail/complete before either of those start execution, but that's not enforced ATM. (quickcheck can finish before `compression_integrity`, then Qualimap and FastQC get submitted before the compression check finishes) So those tasks might need to be manually killed depending on the order of execution, which is an admittedly annoying manual step that might go unnoticed for a day or two.

We could enforce this runs first, but I don't want to add another bottleneck for such a rare edge case. We've seen it twice over thousands of samples? I think the "wait on quickcheck before expensive operations" is good enough.

adthrasher (approved):

PR #58, opened 2021-12-09
fix: replace all underscore suffixes with dot suffixes

a-frantz:
This should satisfy the assumptions of Plaid @zaeleus


adthrasher (approved):
What was the motivation behind this change?

a-frantz:
> What was the motivation behind this change?

Plaid assumes a period separates all the "pieces" (sample name, sequencing type, file type extension) of a filename, but the current naming conventions of these QC files uses an underscore before the extension. Switching those to periods satisfies the Plaid assumptions.

As far as I'm aware, there was no real rhyme or reason for using underscores in these places, so I don't think it makes a difference outside of Plaid.

zaeleus:
No, this doesn't affect internal tooling, as no assumptions are made in relation to this. My past feedback was only a comment on the inconsistent output naming.

a-frantz:
> No, this doesn't affect internal tooling, as no assumptions are made in relation to this. My past feedback was only a comment on the inconsistent output naming.

O ok. My mistake then.
@adthrasher do you have any opinion one way or the other? I think they're completely interchangeable, but if there's no reason to change it (beyond a bit of consistency), it's probably best to leave it as is.

PR #61, opened 2022-04-08
Add optional check for VCF file integrity using bgzip -t

adthrasher:


a-frantz:
I'm hesitant to add an input file that goes through one check without producing output... To an external user of the workflow, it looks completely unrelated to the rest of the workflow.

I understand the need for performing this check as part of our broader pipeline; is there another place we can do it? This doesn't look like the best solution we can come up with...

Alternatively, are there any MultiQC compatible tools that run on gVCFs? I'd be happier with adding this new input if it can sensibly add to the output report.

adthrasher:
It looks like MultiQC accepts VCFtools output. I don't know if there is anything there that would be useful at the QC step or not.

adthrasher:
@claymcleod , @dfinkels, @agout - Andrew raises a good point about performing more than just a file integrity check on the gVCF files. MultiQC supports the `relatedness2` statistic from VCFtools as well as several `TsTv` statistics. The relatedness check could be useful for tumor/normal projects as it would likely detect cases where there was unexpected relatedness which could help screen for sample swapping. Do you think that is worth pursuing or alternatively something with the TsTv statistics? Is there some other VCF QC check that we should be considering?

claymcleod:
We should absolutely include the checks you mention above. Good thinking Andrew and Alex.

PR #67, opened 2022-09-30
[WIP] Integrate PEAseq paired-end mode for ChIP-Seq

adthrasher:
SEAseq has a new mode known as PEAseq. It can now handle paired-end sequencing data properly. This integrates that option into our ChIP-Seq workflow.

### DIFF
--- a/workflows/chipseq/chipseq-standard.wdl
+++ b/workflows/chipseq/chipseq-standard.wdl
@@ -102,22 +102,19 @@ workflow chipseq_standard {
         }
     }
 
-    # if (pairing == "Paired-end"){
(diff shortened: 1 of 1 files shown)


adthrasher:
The final Picard ValidateSam fails with "Mate not found for paired read" and "Both mates are marked as second of pair" errors. This appears to be a result of this code (https://github.com/stjude/seaseq/blob/2fabb66c714412eb224007e5ab519fe06d1cf636/workflows/tasks/samtools.wdl#L80-L92) in SEAseq that manipulates the reads manually. I have reached out to Modupe to clarify what this is meant to do.

adthrasher:
Superseded by #111

PR #70, opened 2022-12-29
ci updates

a-frantz:
This PR started to address 2 issues:
1) the CI on `master` is currently failing.
2) back in 2021 I opened issue #56 and wanted to see that finally addressed. #56 is concerned with tasks pulling images that are "out of date" and no longer being built by the CI. Presumably, there's always a good reason we stop building any particular image, so we only want to pull images which are "current". Note this doesn't mean "latest", as some images we maintain multiple versions. Most significantly this applies to the `util` images, where each version bump comes with more dependencies (and a larger size). So we want to maintain multiple versions and treat them all as valid options.

From there, I also updated the CI to enforce URL WDL imports. This is really just for consistency, because from my testing it looks like relative imports are handled appropriately in all the runners I checked. I'd be fine dropping this specific change.

Part of the reason the CI was failing on `master`, is because we pull from SEAseq `master`, which currently has syntax `miniwdl check` doesn't like. I don't like our CI being dependent on the state of another repo. So me and @adthrasher decided it would be better if we enforce external repos to use tagged releases instead of `master`.

I then went through and addressed all the issues `miniwdl`+`shellcheck` complained about (and found some legitimate bugs in the process).


a-frantz:
The miniwdl-check CI is going to continue to fail till Modupe fixes the SeaSeq repo. But once those coercions are addressed and she replaces the relative imports in her repo that will start passing (and this repo should be more stable/less dependent on the state of external repos)

workflows/chipseq/chipseq-standard.wdl
```diff
@@ -152,7 +151,7 @@ task parse_input {
 
     runtime {
         disk: "1 GB"
-        docker: 'ghcr.io/stjudecloud/util:1.1.0'
+        docker: 'ghcr.io/stjudecloud/util:1.0.0'
```
adthrasher, line 154: Similar question here, why are we rolling back?

.github/workflows/lint-check.yml
```diff
@@ -33,9 +49,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - name: Ensure SemVer'd docker images are being pulled
+    - name: Ensure current SemVer'd docker images are being pulled
```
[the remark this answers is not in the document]
a-frantz: We have multiple "current" versions of some images, like `util`. In `util`'s case, the older images are smaller, so I kept them as valid. Newer versions have more dependencies and larger sizes, so only certain tasks need them.

Presumably if we stop building a version, there's a good reason for it and we want to force an upgrade (most likely it has some undesired behavior). At least that's my thinking.

workflows/chipseq/chipseq-standard.wdl
```diff
@@ -30,13 +30,12 @@
 version 1.0
 
 import "https://raw.githubusercontent.com/stjudecloud/workflows/master/workflows/general/bam-to-fastqs.wdl" as b2fq
-import "https://raw.githubusercontent.com/stjude/seaseq/master/workflows/workflows/mapping.wdl" as seaseq_map
-import "https://raw.githubusercontent.com/stjude/seaseq/master/workflows/tasks/seaseq_util.wdl" as seaseq_util
+import "https://raw.githubusercontent.com/stjude/seaseq/2.3/workflows/workflows/mapping.wdl" as seaseq_map
+import "https://raw.githubusercontent.com/stjude/seaseq/3.0/workflows/tasks/seaseq_util.wdl" as seaseq_util
```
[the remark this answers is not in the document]
a-frantz: Update, made this change and turns out "https://raw.githubusercontent.com/stjude/seaseq/2.3/workflows/tasks/seaseq_util.wdl" doesn't exist. So having them all on the same version won't work. We have to revert that change... How important to you is it that they all pull from the same version?

a-frantz:
I discovered this morning that "docker://ghcr.io/stjudecloud/fqlib:1.1.0" has been broken (but working in Cromwell until yesterday? Not sure what was going on there, haven't done an investigation). I opted to switch the docker import from the version we maintain to the version "BioContainers" maintains, and plan to entirely drop the `fqlib` Dockerfile from this repository. It's already maintained by Michael, I don't see any good reason to duplicate our work (and introduce errors) here.

This made me rethink our initial decision to build and maintain all these docker images that already exist in larger repositories. Are we sure we want to be doing this? I'm thinking we should switch as many of our docker images to officially maintained repos as we can. There are a few of our Docker images that alternatives don't exist for, but I don't see the rationale for maintaining tools like `samtools` and `picard` when those are available from larger orgs.

What was our original rationale for building everything we use manually (or through CI)? @claymcleod

workflows/chipseq/chipseq-standard.wdl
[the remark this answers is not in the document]
adthrasher: Ah, yes. Modupe renamed "util.wdl" to "seaseq_util.wdl" at some point. So we may need to keep the mixed versions for now.

a-frantz:
> Can you add some additional detail on this? I'm curious exactly what is happening here.

```
$ singularity exec --containall docker://quay.io/biocontainers/fq:0.9.1 pwd
FATAL:   Unable to handle docker://quay.io/biocontainers/fq:0.9.1 uri: failed to get checksum for docker://quay.io/biocontainers/fq:0.9.1: Error reading manifest 0.9.1 in quay.io/biocontainers/fq: manifest unknown: manifest unknown
```
```
$ singularity exec --containall docker://quay.io/biocontainers/samtools:latest pwd
FATAL:   Unable to handle docker://quay.io/biocontainers/samtools:latest uri: failed to get checksum for docker://quay.io/biocontainers/samtools:latest: Error reading manifest latest in quay.io/biocontainers/samtools: manifest unknown: manifest unknown
```
When I saw that error I did some googling, and apparently this is a [known issue](https://github.com/apptainer/singularity/issues/749) that Quay.io isn't supported in older version of singularity. The ticket says it's been fixed, and the version of singularity on the cluster is newer than the issue (v3.7.1-1.el7). So I'm not sure exactly what's happening. But every image from BioContainers I tried failed.

a-frantz:
> Interesting, in the first couple I tried, it worked as expected.
> 
> ```
> (base) singularity exec --containall docker://quay.io/biocontainers/bioconductor-rsamtools@sha256:0b5b5ad4aba860e0edd1a0a588f21b9e77ed5e8f7b8685a23a7134caa0840b92  pwd
> INFO:    Using cached SIF image
> WARNING: passwd file doesn't exist in container, not updating
> WARNING: group file doesn't exist in container, not updating
> /home/athrashe
> ```
> 
> I was able to run it on the head node and a compute node. My tests with Cromwell also succeeded.
> 
> I think you may need to check the tags. I see `quay.io/biocontainers/fq:0.9.1--h9ee0642_0`, but not without the suffix.

Ah ok. So the tags I expected just aren't there (and apparently there's no default tag if you leave one out). So nothing's wrong, I just made some bad assumptions.

Looks like we still can consider offloading our image maintenance to BioContainers where applicable.

docker/cellranger/1.1.0/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+FROM ubuntu:20.04
+
+RUN apt-get update && \
+    apt-get upgrade -y && \
+    apt-get install curl -y && \
+    rm -r /var/lib/apt/lists/*
+
+WORKDIR /opt
+
+ARG cellranger_url
```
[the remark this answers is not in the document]
adthrasher: OK, that wasn't my understanding, but I don't think it is important enough to hold this up.

adthrasher (approved):

docker/cellranger/1.1.0/Dockerfile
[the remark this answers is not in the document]
a-frantz: @adthrasher I haven't pushed a `cellranger:1.1.0` image. How do we want to handle which version of CellRanger to use? `1.0.0` uses cellranger `7.0.1`. The link we provide in the new Dockerfile only supplies whatever the latest version is. Manually push with `7.0.1` because that's what's been tested? Bump to latest?

claymcleod:
> What was our original rationale for building everything we use manually (or through CI)? @claymcleod

When we first started the workflows repo, there simply weren't that many official docker containers created. I'm in support of using official Docker containers if they exist.

a-frantz:
> > What was our original rationale for building everything we use manually (or through CI)? @claymcleod
> 
> When we first started the workflows repo, there simply weren't that many official docker containers created. I'm in support of using official Docker containers if they exist.

Great. I think this PR has grown large enough, so I'll open a new PR to address this specific issue.

tools/fq.wdl
```diff
@@ -26,7 +26,7 @@ task fqlint {
     runtime {
         disk: disk_size + " GB"
         memory: memory_gb + " GB"
-        docker: 'ghcr.io/stjudecloud/fqlib:1.0.1'
+        docker: 'ghcr.io/stjudecloud/fqlib:branch-ci-1.2.0'
```
claymcleod: This seems like something we don't want to merge in, right? Was this just done for testing?

workflows/chipseq/chipseq-standard.wdl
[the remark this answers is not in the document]
claymcleod: What's the issue with just using `https://raw.githubusercontent.com/stjude/seaseq/3.0/workflows/workflows/mapping.wdl`?

workflows/chipseq/chipseq-standard.wdl
claymcleod: In my opinion, we should use the latest here regardless of the size. I think this will be too difficult to keep track of why we did what otherwise.

workflows/rnaseq/rnaseq-standard-fastq.wdl
```diff
@@ -35,7 +35,7 @@ import "https://raw.githubusercontent.com/stjudecloud/workflows/master/tools/ngs
 import "https://raw.githubusercontent.com/stjudecloud/workflows/master/tools/htseq.wdl"
 import "https://raw.githubusercontent.com/stjudecloud/workflows/master/tools/samtools.wdl"
 import "https://raw.githubusercontent.com/stjudecloud/workflows/master/tools/deeptools.wdl"
-import "https://raw.githubusercontent.com/stjude/xenocp/master/wdl/workflows/xenocp.wdl" as xenocp_workflow
+import "https://raw.githubusercontent.com/stjude/XenoCP/3.1.4/wdl/workflows/xenocp.wdl" as xenocp_workflow
```
claymcleod, line 38: Could you just explicitly comment on why this is needed?

workflows/rnaseq/rnaseq-standard.wdl
```diff
@@ -41,7 +41,7 @@ import "https://raw.githubusercontent.com/stjudecloud/workflows/master/tools/hts
 import "https://raw.githubusercontent.com/stjudecloud/workflows/master/tools/samtools.wdl"
 import "https://raw.githubusercontent.com/stjudecloud/workflows/master/tools/util.wdl"
 import "https://raw.githubusercontent.com/stjudecloud/workflows/master/tools/deeptools.wdl"
-import "https://raw.githubusercontent.com/stjude/xenocp/master/wdl/workflows/xenocp.wdl" as xenocp_workflow
+import "https://raw.githubusercontent.com/stjude/XenoCP/3.1.4/wdl/workflows/xenocp.wdl" as xenocp_workflow
```
claymcleod, line 44: And this one?

docker/cellranger/1.1.0/Dockerfile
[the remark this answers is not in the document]
claymcleod: I agree with Andrew. I just moved these to the top and tried the docker build, and it worked fine for me.

Also, I don't think we want to be actually packaging cell ranger and distributing it via GitHub packages. @adthrasher have we looked at the licensing at all around this?

claymcleod (changes requested):

claymcleod:
Also, are we sure the cellranger image works? I just tried getting inside the Docker container and just running the command and got this:

```
$ docker run -it cellranger:latest
root@2d6a0b69114f:/opt# cellranger
qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
root@2d6a0b69114f:/opt#
```

tools/fq.wdl
a-frantz: Just for testing, already fixed.

workflows/rnaseq/rnaseq-standard-fastq.wdl
adthrasher: I'll let @a-frantz make the update, but we decided that we should depend on an explicit version for externally maintained tools. This was prompted by changes to SEAseq, where a dependency was renamed ("util.wdl" -> "seaseq_util.wdl") in their repository which left our main branch implementation broken.

workflows/chipseq/chipseq-standard.wdl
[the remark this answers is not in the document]
a-frantz: `v3.0` of SEAseq has some coercions that are considered fine by the spec and `dxWDL` (which is what the SEAseq team uses as their runner), but cause errors in `miniwdl`. I'd say this is a case of `miniwdl` misinterpreting the WDL spec, but regardless, it causes our CI to fail. Rolling back to "seaseq/2.3/workflows/workflows/mapping.wdl" fixes the problem of failing CI.

workflows/rnaseq/rnaseq-standard-fastq.wdl
a-frantz: XenoCP falls into a gray area... It's maintained by us, but is also an external repo. With SEAseq changes causing our CI to fail, me and @adthrasher decided external repos should point to a tagged release for improved stability. "External repo" would include XenoCP.

workflows/chipseq/chipseq-standard.wdl
a-frantz: I don't have a problem with bumping it to the latest. I'll make the change.

docker/cellranger/1.1.0/Dockerfile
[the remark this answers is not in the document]
adthrasher: Yeah, we should probably get their license reviewed. The package is set to internal for now. Even that may not technically be allowable based on the terms of the 10x license.

adthrasher:
> Also, are we sure the cellranger image works? I just tried getting inside the Docker container and just running the command and got this:
> 
> ```
> $ docker run -it cellranger:latest
> root@2d6a0b69114f:/opt# cellranger
> qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
> root@2d6a0b69114f:/opt#
> ```

Ah yes, I built the image on my mac (apple silicon) without a platform argument. Looking at the image, it's `arm64`.

```
        "Architecture": "arm64",
```

adthrasher:
> > Also, are we sure the cellranger image works? I just tried getting inside the Docker container and just running the command and got this:
> > ```
> > $ docker run -it cellranger:latest
> > root@2d6a0b69114f:/opt# cellranger
> > qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
> > root@2d6a0b69114f:/opt#
> > ```
> 
> Ah yes, I built the image on my mac (apple silicon) without a platform argument. Looking at the image, it's `arm64`.
> 
> ```
>         "Architecture": "arm64",
> ```

Rebuilding on a DX node: 

```
dnanexus@job-GP7zy709xbP3Xpk605K4q84g:~/workflows/docker/cellranger/1.1.0$ docker run -it cellranger:latest
root@1b344e701c87:/opt# cellranger
cellranger cellranger-7.1.0
Process 10x Genomics Gene Expression, Feature Barcode, and Immune Profiling data

USAGE:
    cellranger <SUBCOMMAND>

OPTIONS:
    -h, --help       Print help information
    -V, --version    Print version information

SUBCOMMANDS:
    count               Count gene expression (targeted or whole-transcriptome) and/or feature barcode reads from a single sample and GEM well
    multi               Analyze multiplexed data or combined gene expression/immune profiling/feature barcode data
    multi-template      Output a multi config CSV template
    vdj                 Assembles single-cell VDJ receptor sequences from 10x Immune Profiling libraries
    aggr                Aggregate data from multiple Cell Ranger runs
    reanalyze           Re-run secondary analysis (dimensionality reduction, clustering, etc)
    targeted-compare    Analyze targeted enrichment performance by comparing a targeted sample to its cognate parent WTA sample (used as input for targeted gene expression)
    targeted-depth      Estimate targeted read depth values (mean reads per cell) for a specified input parent WTA sample and a target panel CSV file
    mkvdjref            Prepare a reference for use with CellRanger VDJ
    mkfastq             Run Illumina demultiplexer on sample sheets that contain 10x-specific sample index sets
    testrun             Execute the 'count' pipeline on a small test dataset
    mat2csv             Convert a gene count matrix to CSV format
    mkref               Prepare a reference for use with 10x analysis software. Requires a GTF and FASTA
    mkgtf               Filter a GTF file by attribute prior to creating a 10x reference
    upload              Upload analysis logs to 10x Genomics support
    sitecheck           Collect linux system configuration information
    help                Print this message or the help of the given subcommand(s)

```

docker/cellranger/1.1.0/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+ARG cellranger_md5
+
+RUN curl -o cellranger.tar.gz \
+    ${cellranger_url} \
+    && echo "${cellranger_md5}  cellranger.tar.gz" > cellranger.tar.gz.md5 \
+    && md5sum -c cellranger.tar.gz.md5 \
+    && tar -xzvf cellranger.tar.gz \
+    && mv cellranger-* cellranger
+
+ENV PATH "/opt/cellranger:$PATH"
```
adthrasher, line 22: For consistency with our other images, we should add an entrypoint to this image.

a-frantz: Done. But I just poked around and not all our images have an ENTRYPOINT. Outside the scope of this PR to address that for other images? I don't like that it's not consistent, but don't think this PR is the place for that fix.

a-frantz:
@adthrasher we still need to push a `cellranger:1.1.0` image, and switch to using that version in our tasks. Do you think there would be any problems using `v7.1.0` instead of the version you packaged in `cellranger:1.0.0`? I guess the licensing concerns also need to addressed. Will those hold up merging this PR?

If so, I'd rather just drop the cellranger changes from this PR and move them to the PR I'll make later with other Docker changes.

workflows/rnaseq/rnaseq-standard-fastq.wdl
claymcleod: I think it's fine to include the explicit version then. I would consider this an external repo.

adthrasher:
> @adthrasher we still need to push a `cellranger:1.1.0` image, and switch to using that version in our tasks. Do you think there would be any problems using `v7.1.0` instead of the version you packaged in `cellranger:1.0.0`? I guess the licensing concerns also need to addressed. Will those hold up merging this PR?
> 
> If so, I'd rather just drop the cellranger changes from this PR and move them to the PR I'll make later with other Docker changes.

I am fine bumping to use 7.1.0. The image is currently set to `internal`, which should be OK and enable us to use this workflow internally. That is a bit of a departure from our normal operations, but we may have no choice.

docker/cellranger/1.1.0/Dockerfile
adthrasher: Thanks, there may be some images (`util` comes to mind) where an entrypoint does not make sense. But otherwise, I think we should include them. Agree that this should be a separate PR.

claymcleod (approved):

PR #73, opened 2023-01-23
Name sort

a-frantz:
I was hoping `picard sort` would be faster than Qualimap's sorting, but no noticeable difference in speed. However this did remove the transient failure I was seeing earlier (at least removed for my test cohort of 81 samples. I wouldn't be shocked if it comes back with a larger pool of samples). This change will get us a slight improvement in stability, but nothing earth shattering.

### DIFF
--- a/tools/qualimap.wdl
+++ b/tools/qualimap.wdl
@@ -65,12 +65,13 @@ task rnaseq {
     input {
         File bam
         File gtf
-        Int memory_gb = 16
-        Int? disk_size_gb
-        Int max_retries = 1
+        Boolean name_sorted = false
         Boolean paired_end = false
         String provided_strandedness = ""
         String inferred_strandedness = ""
+        Int memory_gb = 16
+        Int? disk_size_gb
+        Int max_retries = 1
     }
 
     String out_directory = basename(bam, ".bam") + ".qualimap_rnaseq_results"
@@ -86,6 +87,7 @@ task rnaseq {
                         if (inferred_strandedness == "Unstranded") then "non-strand-specific" else
                         "unknown-strand" # this will intentionally cause qualimap to error. You will need to manually specify
                                          # in this case
+    String name_sorted_arg = if (name_sorted) then "-s" else ""
     String paired_end_arg = if (paired_end) then "-pe" else ""
 
     Int java_heap_size = ceil(memory_gb * 0.9)
@@ -105,6 +107,7 @@ task rnaseq {
                         -outdir ~{out_directory} \
                         -oc qualimap_counts.txt \
                         -p ~{stranded} \
+                        ~{name_sorted_arg} \
                         ~{paired_end_arg} \
                         --java-mem-size=~{java_heap_size}G
         rm "$gtf_name"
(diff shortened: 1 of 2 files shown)


adthrasher:
Do you know how `qualimap` does its name sorting? I ask because I believe that `samtools` and `picard` produce different results when name sorting. If `qualimap` is expecting sorted data that doesn't match its own definition of "sorted" then we may encounter odd failures in a subset of samples.

a-frantz:
That's very interesting. I did not check for that. However 81 samples went through without any errors. Should I run some different cohorts and see what happens?

a-frantz:
Can you point me to some more information about this difference in sorting? Why would it not just be alphanumeric read name for everything?

a-frantz:
Wow. Sorting is something that there should only be one correct answer for. I'm shocked, but this is familiar now that I've done some reading. I had forgotten about this issue.

The Qualimap docs don't say anything about their sorting algorithm other than that it's "by name". I'd assume that to mean a simple lexical sorting like Picard's.

Is 81 test cases sufficient? How do we proceed?

adthrasher:
The SAM spec also includes this footnote:

>6
It is known that widely used software libraries have differing definitions of the queryname sort order, meaning care should be taken when operating on multiple files of varying provenance. Tools may wish to use the sub-sort field to explicitly distinguish between natural and lexicographical ordering. See Section 1.3.1

Section 1.3.1 then defines three sub sorts, `lexicographical`, `natural`, and `umi`. I believe Picard is `lexicographical` and samtools is `natural`.

>1.3.1 Defined sub-sort terms
While the SS sub-sort field allows implementation-defined keywords, some terms are predefined with specific
meanings.
lexicographical sort order is defined as a character-based dictionary sort with the character order as
defined by the POSIX C locale. For example “abc”, “abc17”, “abc5”, “abc59” and “abcd” are in
lexicographical order.
natural sort order is similar to lexicographical order except that runs of adjacent digits are considered to
be numbers embedded within the text string, ordered numerically when compared to each other and
ordered as single digits when compared to the surrounding non-digit characters. Runs that differ only
in the number of leading zeros (thus are numerically tied) are ordered by more-zeros coming before
fewer-zeros. The characters ‘-’ and ‘.’ are considered as ordinary characters, so apparently negative or
fractional values are not treated as part of an embedded number. For example, “abc”, “abc+5”, “abc5”, “abc.d”, “abc03”, “abc5”, “abc008”, “abc08”, “abc8”, “abc17”, “abc17.+”, “abc17.2”, “abc17.d”, “abc59” and “abcd” are in natural order.
umi is a lexicographical sort by the UMI tag. The MI tag should be used for comparing UMIs. The RX tag
may be used in its absence but is not guaranteed to be unique across multiple libraries.

adthrasher:
> Wow. Sorting is something that there should only be one correct answer for. I'm shocked, but this is familiar now that I've done some reading. I had forgotten about this issue.
> 
> The Qualimap docs don't say anything about their sorting algorithm other than that it's "by name". I'd assume that to mean a simple lexical sorting like Picard's.
> 
> Is 81 test cases sufficient? How do we proceed?

Based on the little information `qualimap` provides, I think it just needs the read pairs sorted together for its paired analysis mode. If that's the case, the overall ordering shouldn't matter. Combining that with the test cohort result, I think it's fine to proceed.

adthrasher (approved):

claymcleod:
Agree. Just FYI, I would assume if Andrew T's assertion were true that `samtools collate` would also be sufficient. Handy command to know about.

PR #74, opened 2023-01-25
Docker refactor and code cleanup

a-frantz:
Switch to using officially maintained images from BioContainers where applicable.

After implementing the Docker refactor, I began thoroughly testing each workflow, and cleaning up code as I went along.

Apologies for the bad practice of not separating out my PRs, but this made the manual testing significantly easier.

**Changes included:**
* deprecating many of our self-maintained Docker images in favor of those in BioContainers
* more consistency in parameters between similar workflows
* bug fixes
* standardizing defaults
* adding example input JSONs for each workflow
* whitespace changes
* applying `detect_nproc` logic consistently for all multi-core processes
* exposing more parameters to be tweaked by users
* more/better documentation
* optional md5 check to `download` task
* ensuring `set -euo pipefail` is set for all multi-command tasks
* softcap of 100 characters per-line in WDL
* Group all "make reference" WFs under `general/`
* add logical ordering to lists/JSONs that were unordered


a-frantz:
Here's my rationale for what I've kept:
* **STAR**: we package it with the custom `sort_star_input.py` script. Could we instead make running that script a full fledged WDL task and remove this image?
* **bwa**: Our tasks depend on BOTH `bwa` and `samtools` being installed. Not aware of any external images with both (could maybe find a "kitchen sink" image, but I think I'd prefer to maintain a targeted, and small, image)
* **cellranger**: not available elsewhere
* **conda-base**: a dependency of other images I've kept. We could probably find a replacement base maintained elsewhere though? Worth investigating?
* **estimate**: I think this one can probably go..... All it does is install one R package, which could be reasonably done by the task itself. Unless we don't want tasks installing things? Iffy because it's R....
* **fastq_screen**: hardcoded config file packaged into the image. Boy would I love to see this one go, a really ugly solution, but I don't see a viable alternative
* **gtfparse**: I just found this on [Quay.io](http://Quay.io), so will be replaced soon. Not sure how I missed it the first go around
* **multiqc**: packaged with a config file. Is there a better way to do this? We can specify those configs on the CL, but with the recent refactor of the `multiqc` task, I don't think we should hardcode them.... It would be a tricky argument to write to the task if we wanted to take that approach, and it could get ugly fast...
* **sequencErr**: not available elsewhere
* **util**: has what we need and nothing else. Won't be able to find an external image with these dependencies that isn't monstrously huge

adthrasher:
> * **STAR**: we package it with the custom `sort_star_input.py` script. Could we instead make running that script a full fledged WDL task and remove this image?

We could, but that requires a decent increase in file system usage. Should we consider simplifying our Dockerfile to something like:
```
FROM biocontainers/star:version

COPY ./sort_star_input.py /home

ENTRYPOINT [ "STAR" ]
```

Maybe not ideal, but we're then using the public image as the base.

> * **bwa**: Our tasks depend on BOTH `bwa` and `samtools` being installed. Not aware of any external images with both (could maybe find a "kitchen sink" image, but I think I'd prefer to maintain a targeted, and small, image)

We're only using `samtools` to convert SAM->BAM correct? If so we could split the task, though that introduces more storage utilization.

> * **cellranger**: not available elsewhere

That image is currently internal-only. I have initiated a review with legal services which will inform any long-term plans.

> * **conda-base**: a dependency of other images I've kept. We could probably find a replacement base maintained elsewhere though? Worth investigating?

So that's just `miniconda` and we update packages with `apt` and explicitly add additional channels. Should we just use `miniconda` as our base and the individual builds can specify their channels?

> * **estimate**: I think this one can probably go..... All it does is install one R package, which could be reasonably done by the task itself. Unless we don't want tasks installing things? Iffy because it's R....

Are we using that anywhere?

> * **fastq_screen**: hardcoded config file packaged into the image. Boy would I love to see this one go, a really ugly solution, but I don't see a viable alternative

Seems reasonable

> * **gtfparse**: I just found this on [Quay.io](http://Quay.io), so will be replaced soon. Not sure how I missed it the first go around

OK

> * **multiqc**: packaged with a config file. Is there a better way to do this? We can specify those configs on the CL, but with the recent refactor of the `multiqc` task, I don't think we should hardcode them.... It would be a tricky argument to write to the task if we wanted to take that approach, and it could get ugly fast...

How bad would it be to make the yaml file an input to the task? Then we would always supply the same one, but other users could add their own version.

> * **sequencErr**: not available elsewhere

I would like to drop this all together. We don't have any use for the outputs. Their GitHub isn't public. I don't think there is any commitment to ongoing development or maintenance. It seems like a good candidate to deprecate.

> * **util**: has what we need and nothing else. Won't be able to find an external image with these dependencies that isn't monstrously huge

Agreed

a-frantz:
> > * **STAR**: we package it with the custom `sort_star_input.py` script. Could we instead make running that script a full fledged WDL task and remove this image?
> 
> We could, but that requires a decent increase in file system usage. Should we consider simplifying our Dockerfile to something like:
> 
> ```
> FROM biocontainers/star:version
> 
> COPY ./sort_star_input.py /home
> 
> ENTRYPOINT [ "STAR" ]
> ```
> 
> Maybe not ideal, but we're then using the public image as the base.

I agree that's a better solution.

> > * **bwa**: Our tasks depend on BOTH `bwa` and `samtools` being installed. Not aware of any external images with both (could maybe find a "kitchen sink" image, but I think I'd prefer to maintain a targeted, and small, image)
> 
> We're only using `samtools` to convert SAM->BAM correct? If so we could split the task, though that introduces more storage utilization.

Passing around an uncompressed SAM seems like something we really want to avoid. Someone running this on the cloud could rack up a huge bill. I think maintaining this image is a fair price to pay to avoid that.

> > * **conda-base**: a dependency of other images I've kept. We could probably find a replacement base maintained elsewhere though? Worth investigating?
> 
> So that's just `miniconda` and we update packages with `apt` and explicitly add additional channels. Should we just use `miniconda` as our base and the individual builds can specify their channels?

I think so. An extra line in each file is ok if it gets rid of a whole image.

> > * **estimate**: I think this one can probably go..... All it does is install one R package, which could be reasonably done by the task itself. Unless we don't want tasks installing things? Iffy because it's R....
> 
> Are we using that anywhere?

Used in [this workflow](https://github.com/stjudecloud/workflows/blob/master/workflows/rnaseq/ESTIMATE.wdl). I wrote it a long time ago, ran it on a few cohorts, and I think the conclusion was "ETIMATE isn't a very good tool (at least for our needs)". But I don't think it hurts to keep the workflow around?

> 
> > * **multiqc**: packaged with a config file. Is there a better way to do this? We can specify those configs on the CL, but with the recent refactor of the `multiqc` task, I don't think we should hardcode them.... It would be a tricky argument to write to the task if we wanted to take that approach, and it could get ugly fast...
> 
> How bad would it be to make the yaml file an input to the task? Then we would always supply the same one, but other users could add their own version.

Hmmm...... I don't think that's a good idea.... Mainly because the QC workflow has certain naming conventions which we like, and MultiQC doesn't. So the QC workflow needs to specify these `extra_fn_clean_exts` when MultiQC is called. I don't like outsourcing that to an input, when it's _always_ the same for `quality-check-standard`.

I think the best way to handle this would be to add `Array[String] extra_fn_clean_exts = []`. Then QC would pass in `[".qualimap_bamqc_results", etc.]`. Those could either be written to a YAML or passed on the CL (depending on which looks better).

> 
> > * **sequencErr**: not available elsewhere
> 
> I would like to drop this all together. We don't have any use for the outputs. Their GitHub isn't public. I don't think there is any commitment to ongoing development or maintenance. It seems like a good candidate to deprecate.
> 

Agreed, it should be dropped in this PR.

a-frantz:
ping @adthrasher re:

>
> > > * **estimate**: I think this one can probably go..... All it does is install one R package, which could be reasonably done by the task itself. Unless we don't want tasks installing things? Iffy because it's R....
> >
> > Are we using that anywhere?

> Used in [this workflow](https://github.com/stjudecloud/workflows/blob/master/workflows/rnaseq/ESTIMATE.wdl). I wrote it a long time ago, ran it on a few cohorts, and I think the conclusion was "ETIMATE isn't a very good tool (at least for our needs)". But I don't think it hurts to keep the workflow around?

I'd prefer to keep the workflow unless you have a good justification for removing it. If we decide to keep it, then this Docker image needs to be addressed. Normally I don't want tasks installing things into the container, but since it's R, I feel like that can be relaxed a little. Do you think it's ok to move that one install line into the task itself, and just run inside a vanilla R container?

adthrasher:
> ping @adthrasher re:
> 
> > > > * **estimate**: I think this one can probably go..... All it does is install one R package, which could be reasonably done by the task itself. Unless we don't want tasks installing things? Iffy because it's R....
> > > 
> > > 
> > > Are we using that anywhere?
> 
> > Used in [this workflow](https://github.com/stjudecloud/workflows/blob/master/workflows/rnaseq/ESTIMATE.wdl). I wrote it a long time ago, ran it on a few cohorts, and I think the conclusion was "ETIMATE isn't a very good tool (at least for our needs)". But I don't think it hurts to keep the workflow around?
> 
> I'd prefer to keep the workflow unless you have a good justification for removing it. If we decide to keep it, then this Docker image needs to be addressed. Normally I don't want tasks installing things into the container, but since it's R, I feel like that can be relaxed a little. Do you think it's ok to move that one install line into the task itself, and just run inside a vanilla R container?

I would just keep the existing Docker image if we're keeping the workflow. I wouldn't push the package install into the task.

a-frantz:
This is passing in `quality-check-standard` without a problem, so I think it's in a pretty solid state. However as a sanity check I'm planning on testing every workflow we have in the repo before merging. Pretty substantial changes in here that I believe effect every task/workflow we've written.

adthrasher:
> This is passing in `quality-check-standard` without a problem, so I think it's in a pretty solid state. However as a sanity check I'm planning on testing every workflow we have in the repo before merging. Pretty substantial changes in here that I believe effect every task/workflow we've written.

Since you're doing this already, could you make sure that the inputs specification is stored in a shared location. That should make future testing simpler as we'll already have a set of test cases.

PR #78, opened 2023-02-28
Cell Ranger updates

adthrasher:
Remove sample_id parameter. This is required to match the fastq file name, so always infer it.


a-frantz (approved):

a-frantz:
Why did you remove the option to override the sample name? Won't this break if a user supplies an unexpected filename?

(it also broke the `scrnaseq-standard` call to this task)

adthrasher:
> Why did you remove the option to override the sample name? Won't this break if a user supplies an unexpected filename?
> 
> (it also broke the `scrnaseq-standard` call to this task)

The `sample` parameter to `cellranger count` must match the sample name in the fastq file. So rather than create potential issues, we should just read it from the fastq files that we're providing to `cellranger count`.

a-frantz:
> > Why did you remove the option to override the sample name? Won't this break if a user supplies an unexpected filename?
> > (it also broke the `scrnaseq-standard` call to this task)
> 
> The `sample` parameter to `cellranger count` must match the sample name in the fastq file. So rather than create potential issues, we should just read it from the fastq files that we're providing to `cellranger count`.

Oof. That's not ideal. Good catch though, I'm ok with removing that option if it could lead to conflicts. But can you add a comment somewhere explaining that rationale? (Otherwise I guarantee somewhere down the line I'll look at this task and go "sample_id should be customizable!" lol)

a-frantz (approved):

PR #80, opened 2023-03-02
Stage changes for main rename

adthrasher:


.github/workflows/lint-check.yml
```diff
@@ -20,25 +20,7 @@ jobs:
...
-          if [ -n "$bad_lines" ]; then
-            >&2 echo "Imports from this repo must use the master branch!"
-            >&2 echo "The following lines are bad:"
-            >&2 echo "$bad_lines"
-            >&2 echo ""
-            EXITCODE=1
-          fi
-          
-          bad_lines=$(echo "$import_lines" | awk '!/https:\/\/raw.githubusercontent.com\/stjudecloud\/workflows/ && /master/') || true
+          bad_lines=$(echo "$import_lines" | awk '!/https:\/\/raw.githubusercontent.com\/stjudecloud\/workflows/ && (/main/ || /master/)') || true
```
a-frantz, line 23:
```suggestion
          bad_lines=$(echo "$import_lines" | awk '/http/ && (/main/ || /master/)') || true
```
Didn't actually test this awk pattern, but it looks like what we want. The current first half of the pattern we expect to be a no-op if we switch to relative imports.

This has me thinking..... if we switch to relative imports, than we should have a check that all `http` imports are to an external repo (i.e. not this one). I think the current check, plus the check I just suggested, should cover every case we want to exclude?

a-frantz (changes requested):
Why don't we go ahead and switch the imports over to relative in this PR. It's either that or update them all to point to `main`

.github/workflows/lint-check.yml
adthrasher: OK. I added a check to ensure nothing uses this repository's URL. I then updated the subsequent check to just look for `http` + `master` or `main` as you suggested above.

adthrasher:
> Why don't we go ahead and switch the imports over to relative in this PR. It's either that or update them all to point to `main`

Sounds good. I was going to ask if you had that in another branch or not.

a-frantz (approved):

PR #83, opened 2023-03-07
Kraken2

a-frantz:


adthrasher (approved):

docker/kraken2/1.0.0/rsync_from_ncbi.pl
```diff
@@ -0,0 +1,199 @@
...
+        $md5 = dirname($md5) . "/" . "md5checksums.txt";
+        last if $ftp->get($md5);
+        warn "$PROG: unable to download $_ on try $try of $ntries: ".$ftp->message()."\n";
+        last if $try == $ntries;
+        sleep $sleepsecs;
+        $sleepsecs *= 3;
+    }
+    die "$PROG: unable to download ftp://${SERVER}${SERVER_PATH}/$_\n" if $try == $ntries;
+    $ftp->quit;
+    system("grep '" . basename($_) . "' md5checksums.txt > md5.txt && md5sum -s -c md5.txt") == 0
```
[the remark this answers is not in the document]
a-frantz: That's fair. I'd be fine with it either way (assuming you've tested the changes). I did have one or two errors relating to this block while constructing the DB, so improvements would be welcome. Up to you whether they get committed or not.

workflows/qc/quality-check-standard.wdl
```diff
@@ -93,24 +91,23 @@ workflow quality_check {
     call fqc.fastqc { input: bam=quickcheck.checked_bam, max_retries=max_retries }
     call ngsderive.instrument as ngsderive_instrument { input: bam=quickcheck.checked_bam, max_retries=max_retries }
     call ngsderive.read_length as ngsderive_read_length { input: bam=quickcheck.checked_bam, bai=bam_index, max_retries=max_retries }
-    
     call ngsderive.encoding as ngsderive_encoding { input: ngs_files=[quickcheck.checked_bam], prefix=prefix, max_retries=max_retries }
-    String parsed_encoding = read_string(ngsderive_encoding.inferred_encoding)
+
+    call picard.bam_to_fastq { input: bam=quickcheck.checked_bam, max_retries=max_retries }
```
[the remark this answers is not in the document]
adthrasher: My thought was that we don't particularly care to get a comprehensive assessment of contamination. Below a certain percentage, it really doesn't matter for our purposes. So I would think that subsampling might be sufficient. We would have to think through what level of contamination is problematic and what level of subsampling would be needed to ensure we would find that threshold.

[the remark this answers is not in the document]
a-frantz: It's not immediately clear to me what kind of performance boost we'd see. Kraken doesn't take that long to run (I've been impressed with its speed). I think `bam_to_fastq` might take longer. And subsampling isn't exactly a fast operation. Are we sure this potential speed boost is worth an investigation? If I recall correctly, (which I might not), the longest running step is FastQC (it bothers me that they have "fast" in the name of something so slow lol). So implementing subsampling for Kraken might not result in any speed bump at all.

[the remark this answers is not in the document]
adthrasher: OK. Let's leave it out unless we come up with a need for it.

workflows/qc/quality-check-standard.wdl
```diff
@@ -93,24 +91,23 @@ workflow quality_check {
...
+    call fq.fqlint { input: read1=bam_to_fastq.read1, read2=bam_to_fastq.read2, max_retries=max_retries }
```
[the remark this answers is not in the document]
adthrasher: I agree, I see no reason for us to validate that bam+fastq are the same sample. That's a user error. I don't know if we should hold things up to implement an option like this either. Just a thought for the future, particularly as we start running in real-time rather than on historical cohorts (for which we've likely deleted the fastq files anyway).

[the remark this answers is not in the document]
adthrasher: I also don't think that we output the fastq files from RNA-Seq v2. For WGS/WES, I don't think that `msgen` does either. So it would require some changes there as well.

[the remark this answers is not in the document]
a-frantz: Agreed then. An idea to keep in mind, but not worth implementing now.

tools/kraken.wdl
```diff
@@ -0,0 +1,300 @@
...
+        Int minimizer_spaces = 7
+        Int max_db_size_gb = -1
+        Float load_factor = 0.7
+        Boolean clean_up = true
+        Int added_memory_gb = 0
+        Int added_disk_size_gb = 0
+        Int ncpu = 1
+        Boolean detect_nproc = false
+        Int max_retries = 1
+    }
```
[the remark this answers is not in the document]
a-frantz: Will do, good call. Although I don't think we need the `clean_up` option anymore. I added it in case I needed to debug something, but I never wound up looking at those files. I think it should always be safe to clean out the files after the tar is complete.

tools/kraken.wdl
```diff
@@ -0,0 +1,336 @@
...
+
+    parameter_meta {
+        library_name: "One of `archaea`, `bacteria`, `plasmid`, `viral`, `human`, `fungi`, `plant`, `protozoa`, `nt`, `UniVec`, or `UniVec_Core`"
+        db_name: "Name for output tarball. The suffix `.tar.gz` will be added."
+        memory_gb: "RAM to allocate for task"
+        added_disk_size_gb: "Additional disk space to allocate for task. Default disk size is determined dynamically based on `library_name`."
+        max_retries: "Number of times to retry in case of failure"
+    }
+
+    Int disk_size_gb = (if library_name=="bacteria" then 300 else 20) + added_disk_size_gb
```
a-frantz, line 71: I say in the meta that `nt` is a valid option to this task, but if `disk` gets used by the backend, this won't have enough disk space allocated. I believe `nt` requires over 2TB of space (over 1TB for the uncompressed library, library gets duplicated during repeat masking).

Should we support it in this clause? I think anybody running on a backend that allocates and charges for disk space would want to avoid using `nt`. Seems like a use case we could ignore? But that is a tad sloppy...

docker/kraken2/1.0.0/rsync_from_ncbi.pl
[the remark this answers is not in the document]
adthrasher: I think we should omit. I would want to test it more thoroughly if we were to include it.

tools/kraken.wdl
adthrasher: I would just note in the parameter meta where you mention `nt` that you would need a corresponding change to `added_disk_size_gb` to account for `nt`. Then it is up to the user to figure out how much to use and they should understand the charges when requesting that much disk.

tools/kraken.wdl
```diff
@@ -0,0 +1,336 @@
...
+        db_name: "Name for output tarball. The suffix `.tar.gz` will be added."
+        kmer_len: "K-mer length in bp"
+        minimizer_len: "Minimizer length in bp"
+        minimizer_spaces: "Number of characters in minimizer that are ignored in comparisons"
+        max_db_size_gb: "Maximum number of GBs for Kraken 2 hash table; if the estimator determines more would normally be needed, the reference library will be downsampled to fit."
+        load_factor: "Proportion of the hash table to be populated (must be between 0 and 1)"
+        added_memory_gb: "Additional RAM to allocate for task. Default RAM is allocated dynamically based on the database size."
+        added_disk_size_gb: "Additional disk space to allocate for task. Default disk size is determined dynamically based on `tarballs` size."
+        ncpu: "Number of cores to allocate for task"
+        detect_nproc: "Use all available cores"
```
adthrasher, line 192: We may want to globally update this with a second sentence like "This is useful in systems where a node meeting minimum requirements is allocated (e.g. cloud providers), rather than a fixed number of cores (e.g. cluster environments". That feels very wordy, especially to repeat everywhere, but makes the intention clear.

tools/kraken.wdl
```diff
@@ -0,0 +1,336 @@
...
+        Boolean detect_nproc = false
+        Int max_retries = 1
+    }
+
+    parameter_meta {
+        read1: "Gzipped FastQ file with 1st reads in pair"
+        read2: "Gzipped FastQ file with 2nd reads in pair"
+        db: "Kraken2 database. Can be generated with `make-qc-reference.wdl`. Must be a flat tarball without a root directory."
+        sample_name: "Name for sample. If missing will be inferred by removing the suffix '_R1.fastq.gz' from the `read1` filename."
+        store_sequences: "Store and output main Kraken2 output instead of just the summary report"
```
adthrasher, line 275: Instead or in addition to? Looking below, I think it is an additional output.
```suggestion
        store_sequences: "Store and output main Kraken2 output in addition to the summary report"
```

tools/kraken.wdl
adthrasher: Maybe just a short option like "Recommended for cloud environments"

workflows/qc/make-qc-reference.wdl
```diff
@@ -25,12 +25,62 @@
...
+            "protozoa",
+            "UniVec_Core"
+        ]
+        Array[String] fasta_urls = []
+        Array[File] fastas = []
+        Int? max_retries
+    }
+
+    parameter_meta {
+        kraken_libraries: "List of kraken libraries to download. Valid options: `archaea`, `bacteria`, `plasmid`, `viral`, `human`, `fungi`, `plant`, `protozoa`, `nt`, `UniVec`, `UniVec_Core`"
```
adthrasher, line 49: Should we just link to the manual? https://github.com/DerrickWood/kraken2/blob/master/docs/MANUAL.markdown#custom-databases

workflows/qc/make-qc-reference.wdl
```diff
@@ -25,12 +25,62 @@
...
+        fasta_urls: "URLs for NCBI formatted FASTAs to download"
```
adthrasher, line 50:
```suggestion
        fasta_urls: "URLs for any additional FASTA files in NCBI format to download and include in the database. This allows the addition of individual genomes (or other sequences) of interest."
```

tools/kraken.wdl
```diff
@@ -0,0 +1,339 @@
...
+        Int max_retries = 3
+    }
+
+    String db_name = "kraken2_"+library_name+"_library"
+
+    parameter_meta {
+        library_name: {
+            help: "Library to download. Note that due to the large size of the `nt` library, the `added_disk_size_gb` parameter might have to be used, depending on your backend. This option was not tested, but we estimate the size required to be around 2TB."
+            choices: ['archaea', 'bacteria', 'plasmid', 'viral', 'human', 'fungi', 'plant', 'protozoa', 'nt', 'UniVec', 'UniVec_Core']
+        }
```
adthrasher, line 68: I think this is much cleaner to read. Assuming that it valid in the spec and well handled by various tools, then I think we should use this format.

tools/kraken.wdl
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    parameter_meta {
+        tarballs: "Tarballs containing the kraken taxonomy and at least one library. Tarballs must not have a root directory."
+        db_name: "Name for output tarball. The suffix `.tar.gz` will be added."
+        kmer_len: "K-mer length in bp"
+        minimizer_len: "Minimizer length in bp"
+        minimizer_spaces: "Number of characters in minimizer that are ignored in comparisons"
+        max_db_size_gb: "Maximum number of GBs for Kraken 2 hash table; if the estimator determines more would normally be needed, the reference library will be downsampled to fit."
+        load_factor: "Proportion of the hash table to be populated (must be between 0 and 1)"
```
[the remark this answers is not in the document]
a-frantz: I just copy and pasted from the Kraken `--help`. I have no idea what it does tbh, but saw it as an option and exposed it. The manual will hopefully have more information. I'll check and update this line once I'm done with the rest of your feedback.

If the manual doesn't make it clear, it might be best to just remove the option. I'm hesitant to expose something we don't understand, and I don't think it's worth the effort of a codedive.

workflows/qc/make-qc-reference.wdl
a-frantz: I think the manual could be linked the in the header. But I don't think it belongs in the parameter_meta

tools/kraken.wdl
a-frantz:
```suggestion
        detect_nproc: "Use all available cores. Recommended for cloud environments. Not recommended for cluster environments."
```
?

tools/kraken.wdl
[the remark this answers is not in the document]
adthrasher: I don't see any discussion in the manual about it. I only see a reference to it in the paper. The help string says only 

>   --load-factor FRAC         Proportion of the hash table to be populated
>                              (build task only; def: $DEF_LOAD_FACTOR, must be
>                              between 0 and 1).

A bit of googling and I think we should probably just remove the parameter. It impacts the speed of retrieval in the hash and 0.75 is a common choice. Given that Kraken 2 uses 0.70 for its standard databases, I think it is reasonable to just use it and not expose the parameter.

workflows/qc/make-qc-reference.wdl
adthrasher: Should this also use the `choices` format?

tools/kraken.wdl
adthrasher: dxCompiler (https://github.com/dnanexus/dxCompiler) replaced it. `description` is an alias for `help` in dxCompiler, so I'd suggest we use that here.

PR #85, opened 2023-03-10
Coverage analysis

a-frantz:
* new workflow inputs:
    * optional array of BED files to analyze coverage for. Accompanied by an optional array (of equal length) of strings to label those regions. Without a label array, default labels of `regions1`, `regions2`, etc. will be used.
        * note: internally, and in the examples in the documentation, we will use the BEDs/labels of `exon` and `CDS`. But we will explicitly allow users to use their own arbitrary BEDs/regions
* use new reference BEDs with `mosdepth`
* stop running `picard CollectWgsMetrics` (will probably keep the task in the repo, in the unlikely case we ever chose to revisit this decision)
* simplify `experiment` param to just `DNA` or `RNA` (would "experiment" still be the right word for this param? Would "molecule" or "sequence_type" be better? Just an idea I'm throwing out there, no strong feelings from me)
* include new workflow for generating all the QC reference files, including a task that creates "exon" and "CDS" BEDs

~~Caveat: We can't get all the mosdepth reports into any multiqc reports (either the per-sample ones, or the aggregate ones I pass to David). This is a future problem to deal with. In the meantime, I will only pass the "whole genome" coverage report to multiqc. This file will be guaranteed to always be generated, while the others are "optional", so makes the most sense to be consistent.~~
It is possible to display these how we want in MultiQC and that has been implemented.


tools/mosdepth.wdl
```diff
@@ -1,13 +1,57 @@
 ## # Mosdepth
 ##
-## This WDL tool wraps the [mosdepth tool](https://github.com/brentp/mosdepth).
+## This WDL file wraps the [mosdepth tool](https://github.com/brentp/mosdepth).
+## It also creates reference BEDs for restricting mosdepth coverage analysis to only exonic or coding regions.
 
 version 1.0
 
+task make_coverage_regions_beds {
```
adthrasher, line 8: It feels a little odd to have this in the `mosdepth` tool. I don't have a better vision for it, other than potentially in `util`. It also uses a different docker image than `mosdepth`.

workflows/qc/quality-check-standard.wdl
```diff
@@ -58,9 +60,11 @@ workflow quality_check {
         bam: "Input BAM format file to quality check"
         bam_index: "BAM index file corresponding to the input BAM"
         reference_fasta: "Reference genome in FASTA format"
+        coverage_beds: "An array of 3 column BEDs which are passed to the `-b` flag of mosdepth, in order to restrict coverage analysis to select regions"
+        coverage_labels: "An array of equal length to `coverage_beds` which determines the prefix label applied to the output files. If omitted, defaults of `regions1`, `regions2`, etc. will be used."
         gtf: "GTF features file. **Required** for RNA-Seq data"
         star_log: "Log file generated by the RNA-Seq aligner STAR"
-        experiment: "'WGS', 'WES', or 'RNA-Seq'"
+        experiment: "'DNA' or 'RNA'"
```
adthrasher, line 67: I don't have any useful thoughts on this. I agree that `experiment` is no longer truly correct. Maybe `molecule`? Or `nucleic_acid`?

workflows/qc/quality-check-standard.wdl
```diff
@@ -97,12 +103,24 @@ workflow quality_check {
     call ngsderive.encoding as ngsderive_encoding { input: ngs_files=[quickcheck.checked_bam], prefix=prefix, max_retries=max_retries }
     String parsed_encoding = read_string(ngsderive_encoding.inferred_encoding)
 
-    if (experiment == "WGS") {
-        call picard.collect_wgs_metrics { input: bam=quickcheck.checked_bam, reference_fasta=reference_fasta, max_retries=max_retries }
-        call mosdepth.coverage { input: bam=quickcheck.checked_bam, bai=bam_index, max_retries=max_retries }
+    call mosdepth.coverage as wg_coverage {
+        input:
+            bam=quickcheck.checked_bam,
```
adthrasher, line 108: Does this need to be `quickcheck.checked_bam` or should it just be the workflow input? I suppose `quickcheck` is quick and might help cut down on wasted compute. But we don't necessarily need to wait for quickcheck to complete to start queueing the remaining tasks.

workflows/qc/quality-check-standard.wdl
```diff
@@ -97,12 +103,24 @@ workflow quality_check {
...
+            bai=bam_index,
+            max_retries=max_retries
+    }
+    scatter(coverage_pair in zip(coverage_beds, parse_input.labels)) {
+        call mosdepth.coverage as regions_coverage {
+            input:
+                bam=quickcheck.checked_bam,
```
adthrasher, line 115: Same question as above.

adthrasher (approved):

tools/mosdepth.wdl
a-frantz: Yea, maybe the best place for it is `util`, but that's getting awfully long. I can move it there for now, but I think part of my `code-cleanup` branch will be splitting `util` into some smaller files. Don't have an exact vision for it yet, but I'd agree it belongs in `util` for now.

workflows/qc/quality-check-standard.wdl
a-frantz: I like `molecule` the best, but don't love it. Maybe @claymcleod has thoughts on a better word?

workflows/qc/quality-check-standard.wdl
a-frantz: The rationale (which I'm willing to reconsider):
`quickcheck` is "quick", so we should validate the input BAM before queuing long running (and potentially expensive) tasks that would fail if `quickcheck` failed. I don't think overall runtime is a valid concern with how fast quickcheck is. So I'd like to keep it this way. Unless you have some concern about readability or something else I haven't thought of (maybe this decision warrants a comment somewhere in the code?)

workflows/qc/quality-check-standard.wdl
claymcleod: I like `molecule` myself. I'm good with this change in general.

PR #86, opened 2023-03-13
Code cleanup

a-frantz:
Continuation of [this PR](https://github.com/stjudecloud/workflows/pull/74) with new branch name since scope changed.

Switch to using officially maintained images from BioContainers where applicable.

After implementing the Docker refactor, I began thoroughly testing each workflow, and cleaning up code as I went along.

Apologies for the bad practice of not separating out my PRs, but this made the manual testing significantly easier. (Another good motivation to finally implement CI testing, so behemoth PRs like this don't happen again)

**Changes included:**
* deprecating many of our self-maintained Docker images in favor of those in BioContainers
* more consistency in parameters between similar workflows
* bug fixes
* standardizing defaults
* adding example input JSONs for each workflow
* whitespace changes
* applying `use_all_cores` logic consistently for all multi-core processes
* exposing more parameters to be tweaked by users
* more/better documentation
* ensuring `set -euo pipefail` is set for all multi-command tasks
* softcap of 90 characters per-line in WDL
* Group all "make reference" WFs under `general/`
* add logical ordering to lists/JSONs that were unordered


tools/bwa.wdl
```diff
@@ -30,23 +30,29 @@ task bwa_aln {
             n_cores=$(nproc)
         fi
 
-        mkdir bwa
+        mkdir /tmp/bwa
 
-        tar -C bwa -xzf ~{bwadb_tar_gz}
-        PREFIX=$(basename bwa/*.ann ".ann")
+        tar -C /tmp/bwa -xzf ~{bwadb_tar_gz}
```
adthrasher, line 35: This change would be inconsistent with the star task. I'm also not sure this is a good approach. There is no guarantee that `/tmp` in a container is mounted to a location with sufficient storage. In our Singularity config, we set `--workdir` to the task directory. So there `/tmp` would be a subdirectory of the main task directory.

tools/bwa.wdl
```diff
@@ -30,23 +30,29 @@ task bwa_aln {
...
+        PREFIX=$(basename /tmp/bwa/*.ann ".ann")
 
-        bwa aln -t "${n_cores}" bwa/"$PREFIX" ~{fastq} > sai
+        bwa aln -t "$n_cores" /tmp/bwa/"$PREFIX" ~{fastq} > sai
```
adthrasher, line 38: I don't object to this, but I do slightly prefer the `${}` syntax for bash variables inside of a WDL command block. To me, that clarifies that it is a bash variable reference and not a WDL variable.

tools/bwa.wdl
a-frantz: I could've sworn I read somewhere a long time ago that it was a WDL best practice to put large files in `/tmp`, but I just did some googling and can't find any reference to that practice. Don't remember the rationale for that statement, so can no longer justify this change.

I think some other tasks with large references I've written I've put them in `/tmp`, but if `/tmp` might not have sufficient storage based on backend, I think I should go through and change all those to use CWD.

workflows/reference/inputs/bwa-db-build-inputs.json
```diff
@@ -0,0 +1,7 @@
+{
+    "bwa_db_build.reference_fa_url": "https://ftp.ncbi.nlm.nih.gov/genomes/all/GCA/000/001/405/GCA_000001405.15_GRCh38/seqs_for_alignment_pipelines.ucsc_ids/GCA_000001405.15_GRCh38_no_alt_analysis_set.fna.gz",
+    "bwa_db_build.reference_fa_name": "GRCh38_no_alt.fa.gz",
+    "bwa_db_build.reference_fa_md5": "a08035b6a6e31780e96a34008ff21bd6",
+    "bwa_db_build.build_bwa_db.bwadb_dir_name": "GRCh38_no_alt.bwadb",
```
adthrasher, line 5:
```suggestion
    "bwa_db_build.build_bwa_db.bwadb_dir_name": "GRCh38_no_alt.bwa_db",
```
Not necessary, but I think this improves readability.

workflows/reference/inputs/star-db-build-inputs.json
```diff
@@ -5,6 +5,7 @@
     "star_db_build.gtf_url": "ftp://ftp.ebi.ac.uk/pub/databases/gencode/Gencode_human/release_31/gencode.v31.annotation.gtf.gz",
     "star_db_build.gtf_name": "gencode.v31.gtf.gz",
     "star_db_build.gt5_md5": "0aac86e8a6c9e5fa5afe2a286325da81",
+    "star_db_build.build_star_db.stardb_dir_name": "GRCh38_no_alt.stardb",
```
adthrasher, line 8:
```suggestion
    "star_db_build.build_star_db.stardb_dir_name": "GRCh38_no_alt.star_db",
```

tools/ngsderive.wdl
```diff
@@ -1,6 +1,6 @@
 ## # ngsderive
 ##
-## This WDL tool wraps the [ngsderive package](https://github.com/stjudecloud/ngsderive)
+## This WDL file wraps the [ngsderive package](https://github.com/stjudecloud/ngsderive)
```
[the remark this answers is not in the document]
a-frantz: Yes I agree that's a problem, but IMO we shouldn't be adding to this already behemoth PR lol. I was planning on getting this merged "as is", and then continuing work on this branch to conform to our style guide and best practices.

adthrasher (approved):

PR #89, opened 2023-04-14
docs: add template tasks to guide style

a-frantz:
This is my proposed style template for **all** of our tasks. Once this gets agreed upon and merged, I will use this as a guide for enforcing consistent style in #86 . I believe that I have covered all the standard conventions* we want to use in the repo; if I missed any or we add a new enforceable convention in the future, this template will have to be updated. However it's intended to be a "write it once" file.

*the `wget` task is a known exception. That will have to be special cased and will follow it's own convention.

Proposed rules: see `style-guide.md`

--------------------

*Outstanding questions:*
- do we want to enforce anything for the header block?
    - I don't think we need licensing in every file, but we have it for some.
    - We've licensed the repo in general, so should the licensing information present in some files be removed? It seems redundant.
    - What other information should be in the header and what do we want to enforce?
    - At the very least, I want a top-level markdown header/title present for all our WDL files. (`## # title`)
        - when CI converts these files to web pages (with `wdldoc`), any pages without top level titles will look sloppy.
- I haven't proposed rules that apply to workflows, other than WDL lines being capped at 100 characters wide and the `Int? max_retries` convention
    - are there conventions we want to enforce for workflows?
    - I have a few in mind, but I'm not sure the best way to document them. I don't think templating like I did for tasks will be very helpful.

**Note:** due to lack of tools, "enforcement" of these rules will have to be manual during code review. Not ideal, but it is what it is.


template/common-parameter-meta.txt
```diff
@@ -0,0 +1,36 @@
...
+        disk_size_gb: "Disk space to allocate for task"
+        modify_memory_gb: "Add to or subtract from dynamic memory allocation. Default memory is determined by the size of the inputs."
+        modify_disk_size_gb: "Add to or subtract from dynamic disk space allocation. Default disk size is determined by the size of the inputs."
+        ncpu: "Number of cores to allocate for task"
+        detect_nproc: "Use all available cores? Recommended for cloud environments. Not recommended for cluster environments."
+        max_retries: "<for a task>Number of times to retry in case of failure"
+        max_retries: "<for a workflow>Number of times to retry failed steps. Overrides task level defaults."
+        outfile_name: "Name for the <type of file> file"
+        prefix: "Prefix for the <type of file> file. The extension `<extension>` will be added."
+        gtf: "GTF features file"
```
a-frantz, line 11: Right now, we use `gtf` everywhere when some tasks would technically allow a GFF. GTF is a stricter more specific form of GFF. Do we want to make that distinction? Or keep telling users "use a more specific file format"? Potentially more maintenance; to my knowledge we've never tested any of our workflows/task with a GFF, and we'd want to perform that sanity check before making a change to the documentation.

The easy option is keep specifying GTF format, if a user wants to use another type of GFF for whatever reason, they can figure it out themselves through trial and error.
That seems lazy and a bad user experience, so I'd suggest documenting (and testing) a wider set of formats.

adthrasher: I don't have strong preferences on the file formats. That said, most (all?) of the major annotation sources provide annotations in both GTF and GFF format. So I would not go out of our way to ensure GFF compatibility. We would also need to be specific if we mean GFF3 or GFF2.

a-frantz: I agree that we shouldn't put in much work to ensure compatibility with the GFF spec. *But* if a task is currently able to handle GFF2 or GFF3 (without us having to make code changes), that should be documented. Only proposing doc changes. Good point on v2/v3 compatibility. Will need some testing.

a-frantz: Another thought on GTFs: we've discussed this through slack, but bringing it up on the public PR.

In some places we assume a gzipped GTF, and in other places we accept either compressed or uncompressed. Those cases should have separate documentation. Or better yet, we just universally update our tasks to accept either compressed or uncompressed GTFs.
I'm hesitant to remove a feature, but I'd also consider dropping support for uncompressed reference files so we can be consistent (this would also be the easiest solution, we wouldn't have to think about the uncompressed case anymore). I wouldn't actually advocate for taking that approach, but thought it was worth mentioning as an option.

jsunny23:
This is good! I think having a style guide for these workflows is a great idea as they make writing these workflows easier. A few questions/comments I had:

1. What are your impressions on other WDL style guides out there [1](https://biowdl.github.io/styleGuidelines.html) [2](https://broadinstitute.github.io/long-read-pipelines/development_guide/wdl_style_guide/) ? The BioWDL guide seems lower level but a pretty good standard for WDL files. If we can totally agree with it, it would be nice to mention it in this style guide as the standard for things like indentation, expressions, line length, etc.

2. On that note, I see they prefer using the `{ }` syntax for `command` blocks and I tend to agree with that. Using `<<< >>>` for all commands seems like we're using brackets for 90% of the file, with 10% (all `command` blocks) using the arrow notation, rather than using brackets for 98% of the files, with 2% of the commands needing to use the arrow notation. The % values were just made up, but my opinion is that it's easier to think of using brackets all the time, unless the bash script needs otherwise.

3. For the memory and disk allocation section, can tasks have both a static and dynamic allocations set for a particular resource? I assume not, but it would be good to make it clear in the style guide.

4. I suggest moving the "Not allowed" parameter names to the style guide or demarcate it within the common parameters file. To me, the `common-parameter-meta` file would be where I go to find the standard name and description of an input in my workflow but the "Not allowed" names makes it difficult to quickly pick out the input meta I want to use. And it won't be an extensive list of disallowed names, so what makes it to the file seems pretty specific to certain previous naming choices.

5. Personal opinion: If a task expects a Gzipped version of a file only, then the name should contain `_gz` to represent that. For example, `reads_one_fastq_gz`. I also think it's good to use compressed versions of file wherever possible, so making it clear to the user what the options are, is helpful.

6. Similarly, I think we should try to use updated versions of file formats as much as we can. So I would suggest using GFF3 as our defaults, instead of GTF. Michael M. is an expert on this so I would defer to his opinion if GFF3 isn't better than GTF. 

7. "All tasks should run in a Docker container". Are there any recommendations on what container to use if the task only uses basic linux commands? The BioWDL style guide prefers using containers and specifically Biocontainers, but allows non-containerized tasks.

8. I think for line length, it should be enforced, as much as possible, across the file. So, the command block as well.

9. Re: Header - I think if we should keep the license at the repo level, but I'm not sure if there is a specific need for licensing each file. What is the difference in documentation between the header and the `meta.description`? I think if we want to use `wdldoc` for all the tasks then the meta block seems redundant. Imo, if we're enforcing `wdldoc` documentation for every task, then we don't need the meta block.

template/common-parameter-meta.txt
```diff
@@ -0,0 +1,37 @@
+        memory_gb: "RAM to allocate for task"
+        disk_size_gb: "Disk space to allocate for task"
+        modify_memory_gb: "Add to or subtract from dynamic memory allocation. Default memory is determined by the size of the inputs."
+        modify_disk_size_gb: "Add to or subtract from dynamic disk space allocation. Default disk size is determined by the size of the inputs."
+        ncpu: "Number of cores to allocate for task"
+        detect_nproc: "Use all available cores? Recommended for cloud environments. Not recommended for cluster environments."
+        max_retries: "<for a task>Number of times to retry in case of failure"
+        max_retries: "<for a workflow>Number of times to retry failed steps. Overrides task level defaults."
```
jsunny23, line 8:
```suggestion
        max_retries: "<for a task> Number of times to retry in case of failure"
        max_retries: "<for a workflow> Number of times to retry failed steps. Overrides task level defaults."
```

a-frantz:
Thanks for the feedback Jobin! Much appreciated. GH is acting up and not letting me properly "quote reply", so my response's formatting will be simplified.

> 1.

I like certain pieces of both of them, and similarly dislike pieces. But my main problem is that both of them lack enough details and I'd like to be more specific than either of them. I would consider using one of them as a "base" upon which we build. I hadn't thought of doing that before right now, but I actually think it's a really good idea. I'm not sure how much traction either of these style guides has, but in theory, if we pitch this as an optional "extension" of a popular style guide, we could see wider adoption than my current plan which is to enter the race as a competitor (for lack of better phrasing).

On the other hand, do we really care if other people follow our style guide? Probably not. It would be neat, but ultimately our focus should be something *we* like, for *this* repo. And I have some disagreements with both BioWDL and Broad...

Thanks for bringing this up, I'm struggling to come to a decision on the best approach here. Would welcome other thoughts on the idea of using BioWDL or Broad's style guide as a base we add rules to.

> 2.

I see where you're coming from, but I disagree. Mainly because I can't remember what Bash constructs cause problems haha. If I remember, correctly it was something that I wouldn't **expect** to cause problems. I remember having a difficult time debugging a task and the problem ended up being the bracket notation. I'd rather avoid the problem entirely (and not waste time debugging some weird case) by simply always using the HEREDOC style which never causes an issue. But maybe I'll get outvoted on this, I won't raise a stink if other people agree with you.

> 3.

I'm not sure I follow what you're asking, but that's a good argument that I need to make that section of the guide clearer. I'll try to answer anyway:
A task can have a dynamic disk space allocation and a static memory allocation (and all other such combinations are allowed).
This isn't written in the guide, but I intended for it to be disallowed to provide a way to users/contributors to mix for the same resource. i.e. a task cannot have both `memory_gb` and `modify_memory_gb`. We could definitely come up with an implementation that would allow that construct, but I think it would be very confusing to users (and anyone looking at the code). I'll update the docs to explicitly disallow this convention.

> 4.

I agree with you. I'll make an update.

> 5.

I agree on this as well. We still need to address the fact that we are inconsistent with our assumptions about the compression of parameters. But in the case that *only* gzipped files are accepted, I'd agree with adding a `_gz` suffix to the variable name.

> 6.

Interesting idea, I'll definitely look into it. I am ignorant of all the intricacies of the different GFF formats. Will do some research on this.

> 7.

I believe in that case, they should use the `util` image packaged with the repo. But I'm open to other suggestions for a default image to recommend. You are right that I can't really say "use a docker image" and then not offer any default!

> 8.

Thanks for the vote on this. I'm leaning towards agreeing with you, but don't feel particularly strongly.

> 9.

The documentation in the header would serve to document **the file**, which we group by tool (so our header documentation usually links out to the documentation for whatever tool the file is named after). The `meta` sections document a specific task. IMO there's enough differentiation there in purpose that we need both.
I think links to external documentation are appropriate for the header and should potentially be enforced, but probably inappropriate for `meta` sections (maybe there could be exceptions, but I can't think of any that would make sense).
The header wouldn't document individual tasks like the `meta` does.

a-frantz:
> I have a strong preference for using the <<< >>> syntax as that forces WDL expressions to use the syntax ~{} and that disambiguates from bash's ${} syntax, which is allowed in the {} WDL syntax.

I forgot about that piece of it (I've been *only* using the `~{}` notation for so long I forgot about this problem), but I definitely agree. In my mind, with what @adthrasher contributed it's an open and shut case in favor of always using `<<<>>>`.  @jsunny23 do you still disagree?

re: using one of the existing guides as a "base" we build on
BioWDL style guide appears as if it's no longer maintained, so seems a poor choice for us to adopt. I also disagree with a few of their decisions, and don't want to be tied to those (I can be more specific about my disagreements if anyone is curious, but I don't think the specifics are relevant). I am curious if we can find out if anyone other than BioWDL follows the guide though. That's more of a curiosity than something which would impact our decision here.
I hadn't realized that the Broad guide was still in development. I think that's more than sufficient to disqualify it from adoption. It could very well be changed or added to, and that would be a headache to keep up with as we'd be downstream. I don't really have specific disagreements with Broad, but they include very few hard and fast rules, so that's not saying much. It looks like a document that's still being written. I'd like to keep an eye on it, just to see how it develops, but I think it's obviously a poor choice to base our own style guide on.

With all that being said, I think we should continue with proposing this as it's own standalone guide, not tied to anything else out there.

When I wrote this, I was really only thinking about **this** repository and our team. But I'm now taking a step back and realizing there isn't a satisfactory guide for writing WDL, and that's a vacuum we could fill. What I have written here doesn't work as a general use guide, it contains a lot which would only apply to our repository (best example is enforcing specific parameter_meta strings. Works for us, would be odd to enforce for someone else's repository).

How do we feel about attempting to turn this into a general use WDL guide? It's something I think the WDL community could really use. If we want to move in that direction, what I've proposed here will need to be split into 2 documents. The "general use guidelines" and the "rules specific to our repo (that only contributors would have to read)". The "general use guidelines" would have to live in their own repository/site and not be tied to these workflows and tasks. How does that sound to yall?

template/task-templates.wdl
```diff
@@ -0,0 +1,228 @@
...
+
+    parameter_meta {
+        prefix: "Prefix for the <type of file> file. The extension `<extension>` will be added."
+        memory_gb: "RAM to allocate for task"
+        disk_size_gb: "Disk space to allocate for task"
+        max_retries: "Number of times to retry in case of failure"
+    }
+
+    input {
+        String prefix = basename(<input file>, ".<extension>") + ".<new extension>"
```
adthrasher, line 167: Should this have `+ ".<new extension"`? That doesn't seem like it should be in the default value of `prefix`, but rather should be added as needed in the task.

style-guide.md
```diff
@@ -0,0 +1,51 @@
...
+  - Certain Bash constructions cause problems with the bracket notation
+- output file names should *always* be determined with either the `outfile_name` parameter or the `prefix` parameter.
+  - TODO: address tasks with multiple outputs
+- output variable names should be short but descriptive
+- All tasks should run in a Docker container
+  - TODO: provide a default Docker image to use
+- no whitespace on empty lines
+- WDL lines should be less than 100 characters wide whenever possible
+  - Exceptions would be long strings that WDL doesn't allow to be broken up
+  - TODO: Do we enforce this within the `command` block? Or do we allow Bash to exceed 100 characters?
```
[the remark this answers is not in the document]
a-frantz: I wouldn't say "many", but definitely some. I think all the `parse_inputs` tasks have some ugly Bash `if` statements that get long. I also think some tasks with a lot of parameters were originally written as one line, but as I've come across those I've added line breaks. I don't know how many still exist.

template/task-templates.wdl
```diff
@@ -0,0 +1,228 @@
...
+    input {
+        Int modify_memory_gb = 0
+        Int modify_disk_size_gb = 0
+        Int max_retries = 1
+    }
+
+    Int input_size_gb = ceil(size(<input files>, "GiB"))
+
+    Int disk_size_gb = ceil(input_size_gb * 1.5) + modify_disk_size_gb
+    Int memory_gb = ceil(input_size_gb * 1.2) + modify_memory_gb
```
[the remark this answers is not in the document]
a-frantz: Good catch. I think you're right, that replacing those Floats with undefined variables is the right approach.

style-guide.md
```diff
@@ -0,0 +1,51 @@
...
+      - If a user does not supply `max_retries`, those task level defaults will get used
+      - If a user does supply `max_retries`, it should override the default for *every* task called
+- multi-core tasks should always follow the conventions laid out in the `detect_nproc_task` example (see `template/task-templates.wdl`)
+- `command` blocks should be wrapped with arrows (`<<< >>>`) instead of brackets (`{ }`)
+  - Certain Bash constructions cause problems with the bracket notation
+- output file names should *always* be determined with either the `outfile_name` parameter or the `prefix` parameter.
+  - TODO: address tasks with multiple outputs
+- output variable names should be short but descriptive
+- All tasks should run in a Docker container
+  - TODO: provide a default Docker image to use
```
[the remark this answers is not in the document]
a-frantz: Maybe we don't need a default. Given some more thought, I think each task has such specific needs that it becomes a futile effort to suggest a default. I think maybe instead of suggesting a default, I add the following bullet points:

```
- whenever possible, prefer an image maintained by an external source (such as BioContainers) to creating your own
- general purpose tasks can use the `util` image maintained in this repo
```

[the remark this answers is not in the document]
a-frantz: tangentially related: should we address how we version Docker images in this PR? Or is that out of scope?

I can't remember how we landed on the current versioning scheme, and I'm not a huge fan of it. But I'm also not sure just using the underlying tools version is appropriate, because we make changes/additions and those should be tracked in the version number as well. It's not uncommon to update our docker image without touching the underlying tool. Would we just overwrite the previous version? I don't like that.

So I'm not sure what the right approach is, I see pros and cons to the two obvious choices. I don't know which one wins my vote, or if there's another possibility I haven't thought of.

jsunny23:
@a-frantz @adthrasher 
Base WDL guides: I agree with what both of you said. If there isn't full agreement with a guide, we don't have to use it as a base, it is just an option to delegate basic style guidelines to other guides. If there are sections of other guides that seem useful to have, they could be copied over to this guide. But it might not be necessary because anyone working on this repository is generally consistent with the existing formatting.

`command` block syntax: I'm good with using the HEREDOC style. I think the guide should work for the people who need to use it, and you two have been write more WDL recently, so I defer to your opinions about it.

> This isn't written in the guide, but I intended for it to be disallowed to provide a way to users/contributors to mix for the same resource. i.e. a task cannot have both memory_gb and modify_memory_gb. 

Yeah, that was what I was asking. I think it would be good to make it explicit that those are mutually exclusive and shouldn't be used together.

> The documentation in the header would serve to document the file, which we group by tool (so our header documentation usually links out to the documentation for whatever tool the file is named after). The meta sections document a specific task.

Ah okay, I get the distinction. That sounds good to me.

style-guide.md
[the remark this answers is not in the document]
a-frantz: Potential third option: It would require breaking from SemVer so I'm a little hesitant right off the bat. (Although some of the tools we use don't follow SemVer; so adopting the "version based on underlying tool" approach also means leaving SemVer).

I'm proposing adding another tracking number to the underlying tool, which represents "our" patch number for the tool.
For example, we use STAR version `2.7.10b` (not SemVer). Our Docker image would get tagged as `star:2.7.10b.0`. Then, say we make a change like updating `sort_star_input.py`, we'd bump the Docker tag to `star:2.7.10b.1`.
When we update the tools version, "our patch" number gets reset to `0`.

I think this is a good compromise that should leave everyone happy. It makes it explicit which version of the tool is being used, and also allows us to track changes for our own purposes.

What do you think @adthrasher ?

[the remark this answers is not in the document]
adthrasher: As I was rereading this thread, your latest proposal was exactly what popped into my head. I think we would need to consider exactly how to tag the images, but incorporating the underlying tool version along with our own incrementable version seems like a good compromise. Maybe a Slack discussion with the genomics team would be warranted. I'm curious to hear what others, who have more experience versioning, think.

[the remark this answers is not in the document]
a-frantz: Sounds good to me. I'll writeup a summary and post it in the genomics channel!

style-guide.md
```diff
@@ -0,0 +1,51 @@
+# WDL Style Guide
+
+All rules below should be followed by contributers to this repo. Pull Requests which do not conform to these specifications will be asked to change.
+
+## Rules
+
+- All WDL should be written in v1.0
```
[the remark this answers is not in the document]
claymcleod: My two cents: [the history of the WDL 1.1 doc](https://github.com/openwdl/wdl/commits/main/versions/1.1/SPEC.md) suggests that the spec has been adopted for at least two years. To that end, I would not characterize it as particularly "cutting edge". This then turns into a question of "should we lag our support because Cromwell/Terra don't support the latest version?" In my mind, particularly if we are using miniwdl and miniwdl supports v1.1, I think the answer to that question should be "no".

style-guide.md
```diff
@@ -0,0 +1,56 @@
...
+- The following sections must be present and in this order: `meta`, `parameter_meta`, `input`, `command`, `output`, `runtime`
+- The `meta` section should have a `description` of the task
+  - `meta` should *not* include `author` or `email` keys
+- All inputs must have a corresponding `parameter_meta` entry
+  - These texts should be copy and pasted from other tasks with the same input when possible
+    - see `template/common-parameter-meta.txt` for common description strings.
+  - Inputs and parameter meta must be in the same order
+  - That order should be logical
```
claymcleod, line 15: I think this needs more defining. How would you implement this if you had to write a tool to check for it?

style-guide.md
```diff
@@ -0,0 +1,56 @@
...
+  - This should be defaulted to `1` for nearly all tasks
+  - Some tasks are particularly error prone and can have a higher default `max_retries`
+  - **rule specific to workflows:** `max_retries` should be an optional `Int?`
+    - This allows each task to have it's own specific default `max_retries`
+      - If a user does not supply `max_retries`, those task level defaults will get used
+      - If a user does supply `max_retries`, it should override the default for *every* task called
+- multi-core tasks should *always* follow the conventions laid out in the `detect_nproc_task` example (see `template/task-templates.wdl`)
+  - this is catering to cloud users, who may be allocated a machine with more cores than are specified by the `ncpu` parameter
+- `command` blocks should be wrapped with arrows (`<<< >>>`) instead of brackets (`{ }`)
+  - Certain Bash constructions cause problems with the bracket notation
```
claymcleod, line 45: What should we do in these cases where bash constructions cause problems?

style-guide.md
```diff
@@ -0,0 +1,56 @@
...
+- output file names should *always* be determined with either the `outfile_name` parameter or the `prefix` parameter.
+  - `outfile_name` should be preferred if no downstream tasks/tools rely on the file name/extension
+  - tasks with multiple outputs should always use the `prefix` convention
+- output variable names should be short but descriptive
+- All tasks should run in a Docker container
+  - whenever possible, prefer an image maintained by an external source (such as BioContainers) rather than creating your own image
+  - general purpose tasks can use the `util` image maintained in this repo
+- no whitespace on empty lines
```
claymcleod, line 53: No whitespace at the end of lines?

style-guide.md
```diff
@@ -0,0 +1,56 @@
+# WDL Style Guide
```
claymcleod, line 1: Perhaps we don't do it in this version, but I think some details on what a "good" documentation header block looks like would be helpful. Most of the time, people just want a template they can fill in.

claymcleod (changes requested):

style-guide.md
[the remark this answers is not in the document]
adthrasher: To be blunt, I also have no confidence in the Broad/Cromwell/Terra teams. They are already 2+ years behind while completely ignoring multiple comments, issues, etc. on both GitHub and JIRA. There was one blog post by Terra end of 2022 that indicated support in Q1 2023, but that has gone with zero update from that team.

So I agree that we should decide if we want 1.1 features and people will just have to use `miniwdl`. We also have the opinion that `miniwdl` is the better tool and a better user experience.

style-guide.md
adthrasher: I would suggest we be more concrete. For example, input files, output naming, tool arguments, then common arguments, or something similar.

style-guide.md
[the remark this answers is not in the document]
a-frantz: My vote would also be to update to v1.1 sooner rather than later. But when we do that, we also need to update `plaid` which currently uses cromwell. Just an extra complication to consider.

I think we wait till `miniwdl-lsf` enters production (any update on that @adthrasher ?). Once that's in prod, we update this repo to 1.1

style-guide.md
adthrasher: Using the arrow notation should be a requirement and that solves the issues. It is the modern functionality and other than consistency with delimiters of other blocks, it seems to be strictly better.

style-guide.md
a-frantz: I agree this is too vague. I was struggling to come up with concrete rules, but I think Andrew is on the right track. Implementing that check in a linter would be difficult.

Unless we wanted to enforce something like: `File`, `File?`, `String`, `String?`, `Int`, etc. That would be trivial to test in a linter, but I don't think it leads to the best code.

style-guide.md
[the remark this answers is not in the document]
claymcleod: I like that plan.

style-guide.md
claymcleod: It would be really nice if WDL supported semantic tagging of some of these things using keywords. That can get messy really quickly, but even just distinguishing between inputs, outputs, and non-descript parameters would be a pretty good balance.

The other option you have to consider is something to do with alphabetizing things. I don't have a clear answer on this, but would advocate we choose something practical and objective that a tool can achieve, even at the risk of less elegant code. We can always build on it from there.

style-guide.md
claymcleod: Sorry, I misread the original text here. Okay, makes sense now.

style-guide.md
a-frantz: Best compromise between "easy to implement in code" and "elegant looking" that I could come up with:
- Required inputs first
  - next optional
  - last inputs with defaults
- within above groups:
  - group by var type
  - `File` first
  - IDK what the rest of the ordering should be haha

I'm just going to play around with some tasks in the repo and see what looks best to me. But I'd like to avoid alphabetizing (annoying to insert things into the right order) and I think I'm a fan of leaving a little bit of wiggle room to developers. IMO it would be best to leave it up to the developer whether `File bam` or `File fasta` comes first (though I do believe `bam` should come first, but I don't want to get into the business of guessing what the intended uses of the variables are)

Another benefit of leaving *some* ordering up to developers that I'm just thinking of: it will allow you to match the ordering of `foo_tool -h` output. That's obviously an arbitrary ordering per tool, but IMO the order should be reflected in WDL (and there's no way for a linter to know what that ordering should be).

contributor-style-guide.md
```diff
@@ -0,0 +1,38 @@
+# WDL Style Guide
+
+All rules below should be followed by contributors to this repo. Contributors should also follow the less specific rules outlined in `style-guide.md` Pull Requests which do not conform to these specifications will be asked to change.
+
+## Rules
+
+- All WDL should be written in v1.0
+- All inputs must have a corresponding `parameter_meta` entry
```
adthrasher, line 8: Should this be only for our repo? I think this should be expected for any repo.

contributor-style-guide.md
```diff
@@ -0,0 +1,38 @@
...
+  - **rule specific to workflows:** `max_retries` should be an optional `Int?`
+    - This allows each task to have it's own specific default `max_retries`
+      - If a user does not supply `max_retries`, those task level defaults will get used
+      - If a user does supply `max_retries`, it should override the default for *every* task called
+- multi-core tasks should *always* follow the conventions laid out in the `detect_nproc_task` example (see `template/task-templates.wdl`)
+  - this is catering to cloud users, who may be allocated a machine with more cores than are specified by the `ncpu` parameter
+- output file names should *always* be determined with either the `outfile_name` parameter or the `prefix` parameter.
+  - `outfile_name` should be preferred if no downstream tasks/tools rely on the file name/extension
+  - tasks with multiple outputs should always use the `prefix` convention
+- All tasks should run in a Docker container
```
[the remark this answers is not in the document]
a-frantz: Ok, I wasn't sure if there was a valid use case where you wouldn't want one. I don't have a problem putting it in the linter if we agree it should always apply.

But again, this is still in the main style guide, same copy-paste situation as the `parameter_meta` point. I'll make it clearer.

style-guide.md
```diff
@@ -0,0 +1,65 @@
...
+- output file names should *always* be determined with either the `outfile_name` parameter or the `prefix` parameter.
+  - `outfile_name` should be preferred if no downstream tasks/tools rely on the file name/extension
+  - tasks with multiple outputs should always use the `prefix` convention
+  - TODO: IMO another "best practice" that shouldn't be enforced by a linter. Remove from guide?
+- output variable names should be short but descriptive
+  - TODO: This I'm more inclined to enforce with the linter. I think disallowing specific output names is better than forcing users into using specific ones.
+  - Disallowed output names: `out*`, any name less than 3 characters
+    - TODO: is this sufficient?
+- All tasks should run in a Docker container
+  - TODO: is there a valid reason not to use a Docker container? Should this be enforced by the linter?
```
a-frantz, line 59: me and @adthrasher discussed this offline. We've decided there could be a use case where a user is unconcerned with a "reproducible" workflow, and just wants it to work on their customized cluster environment.

So the approach we decided to take is to leave this line in the style guide, and by default enforce that every task has a docker image with the linter, but provide a CL flag for disabling that specific check. It'll be up to the user if they want to ignore this rule, but we should encourage them not to.

contributor-style-guide.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Rules
+
+- All WDL should be written in v1.0
+- All inputs must have a corresponding `parameter_meta` entry
+  - These texts should be copy and pasted from other tasks with the same input when possible
+    - see `template/common-parameter-meta.txt` for common description strings.
+  - If applicable, use the same parameter name, help string, and parameter ordering as the underlying tool called by the task
+- TODO ordering of inputs more specifically than `style-guide.md`
+  - e.g. RAM allocations before disk space allocations
```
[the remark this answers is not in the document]
a-frantz: I'd just like consistency. The actual order (past whatever definition we come up with for the general guide) seems like a largely arbitrary decision that doesn't impact anything, including readability.

For example, I'd like the last parameters to always be:
```
Int <modify_>memory_gb = X
Int <modify_>disk_size_gb = Y
Int max_retries = 1
```
I don't care what order those 3 appear in, but it should be the same for every task in our repository.

style-guide.md
```diff
@@ -0,0 +1,65 @@
...
+      - Required inputs at the top
+      - "sample" files before reference files
+      - Resource configuration at the bottom
+        - memory allocations come before disk space allocations
+      - `max_retries` should be last
+      - If applicable, `detect_nproc` immediately before `max_retries`
+- Tasks with string parameters for which a limited number of choices are valid, must be documented following the template in `string_choices_task` (see `template/task-templates.wdl`)
+  - TODO: this can't be enforced by a linter. Do we keep it in the guide?
+  - they should also fail quickly with an informative error message if an invalid input is provided
+    - In most cases, just passing the parameter to the underlying tool should produce a satisfactory error, but this must be checked for each task
```
adthrasher, line 28: This should probably be in our style guide, not necessarily the general guide, since it can't be enforced by a tool.

style-guide.md
```diff
@@ -0,0 +1,65 @@
...
+- All tasks must have configurable memory and disk space allocations
+  - TODO: only enforceable if we require specific var naming. Do we want to enforce specific variable names? Keep this in the guide?
+  - see the various tasks in the template directory for possible ways to allocate resources
+    - Contributors can mix and match the available templates, copy and pasting subsections as appropriate
+    - It is allowed to have one resource allocated dynamically, and another allocated statically in the same task.
+    - It is *not* allowed to have a resource which can be allocated *either* statically or dynamically.
+      - This is technically feasible, but is too complicated for maintenance and end-users.
+      - e.g. `memory_gb` and `modify_memory_gb` cannot be present in the same task.
```
adthrasher, line 36: We could try to write some sort of pattern like `[0-9]+[GB|MB|KB]?`. That seems fraught with difficulty. The spec has:

```
Int - Intepreted as bytes
String - This should be a decimal value with suffixes like B, KB, MB or binary suffixes KiB, MiB. For example: 6.2 GB, 5MB, 2GiB. 
```

style-guide.md
```diff
@@ -0,0 +1,65 @@
...
+  - TODO: again, only enforeceable with specific var names. Not sure if we want to get into that business. Keep it in the guide?
+  - This should be defaulted to `1` for nearly all tasks
+  - Some tasks are particularly error prone and can have a higher default `max_retries`
+  - **rule specific to workflows:** `max_retries` should be an optional `Int?`
+    - This allows each task to have it's own specific default `max_retries`
+      - If a user does not supply `max_retries`, those task level defaults will get used
+      - If a user does supply `max_retries`, it should override the default for *every* task called
+- multi-core tasks should *always* follow the conventions laid out in the `detect_nproc_task` example (see `template/task-templates.wdl`)
+  - this is catering to cloud users, who may be allocated a machine with more cores than are specified by the `ncpu` parameter
+  - TODO: this would be tough to validate with a linter. It is definitely a best practice, but IMO doesn't belong in the guide
```
adthrasher, line 47: I think this also should move to our contributor guide.

style-guide.md
```diff
@@ -0,0 +1,65 @@
...
+- `command` blocks should be wrapped with arrows (`<<< >>>`) instead of brackets (`{ }`)
+  - Certain Bash constructions cause problems with the bracket notation
+- output file names should *always* be determined with either the `outfile_name` parameter or the `prefix` parameter.
+  - `outfile_name` should be preferred if no downstream tasks/tools rely on the file name/extension
+  - tasks with multiple outputs should always use the `prefix` convention
+  - TODO: IMO another "best practice" that shouldn't be enforced by a linter. Remove from guide?
```
adthrasher, line 53: Agreed, let's move it to the contributor guide.

style-guide.md
```diff
@@ -0,0 +1,65 @@
...
+- output variable names should be short but descriptive
+  - TODO: This I'm more inclined to enforce with the linter. I think disallowing specific output names is better than forcing users into using specific ones.
+  - Disallowed output names: `out*`, any name less than 3 characters
+    - TODO: is this sufficient?
```
adthrasher, line 57: I think this is reasonable to enforce. Getting the pattern right is going to be challenging. 3+ characters is probably a good start. Can you think of any descriptive name that would start with `out`, maybe something with `outer` or `outside` though I lack a good example?

style-guide.md
```diff
@@ -0,0 +1,65 @@
...
+- All tasks should run in a Docker container
+  - TODO: is there a valid reason not to use a Docker container? Should this be enforced by the linter?
+- no whitespace on empty lines
+- no whitespace at the end of lines
+- WDL lines should be less than 100 characters wide whenever possible
+  - Exceptions would be long strings that WDL doesn't allow to be broken up
+  - This restriction applies to embedded code in the `command` block as well.
+    - TODO: I'm still not sure about this decision. I have reservations applying this rule to the command block. I've found a small handful of cases where I think the most readable formatting extends over 100 characters. They are rare, but exist frequently enough I'm not sure I like this rule.
```
adthrasher, line 65: `parameter_meta` is also likely to exceed this mark, for certain parameters. For example below, we have
```
        modify_disk_size_gb: "Add to or subtract from dynamic disk space allocation. Default disk size is determined by the size of the inputs. Specified in GB."
```

a-frantz:
Pending moving a bunch of the bullet points out of `style-guide.md` and into `contributor-style-guide.md` as @adthrasher suggested: I generally agree they'd be too hard to enforce with a linter.

But what if we renamed that file to `best-practices.md`? I believe everything in there could and should apply to every WDL repo out there. We just don't want to open the can of worms of enforcing those rules. It would give individual repos freedom to name their variables however they want and the specific implementation is left up to them. But we as a team are encouraging every WDL writer to avoid hardcoded file names, resource allocation, ncpus, etc.

IMO those are guidelines that every repo should follow. But I don't want to be in the business of writing the logic to handle all those cases.

style-guide.md
a-frantz: I think this also belongs in the contributor guide. I think it would be too difficult to enforce. Because a check that "does an input fitting pattern `foo` exist?" isn't enough (and getting that pattern will be a challenge by itself). We would also need to check that it gets used in the allocation logic, which frequently involves multiple WDL statements. Technically possible to do, but IMO not worth implementing for v1 of a linter. We could come back to this in a potential v2 if we wanted.

style-guide.md
a-frantz: Yes, `parameter_meta` will have to be excluded from this check. WDL v1 does not offer a way to insert line breaks in long strings for this section. There will probably be other exceptions as well, though I can't think of any atm. In these cases where conforming to the WDL spec involves breaking the 100 character limit, we'll just need to ignore it. Nothing we can do that doesn't break things.

_But_, the `command` block can theoretically ALWAYS be forced into under 100 characters. I just don't think that that always leads to the most readable code. It does _most_ of the time, but there are exceptions. For example, we have a few lines like the below:
```
        kraken2-build --build \
            ~{if protein then "--protein" else ""} \
            --kmer-len ~{kmer_len} \
            --minimizer-len ~{minimizer_len} \
            --minimizer-spaces ~{minimizer_spaces} \
            ~{if (max_db_size_gb > 0) then "--max-db-size" else ""} ~{if (max_db_size_gb > 0) then max_db_size_bytes else ""} \
            --threads "$n_cores" \
            --db ~{db_name}
```
IMO this is the most readable format we could write, and it's 128 characters. It could be rewritten to be under 100 like so:
```
            --minimizer-spaces ~{minimizer_spaces} \
            ~{if (max_db_size_gb > 0) then "--max-db-size" else ""} \
                ~{if (max_db_size_gb > 0) then max_db_size_bytes else ""} \
            --threads "$n_cores" \
```
But IMO that's harder to follow.

Maybe we could add a way to the linter to disable the error on a per-line basis? Ending the line with something like `# --ignore-long-line`? Problem with this is it compounds the problem. We'd be going "this line is readable when it's long so it should be an exception. To signify that we'll make it _EVEN LONGER_" 😆 

Maybe we have to accept that _rarely_, forcing a line under 100 characters will be slightly worse, but it leads to more readable code the majority of the time.

We have some tasks that were written like:
```
tool_foo --parameter1 {foo} --parameter2 {foo} --parameter3 {foo} etc.
```
that extend well past 100 characters and are just terrible to look at. I most definitely want our tool to flag those lines as needing to be rewritten.

Not sure what the answer is here. I just threw all my thoughts up against the wall here. Do you have any thoughts to contribute on this?

style-guide.md
a-frantz: `outer*` and `outside*` are good exceptions. Maybe also `outsource*`, `outward*`, `outrange*`, `outbreak*`, `outskirt*`, `outlie*`, and `outlay*`?

I just googled "words that start with out" and went through [this list](https://www.thefreedictionary.com/words-that-start-with-out) (which took longer than I'd like to admit. More words start with "out" than I realized lol). But I kept an eye out for anything that might make sense for this context, and I think we'd be fine banning anything not in the above exceptions.

style-guide.md
a-frantz: Ok, I spent some time weighing the options here, and I've decided that I'm ok with enforcing character limits within the `command` block. Yes, there will be a handful of rare cases where this makes things less readable, but the collapsed alternatives aren't worse enough to justify dealing with making exceptions.

However I was reading through the the [`black` rules](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#), mainly looking for whitespace rules to steal (and those have been committed), but they have a section justifying why their character limit is 88 wide. I chose 100 pretty arbitrarily. `black` allows the line length to be configured, not sure if we want to deal with that or just pick something sane and enforce it. But here's what they have to say about lengths longer than the 88 default:

> You can also increase it, but remember that people with sight disabilities find it harder to work with line lengths exceeding 100 characters. It also adversely affects side-by-side diff review on typical screen resolutions. Long lines also make it harder to present code neatly in documentation or talk slides.

Now I'm considering dropping our line length from 100 to 90. "88" is just a weird number to choose if you ask me, so my opinion is to take their suggestion and round to a nicer number. That's 90.

@adthrasher any opinion on either point I make here?

adthrasher: Thanks for looking that up. The two points made by `black` are good. I'm fine with picking a round number. Their documentation also says "90-ish" is a good choice, so using the round number seems fine.

style-guide.md
a-frantz: Note that I'm skirting the issue of enforcing a case system. WDL itself is inconsistent (e.g. `maxRetries` and `parameter_meta` both exist in the spec). I think:
1) it would be difficult to enforce sanely
2) not really our job
3) the language itself is inconsistent, so what's the point

I have a strong preference for snake case, but I'm not ready to enforce snake case on other people's code (now that I'm writing this comment though, I will add it to our "best practices" guide).

The regexs in the previous comment also allows a writer to get past the filter by ALL CAPSing their variable names (or at least the `IN`/`OUT` prefixes). Which... I don't really like, but I'm hesitant to cast a wider net. If someone is that insistent on writing ugly code, I don't think any tool we implement will stop them. They probably aren't our intended user 🤷

template/common-parameter-meta.txt
a-frantz: Closing due to being out of scope for this PR

a-frantz (commented):
I'm happy with the state of this PR, minus a few things I'll ask below. But as it stands, I think we can merge once the last few ends are tied up.

Not sure where it will fall on my priority list, but I think we should take on this proposed linter/auto-formatter. When we do, `style-guide.md` will get moved to the linter's repo and we'll add a link in our README saying "formatting is performed by tool <foo>".
I don't know where `best-practices.md` should live in the long term. I think it's mostly good advice that every WDL repo should follow, but there are some specifics in there that really just apply to us (Docker versioning conventions, "use WDL v1.0"). I think sometime in the future (after the linter is published) we should consider publishing some sort of blog post with these rules. It would be good for the WDL community to follow the guidelines we've developed here.

**Outstanding questions**
We've avoided talking about WDLdoc headers in this PR. Is that the right move? Should we provide a template here? What does a "good" header look like? We've been inconsistent about putting licensing information in the header. I think the repo as a whole should be licensed and then we drop all licensing from the files themselves. It's just chuff that gets skipped over.

Continuing that thought, I've envisioned that this proposed linter will be highly opinionated, and as an extension of that provide some bare bones templating. e.g. it will write in empty `meta` and `parameter_meta` blocks if they're missing. Could this tool also write a bare bones WDLdoc header template if none is found?

adthrasher:
> Pending moving a bunch of the bullet points out of `style-guide.md` and into `contributor-style-guide.md` as @adthrasher suggested: I generally agree they'd be too hard to enforce with a linter.
> 
> But what if we renamed that file to `best-practices.md`? I believe everything in there could and should apply to every WDL repo out there. We just don't want to open the can of worms of enforcing those rules. It would give individual repos freedom to name their variables however they want and the specific implementation is left up to them. But we as a team are encouraging every WDL writer to avoid hardcoded file names, resource allocation, ncpus, etc.
> 
> IMO those are guidelines that every repo should follow. But I don't want to be in the business of writing the logic to handle all those cases.

I think that rename makes a lot of sense.

best-practices.md
```diff
@@ -0,0 +1,46 @@
...
+- Whenever possible, prefer a Docker image maintained by an external source (such as BioContainers) rather than creating your own image
+- When adding a Dockerfile to this repository, follow the below conventions
+  - The `Dockerfile` should be nested under the `docker/` directory, a folder with a name for the image (in most cases the name of the primary tool), and finally a folder named after the version being built.
+  - Docker images should be versioned according to the following convention
+    - Start with the version of whatever tool is named in the path to the `Dockerfile`
+      - If no specific tool is named (e.g. the `util` image), default to SemVer. Ignore the next 3 bullet points.
+    - Followed by a dash-zero (`-0`)
+      - If the Docker image gets updated, *without* updating the base tool's version, increment the number after the dash (`-`) by one
+      - If the Docker image gets updated, *including* updating the base tool's version, revert back to a dash-zero (`-0`)
+- general purpose tasks can use the `util` image maintained in this repo
```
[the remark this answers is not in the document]
adthrasher: I think that is clear. We could be explicit that the `-N` is to version changes to the image itself (not including tool version changes). I'm not sure how to write that succinctly, so I'm fine to leave it as well.

adthrasher (approved):

claymcleod:
> We've avoided talking about WDLdoc headers in this PR. Is that the right move? Should we provide a template here? What does a "good" header look like? We've been inconsistent about putting licensing information in the header. I think the repo as a whole should be licensed and then we drop all licensing from the files themselves. It's just chuff that gets skipped over.

I think we should follow the pattern that other successful languages have followed, which is (in short) to give you the tools to write whatever documentation makes sense for one's own use case while also providing a strong convention that works well. Individuals can look at our convention for ideas, but they are free to break from it if it suits their purpose. To that end, I think the header should be in markdown, and how we specifically write headers in our repos would fall into a "St. Jude Cloud Best Practices" guide (or similar).

> Continuing that thought, I've envisioned that this proposed linter will be highly opinionated, and as an extension of that provide some bare bones templating. e.g. it will write in empty meta and parameter_meta blocks if they're missing. Could this tool also write a bare bones WDLdoc header template if none is found?

Again, with the linter, I think following the successes in the community is the right thing to do. For example, the Rust linter is highly opinionated—it goes so far as to enforce snake case variables for functions and camel case names for enums. However, you have the ability to disable these lints if you'd like. I think that would be the right way to go.

You're also getting into the territory of like a language server for autocomplete and things like that. I would suggest we tackle the linter first and then consider whether the language server is worth the effort.

style-guide.md
```diff
@@ -0,0 +1,49 @@
...
+    - `struct`
+    - `Object`
+    - `Map[*, *]`
+    - `Pair[*, *]`
+    - `String`
+    - `Boolean`
+    - `Float`
+    - `Int`
+  - for ordering of the same compound type, drop the outermost type and recursively apply above sorting
+  - once the above ordering is satisfied, it is up to the developer for final order of inputs of the same type.
```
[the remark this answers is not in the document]
claymcleod: Yes I'm fine with this ordering.

claymcleod (approved):

PR #90, opened 2023-05-11
feat: put strandedness value into HTSeq file name

a-frantz:
There is a problem with relying on the derived values for HTSeq. Lazy users won't check the `ngsderive` output results and might not realize that `HTSeq` was run with an "unexpected" strandedness mode. This PR is the best way me and @adthrasher could come up with to cater to those lazy users. Of course, maybe we don't need to and that's solely a "them problem" we don't need to solve. It's taking me more time to write this PR summary than it did to make the code changes, so not a problem if we opt to close without merging.

WDL lacks any real way to communicate with users. So the best we could come up with is this heavy-handed approach of modifying the filename to reflect the strandedness mode. Neither me or Andrew are particularly thrilled with it.

But it addresses a potentially major problem for users. If they use the feature counts files in an analysis assuming it was generated by strandedness value X instead of Y, their whole analysis becomes junk. I suppose the question is "do we need to solve that problem for them?" At the end of the day it would be user error because they didn't look at the complete set of result files. There's a good argument we don't need to make any changes to cater to lazy users. But it was trivial to "fix", so shouldn't we go ahead and fix it for them?

Just putting this PR out there to generate discussion.


adthrasher (approved):

claymcleod:
I'm fine with this.

claymcleod (approved):

PR #92, opened 2023-05-13
feat: use samtools collate and fastq instead of picard

a-frantz:
Title explains it. This increases storage costs (by a significant amount, though I don't have hard numbers to share), however decreases runtime by a large margin (again, no hard numbers. Just my eyeballing), which will result in lower costs and faster results.

Note: this does change the behavior of the `bam_to_fastq` step. `picard SamToFastq` will error and fail if it finds an "illegal mate state". `samtools fastq` simply discards the read(s). This is an improvement in behavior, it shouldn't be the `bam_to_fastq` step's job to catch errors like this. That should be done by `picard ValidateSam` (which does log those "illegal mate states" and in more detail than `SamToFastq`). But it's an important difference that should be noted.


template/common-parameter-meta.txt
```diff
@@ -4,7 +4,7 @@ PERMITTED META DESCRIPTIONS:
         modify_memory_gb: "Add to or subtract from dynamic memory allocation. Default memory is determined by the size of the inputs. Specified in GB."
         modify_disk_size_gb: "Add to or subtract from dynamic disk space allocation. Default disk size is determined by the size of the inputs. Specified in GB."
         ncpu: "Number of cores to allocate for task"
-        detect_nproc: "Use all available cores? Recommended for cloud environments. Not recommended for cluster environments."
+        use_all_cores: "Use all cores? Recommended for cloud environments. Not recommended for cluster environments."
```
adthrasher, line 7:
```suggestion
        use_all_cores: "Detect and use all cores? Recommended for cloud environments. Not recommended for cluster environments."
```
I think it's worth clarifying that this detects the number and then uses that.

tools/samtools.wdl
```diff
@@ -327,3 +326,152 @@ task addreplacerg {
         read_group_id: "Existing read group ID in BAM to add to reads"
     }
 }
+
+task collate {
+    meta {
+        description: "This WDL tool runs `samtools collate` on the input BAM file. Shuffles and groups reads together by their names."
+        outputs: {
+            collated_bam: "A collated BAM (reads sharing a name next to each other, no other guarentee of sort order)"
```
adthrasher, line 334:
```suggestion
            collated_bam: "A collated BAM (reads sharing a name next to each other, no other guarantee of sort order)"
```

tools/samtools.wdl
```diff
@@ -327,3 +326,152 @@ task addreplacerg {
...
+    runtime {
+        cpu: ncpu
+        memory: memory_gb + " GB"
+        disk: disk_size_gb + " GB"
+        docker: 'quay.io/biocontainers/samtools:1.17--h00cdaf9_0'
+        maxRetries: max_retries
+    }
+}
+
+task bam_to_fastq {
```
adthrasher, line 388: I think we're using `picard SamToFastq` elsewhere. Should we be consistent?

tools/samtools.wdl
```diff
@@ -327,3 +326,152 @@ task addreplacerg {
...
+    meta {
+        description: "This WDL tool runs `samtools fastq` on the input BAM file. Splits the BAM into FastQ files."
+        outputs: {
+            read1: "A gzipped FastQ containing read ones"
+            read2: "A gzipped FastQ containing read twos"
```
adthrasher, line 393:
```suggestion
        read_one_fastq_gz: "Gzipped FastQ file with 1st reads in pair"
        read_two_fastq_gz: "Gzipped FastQ file with 2nd reads in pair"
```
This is what we have in the template.

tools/samtools.wdl
```diff
@@ -327,3 +326,152 @@ task addreplacerg {
...
+            singleton_reads: "A gzipped FastQ containing singleton reads"
```
adthrasher, line 394:
```suggestion
            singleton_reads_fastq_gz: "A gzipped FastQ containing singleton reads"
```

tools/samtools.wdl
```diff
@@ -327,3 +326,152 @@ task addreplacerg {
...
+            interleaved_reads: "An interleaved gzipped paired-end FastQ"
```
adthrasher, line 395:
```suggestion
            interleaved_reads_fastq_gz: "An interleaved gzipped paired-end FastQ"
```

tools/samtools.wdl
```diff
@@ -327,3 +326,152 @@ task addreplacerg {
...
+            } \
+            -0 /dev/null \
+            ~{bam}
+    >>>
+
+    output {
+        File? read1 = "~{prefix}_R1.fastq.gz"
+        File? read2 = "~{prefix}_R2.fastq.gz"
+        File? singleton_reads = "~{prefix}.singleton.fastq.gz"
+        File? interleaved_reads = "~{prefix}.fastq.gz"
```
adthrasher, line 467:
```suggestion
        File? read_one_fastq_gz = "~{prefix}_R1.fastq.gz"
        File? read_two_fastq_gz = "~{prefix}_R2.fastq.gz"
        File? singleton_reads_fastq_gz = "~{prefix}.singleton.fastq.gz"
        File? interleaved_reads_fastq_gz = "~{prefix}.fastq.gz"
```

workflows/qc/quality-check-standard.wdl
```diff
@@ -119,12 +120,29 @@ workflow quality_check {
     call ngsderive.read_length as ngsderive_read_length { input: bam=quickcheck.checked_bam, bam_index=bam_index, max_retries=max_retries }
     call ngsderive.encoding as ngsderive_encoding { input: ngs_files=[quickcheck.checked_bam], prefix=prefix, max_retries=max_retries }
 
-    call picard.bam_to_fastq { input: bam=quickcheck.checked_bam, max_retries=max_retries }
```
adthrasher, line 122: Why the switch from Picard to samtools?

adthrasher (changes requested):

template/common-parameter-meta.txt
adthrasher: Realizing now that we have this in nearly (?) every task. I'm conflicted about making that change. I wonder if we should hold off and make sure we're happy with the strings in `common-parameter-meta` and then make the update in the Code cleanup PR.

workflows/qc/quality-check-standard.wdl
a-frantz: This switch was inspired by problems with `SJBALL033086_G1.WholeGenome` in the alpha test cohort. Unfortunately the QC results didn't elucidate much details as to why, but `Picard SamToFastq` took over a week to convert this BAM. That same BAM was processed by `samtools collate`->`samtools fastq` in under 24 hours.

I haven't investigated the "average case", but I'm pretty sure it's saves time and memory in all cases. That's really an unchecked assumption on my part, as in the name of freeing storage I `rm`d all my test logs from before this switch. That was shortsighted of me.

But regardless of previous performance on "average cases", we have an example of runtime dropping from more than a week to under 36 hours with this switch. That's more than enough to justify it if you ask me.

As for using this approach everywhere? I'd definitely consider it. There are 2 downsides:
1) it complicates things by replacing one call with 2
2) it is a pretty significant bump in storage. Collated BAMs are really beefy and need to be stored.

I just got an idea that could solve both those problems though. One task which calls both tools in the same command block. By default `rm`s the collated BAM after FastQ conversion. Can be kept with a Boolean (which we'd need for QC to pass to Qualimap). It's only a partial solution to storage though, as the very large BAM still has to be created. So the storage capacity has to exist, it's just not used for long.

a-frantz:
@adthrasher I think this is good to go, except I want to validate our assumption that piping uncompressed BAM will lead to a speed up for WGS and WES. Other than that I think all the changes I want are here.

But I've realized this is a pretty significant behavior change (for the better) for `rnaseq-standard`. In v2/v3, if an input BAM had "illegal mate states" (as defined by Picard?) it couldn't be processed by the pipeline (without manually addressing those weird read1/read2 bits).
New behavior with this PR is to simply discard any reads with wonky read1/2 bit settings. I think this is much better, but it does need to be noted _somewhere_ that reads will potentially be silently thrown out.

It's making me think that we should really be keeping CHANGELOGs for our pipelines to track things like this. I'm not looking forward to writing all the changes I've introduced between v1 and v2 of QC, but it's something we should be doing. Do you agree?

tools/picard.wdl
```diff
@@ -177,7 +177,7 @@ task bam_to_fastq {
     meta {
         author: "Andrew Thrasher, Andrew Frantz"
         email: "andrew.thrasher@stjude.org, andrew.frantz@stjude.org"
-        description: "This WDL task converts the input BAM file into paired FastQ format files."
+        description: "This WDL task converts the input BAM file into FastQ format files. This task has been deprecated in favor of `samtools.collate_to_fastq` which is more performant and doesn't error on 'illegal mate states'."
```
adthrasher, line 180:
```suggestion
        description: "[Deprecated] This WDL task converts the input BAM file into FastQ format files. This task has been deprecated in favor of `samtools.collate_to_fastq` which is more performant and doesn't error on 'illegal mate states'."
```

tools/samtools.wdl
```diff
@@ -327,3 +326,263 @@ task addreplacerg {
...
+        n_cores=~{ncpu}
+        if [ "~{use_all_cores}" = "true" ]; then
+            n_cores=$(nproc)
+        fi
+
+        samtools collate \
+            --threads "$n_cores" \
+            ~{if fast_mode then "-f" else ""} \
+            # skip compression (and decompression) if not storing the output
+            ~{if store_collated_bam then "" else "-u"} \
```
adthrasher, line 549: I thought we weren't using the uncompressed BAM output. However looking at this, since we're not storing it, if the subsequent `samtools fastq` call accepts it without error, I'm fine using it here.

workflows/general/bam-to-fastqs.wdl
```diff
@@ -36,50 +36,47 @@ version 1.0
...
-                read1=reads,
-                max_retries=max_retries
-            }
+
+    scatter (reads in 
+        zip(bam_to_fastq.read_one_fastq_gz, bam_to_fastq.read_two_fastq_gz)
+    ) {
+        call fq.fqlint { input:
+            read1=select_first([reads.left, "undefined"]),
+            read2=reads.right,
```
adthrasher, line 73: Is this correct? I would expect read one to always exist and read2 to be the optional input.

tools/samtools.wdl
a-frantz: I'm about to submit jobs verifying this works (and seeing if it has any effect on runtime). If it's neutral on runtime I plan to remove it for simplicity's sake

workflows/general/bam-to-fastqs.wdl
a-frantz: Yes this is an unfortunate side effect of enabling the full spectrum of options in `samtools fastq`. The `_R1` file won't be generated if `interleaved==true` (because then no `_R*` extension is used). However we know from the context of this workflow that the `_R1` file will always exist. Just WDL doesn't know that. It is impossible (or should be impossible) to trip the `"undefined"` part.

The `read2` parameter accepts `File?` input so no special handling of it needs to be made.

workflows/general/bam-to-fastqs.wdl
```diff
@@ -36,50 +36,47 @@ version 1.0
...
+            max_retries=max_retries
         }
     }
 
     output {
-        Array[File] read1s = bam_to_fastq.read1
-        Array[File?] read2s = bam_to_fastq.read2
+        Array[File] read1s = select_all(bam_to_fastq.read_one_fastq_gz)
+        Array[File?] read2s = bam_to_fastq.read_two_fastq_gz
```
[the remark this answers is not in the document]
a-frantz: Both `read_one_fastq_gz` and `read_two_fastq_gz` are `Array[File?]` type. So `read_one_fastq_gz` has to be coerced into a required type, but because of the declaration of `read2s` (`Array[File?]`) nothing needs to be done to `read_two_fastq_gz`

workflows/qc/quality-check-standard.wdl
```diff
@@ -257,6 +270,7 @@ workflow quality_check {
         File? qualimap_rnaseq_results = qualimap_rnaseq.results
         File? junction_summary = junction_annotation.junction_summary
         File? junctions = junction_annotation.junctions
+        Array[File]? intermediate_files = intermediate_files_output
```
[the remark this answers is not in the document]
a-frantz: Good idea. I think this would be a sensible use of `Struct`.

workflows/qc/quality-check-standard.wdl
```diff
@@ -133,19 +137,24 @@ workflow quality_check {
         max_retries=max_retries
     }
 
-    call picard.bam_to_fastq { input:
+    call samtools.collate_to_fastq { input:
         bam=quickcheck.checked_bam,
+        store_collated_bam=molecule=="RNA",
```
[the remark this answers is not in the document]
a-frantz: Just pushed this:
```
        bam=quickcheck.checked_bam,
        # RNA needs a collated BAM for Qualimap
        # DNA can skip the associated storage costs
        store_collated_bam=molecule=="RNA",
```
What do you think?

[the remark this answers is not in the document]
adthrasher: That works for me. It is more verbose but is more explicit to the reader.

workflows/qc/quality-check-standard.wdl
```diff
@@ -69,6 +71,8 @@ workflow quality_check {
         coverage_labels: "An array of equal length to `coverage_beds` which determines the prefix label applied to the output files. If omitted, defaults of `regions1`, `regions2`, etc. will be used."
         gtf: "Gzipped GTF features file. **Required** for RNA-Seq data"
         star_log: "Log file generated by the RNA-Seq aligner STAR. Will be added to MultiQC report."
+        output_intermediate_files: "Output intermediate files? FastQs and if RNA a collated BAM. *WARNING* these files can be large and your backend may copy these files instead of linking to the files that already exist."
```
a-frantz, line 74: @adthrasher I believe this is true. Pretty sure Cromwell used to copy _everything_. Besides Cromwell, I wouldn't be surprised if other backends do a hardcopy, but I don't know of any for sure. So I could be making up this warning. Do we leave it in?

adthrasher:
```suggestion
        output_intermediate_files: "Output intermediate files? FastQs and if RNA a collated BAM. *WARNING* these files can be large."
```

I would probably just leave it like this. It's still true and if they use a backend that duplicates the files, they still have a problem.

adthrasher (approved):

PR #93, opened 2023-05-20
feat: add subsampling option to QC and rnaseq-fastq

a-frantz:
Title. This also adds a new task for subsampling FastQ files (using the tool `fq`).


tools/fq.wdl
```diff
@@ -53,3 +53,68 @@ task fqlint {
...
+            subsampled_read2: "Subsampled read2 FastQ"
+        }
+    }
+
+    parameter_meta {
+        read1: "Input FastQ with read one"
+        read2: "Input FastQ with read two"
+        prefix: "Prefix for the output FastQ file(s). The extension `_R1.subsampled.fastq.gz` and `_R2.subsampled.fastq.gz` will be added."
+        probability: "The probability a record is kept, as a percentage (0.0, 1.0). Cannot be used with `record-count`. Any `probability<=0.0` or `probability>=1.0` to disable."
+        record_count: "The exact number of records to keep. Cannot be used with `probability`. Any `record_count<=0` to disable."
```
adthrasher, line 71: Is this enforced in our task?

tools/fq.wdl
```diff
@@ -53,3 +53,68 @@ task fqlint {
...
+        memory_gb: "RAM to allocate for task, specified in GB"
+        modify_disk_size_gb: "Add to or subtract from dynamic disk space allocation. Default disk size is determined by the size of the inputs. Specified in GB."
+        max_retries: "Number of times to retry in case of failure"
+    }
+
+    input {
+        File read1
+        File? read2
+        String prefix = basename(read1, "_R1.fastq.gz")
+        Float probability = 1.0
```
adthrasher, line 81: So the default is that this task does not perform a subsample. That seems an odd default. It would be hard to determine what a "good" default is, since that likely varies by the application. However, it seems like it should probably be less than 1.0.

adthrasher (approved):

tools/fq.wdl
a-frantz: No, I'm relying on `fq` to produce an informative error if both are supplied:
```
$ fq subsample -p 0.5 -n 100
error: the argument '--probability <PROBABILITY>' cannot be used with '--record-count <RECORD_COUNT>'
```

tools/fq.wdl
adthrasher: I'm not sure. I would be OK to pick a "random" value (such as `0.2`) as a default. My expectation would be if I ran the task without providing values that it would do a subsampling at some rate.

workflows/rnaseq/rnaseq-standard-fastq.wdl
```diff
@@ -57,6 +58,7 @@ workflow rnaseq_standard_fastq {
         gtf: "Gzipped GTF feature file"
         stardb: "Database of reference files for the STAR aligner. The name of the root directory which was archived must match the archive's filename without the `.tar.gz` extension. Can be generated by `star-db-build.wdl`"
         mark_duplicates: "Add SAM flag to computationally determined duplicate reads?"
+        subsample_n_reads: "Only process a random sampling of `n` reads. Any `n`<=`0` for processing entire input."
```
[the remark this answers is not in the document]
adthrasher: OK, I removed the extra sentence. I don't think we need to state that it is exact.

tools/fq.wdl
a-frantz: I don't think I agree with you. When would a user ever say "subsample my input. I don't care by how much"? I suppose a user might say that if they're just testing the option to see what it does. But IMO in that case, leave it up to them to pick a random value.

In all use cases I can think of, the user would have a specific amount (either read count or probability) they want to be sampled. In which case it makes sense for the options to act as required (the approach I've taken by setting invalid defaults is the closest I can get to "XOR required" in WDL).

adthrasher: OK, that second paragraph helped. When you do `1.0` and `-1`, you're not passing either option to `fq` so it will error. In that case, I agree that that is a reasonable default.

PR #94, opened 2023-05-29
feat: Style Guide and Best Practices confirmity

a-frantz:


tools/bwa.wdl
```diff
@@ -66,69 +66,69 @@ task bwa_aln {
...
     }
 
     input {
-        File fastq1
-        File fastq2
-        String output_bam = basename(fastq1, ".fq.gz") + ".bam"
+        File read_one_fastq_gz
+        File read_two_fastq_gz
         File bwadb_tar_gz
+        String prefix = basename(read_one_fastq_gz, ".fq.gz")  # TODO is this right?
```
a-frantz, line 91: @adthrasher everywhere else we use a `.fastq` extension. Do you have any idea whether `.fastq` or `.fq` is more common? Should we try and handle both? It's ugly but the below snippet should work:
```
String prefix = basename(basename(read_one_fastq_gz, ".fq.gz"), ".fastq.gz")
```

tools/cellranger.wdl
```diff
@@ -20,35 +18,43 @@ task count {
     }
 
     input {
-        String id
-        File transcriptome_tar_gz
         File fastqs_tar_gz
-        Int ncpu = 1
+        File transcriptome_tar_gz
+        String id
+        String jobmode = "local"  # TODO would a non-local jobmode work?
```
a-frantz, line 24: I am fairly certain this parameter _must_ be set to `local`. In which case we should hard code instead of exposing it like this. Or am I wrong and other viable `jobmode` options exist?

tools/cellranger.wdl
```diff
@@ -81,24 +85,21 @@ task count {
         File raw_matrix = glob("*/outs/raw_feature_bc_matrix/matrix.mtx.gz")[0]
         File mol_info_h5 = glob("*/outs/molecule_info.h5")[0]
         File web_summary = glob("*/outs/web_summary.html")[0]
-        File loupe = glob("*/outs/cloupe.cloupe" )[0]
+        File loupe = glob("*/outs/cloupe.cloupe" )[0]  # TODO should this be "cloupe" instead of "loupe"?
```
a-frantz, line 88: I have no idea what this file is. But looks to me like there's a missing `c` in the WDL

tools/picard.wdl
```diff
@@ -53,16 +54,17 @@ task mark_duplicates {
...
+        disk: disk_size_gb + " GB"
         docker: 'quay.io/biocontainers/picard:2.27.5--hdfd78af_0'
         maxRetries: max_retries
     }
 }
 
 task validate_bam {
+    # TODO should this be refactored to behave as "default" Picard behaves?
+    #   Default Picard has some weird/not ideal behaviors
+    #   e.g. `max_errors = 100`
```
a-frantz, line 66: What do you think @adthrasher ? Especially in light of potentially moving to a "community model".

Like the comment says, I don't like Picard's default behavior, and I think we have _some_ room to "improve" the default behavior. But the current state of this task is *very* far from default. And probably too much of a deviation for a "community resource"

tools/samtools.wdl
```diff
@@ -108,90 +117,92 @@ task flagstat {
...
+    parameter_meta {
+        bam: "Input BAM format file to index"
+    }
+
     input {
         File bam
+        String outfile_name = basename(bam) + ".bai"  # TODO we might want to disable customizing the name for the BAI? Breaks our best practices, but also please do not change the BAI name
+        Boolean use_all_cores = false
+        Int memory_gb = 15  # TODO this seems like more RAM than should be necessary
+        Int modify_disk_size_gb = 0
```
a-frantz, line 159: Do you have thoughts on either comment here?

I am in favor of letting the user customize these runs. But I think I'm more in favor of "please don't rename your bam index". I'd vote to remove the `outfile_name`/`prefix` parameter.

workflows/chipseq/chipseq-standard.wdl
```diff
@@ -29,6 +29,7 @@
 
 version 1.0
 
+# TODO should we enfore an order to imports? Alphabetical? Order referenced? URLs before/after relatives?
```
a-frantz, line 32: I think we should enforce an order. But I'm pretty ambivalent towards what it is. Probably alphabetical is the right approach? Do we take into special consideration paths/URLs?

workflows/general/alignment-post.wdl
```diff
@@ -10,27 +14,27 @@ workflow alignment_post {
...
+        max_retries: "Number of times to retry failed steps. Overrides task level defaults."
         xenocp_aligner: {
             description: "Aligner to use to map reads to the host genome for detecting contamination"
             choices: [
-                'bwa aln',
-                'bwa mem',
-                'star'
+                'bwa aln',  # TODO this is SE. Is `bwa_aln_pe` a possibility?
+                'bwa mem',  # TODO this is also SE.
+                'star'  # TODO I think only this will work with paired end? What would DNA use?
```
a-frantz, line 23: Can you should some light here @adthrasher ? Am I correct that these bwa tasks are for SE data?

tools/samtools.wdl
```diff
@@ -108,90 +117,92 @@ task flagstat {
...
-    meta {
-        author: "Andrew Thrasher, Andrew Frantz"
-        email: "andrew.thrasher@stjude.org, andrew.frantz@stjude.org"
-        description: "This WDL task runs Samtools flagstat on the input BAM file. Produces statistics about the alignments based on the bit flags set in the BAM."
-    }
-
-    parameter_meta {
-        bam: "Input BAM format file to index"
+    runtime {
+        dx_instance_type: "azure:mem2_ssd1_x4"
```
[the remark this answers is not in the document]
adthrasher: It's still valid for dxCompiler (https://github.com/dnanexus/dxCompiler/blob/develop/doc/ExpertOptions.md#instance-type). I would be in favor of removing it though. I don't recall how that instance type was decided or what project it was part of. 

It doesn't look like dxCompiler supports the newer syntax for runner-specific options in the `runtime` section. Hopefully they'll add support in the future.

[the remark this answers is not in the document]
a-frantz: Sounds good and looks promising. I'll delete this line.

a-frantz:
re:
> * (potential fix? Effect not investigated) when creating a collated BAM, write secondary and supplementary alignments

From reading the Qualimap docs, there are metrics computed with secondary and supplementary alignments. However I don't believe any of those computations end up in the MultiQC report.

It's a bit of a slow down, but if a user ever wants to dig into the Qualimap report instead of using the MultiQC reports, they would expect information regarding sec. and supp. alignments. So I'm going to keep this change. RNA-Seq performance is still pretty fast, so I'm willing to accept the hit on run-time.

tools/star.wdl
```diff
@@ -18,49 +16,54 @@ task build_star_db {
...
-    String stardb_out_name = stardb_dir_name + ".tar.gz"
+    String stardb_out_name = stardb_name + ".tar.gz"
     Float reference_fasta_size = size(reference_fasta, "GiB")
     Float gtf_size = size(gtf, "GiB")
-    Int disk_size = select_first([disk_size_gb, ceil(((reference_fasta_size + gtf_size) * 3) + 10)])
+    Int disk_size_gb = ceil(
+        ((reference_fasta_size + gtf_size) * 3) + 10
+    ) + modify_disk_size_gb
+
+    String memory_limit_bytes = (memory_gb - 2) + "000000000"
```
adthrasher, line 37:
```suggestion
    # Leave 2GB as system overhead
    String memory_limit_bytes = (memory_gb - 2) + "000000000"
```
I think a comment is useful here. Though I'm not sure this is the exact language to use.

tools/star.wdl
```diff
@@ -18,49 +16,54 @@ task build_star_db {
...
         n_cores=~{ncpu}
-        if [ "~{use_all_cores}" = "true" ]; then
+        if ~{use_all_cores}; then
             n_cores=$(nproc)
         fi
 
         orig_gtf=~{gtf}
         gtf_name=$(basename "${orig_gtf%.gz}")
-        gunzip -c ~{gtf} > "$gtf_name" || cp ~{gtf} "$gtf_name"
+        gunzip -c ~{gtf} > "$gtf_name" || ln -s ~{gtf} "$gtf_name"
```
adthrasher, line 49: Isn't `$(basename "${orig_gtf%.gz}")` equivalent to doing `basename $orig_gtf ".gz"`? 
Also is the `ln -s` (or `cp`) necessary? If the input wasn't gzipped, then `orig_gtf` should be the same as `gtf` right?

adthrasher: I suppose this might be localizing the file for STAR. We should add some comments to this command though.

tools/util.wdl
```diff
@@ -278,74 +280,81 @@ task qc_summary {
...
+        memory: memory_gb + " GB"
+        disk: disk_size_gb + " GB"
         docker: 'quay.io/biocontainers/samtools:1.16.1--h6899075_1'
         maxRetries: max_retries
     }
 }
 
 task unpack_tarball {
+    # TODO I don't think I've ever checked if this task actually works.
+    #   I wrote it then realized there was a better way before this made it to testing
```
adthrasher, line 343: This would also destroy any structure. If we keep the task, it should note that.

tools/util.wdl
```diff
@@ -354,27 +363,28 @@ task unpack_tarball {
...
-        disk: disk_size + " GB"
+        memory: memory_gb + " GB"
+        disk: disk_size_gb + " GB"
         docker: 'ghcr.io/stjudecloud/util:1.2.0'
         maxRetries: max_retries
     }
 }
 
 task make_coverage_regions_beds {
+    # TODO should this be customizable?
```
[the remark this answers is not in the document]
adthrasher: Should this also be specified to work with gencode-format GTFs?

workflows/chipseq/chipseq-standard.wdl
adthrasher: I think alphabetical is fine. Internal before external? Add a separator between them? I would say yes to both.

workflows/general/alignment-post.wdl
adthrasher: XenoCP converts reads to fastq, each mate in a pair gets a fastq record, and it then processes the reads as single-end. So each of the aligners here is run in single-end mode. All three can be run in paired-end mode for other applications, but XenoCP only uses the single-end mode.

In `workflows`, we have `bwa_aln` (single-end), `bwa_aln_pe` (paired-end), `bwa_mem` (single-end), and `star.alignment` (single-end or paired-end). I don't believe `bwa_aln_pe` is being used by any code at the moment. It was going to be part of the ChIP-Seq workflow until it was decided to use SEAseq.

tools/bwa.wdl
```diff
@@ -20,44 +17,47 @@ task bwa_aln {
...
     }
 
+    String output_bam = prefix + ".bam"
+
     Float input_fastq_size = size(fastq, "GiB")
     Float reference_size = size(bwadb_tar_gz, "GiB")
-    Int disk_size = select_first([disk_size_gb, ceil((input_fastq_size * 2) + (reference_size * 2))])
+    Int disk_size_gb = ceil(
+        (input_fastq_size + reference_size) * 2
+    ) + modify_disk_size_gb
```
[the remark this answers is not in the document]
adthrasher: This was discussed offline and a decision was made.

tools/star.wdl
a-frantz: > Isn't `$(basename "${orig_gtf%.gz}")` equivalent to doing `basename $orig_gtf ".gz"`?

yes that should be changed to just use basename

> Also is the `ln -s` (or `cp`) necessary? If the input wasn't gzipped, then `orig_gtf` should be the same as `gtf` right?

I think the main purpose of the `ln -s` (and what was a `cp`) is so that the same variable (`$gtf_name`) can be passed to STAR. As opposed to having along the lines of `if (gzipped) then $var1  else $var2`.

I would hope the files don't need to be localized for STAR, though I'm not sure. Regardless, this section can definitely get cleaned up

workflows/chipseq/chipseq-standard.wdl
a-frantz: > Internal before external? Add a separator between them?

No opposition to either of these. Let's adopt those rules.

But I'm still unsure how to handle paths (in either relative or URL imports). e.g. does `./rnaseq-core.wdl` come before or after `../../tools/util.wdl`? Are we sorting on `basename`? How do we handle the case of `../foo.wdl` and `../../bar/foo.wdl`?

I think the easiest solution is just do a naive sort on the strings. No special handling for paths/basenames. As long as it's consistent I'll be happy, so I vote for whatever order is easiest to implement. There are smarter sorts out there but I don't think we want/need them.

workflows/general/alignment-post.wdl
a-frantz: Well it's good to know that STAR SE mode is being tested 😅  TBH I wasn't positive it would run correctly, but I guess XenoCP has been using it just fine.

I think we need documentation somewhere for the differences between these aligners. Not sure _here_ is the right place, but we could use that `external_documentation` key to point to the relevant docs. Does that documentation already exist?

tools/picard.wdl
```diff
@@ -109,23 +112,26 @@ task validate_bam {
             > ~{outfile_name}
 
         rc=$?
-        if [ $rc -le -1 ] || [ $rc -ge 4 ]; then
+        if [ $rc -le -1 ] || [ $rc -ge 4 ]; then  # TODO explain this
```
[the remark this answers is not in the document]
a-frantz: I don't know to either question haha.

> 0 = success
> 1 = process failure
> 2 = validation warnings
> 3 = validation errors

If that's the case, we are mishandling exit code `1`. I'd like to take a stab at refactoring and documenting this error handling. Where'd you pull these codes from?

I'd make a solid bet that WDL1.1 `returnCodes` don't change the behavior of `set -eo pipefail` (i.e. the task would still terminate on the first non-zero exit code). So I'm not sure how you're intended to use `returnCodes`.
It makes sense for the case of a single pipeless command that you might want "success" on alternative exit codes, but for any type of script with multiple potential fail points, I don't see how you can sensibly use the feature. (TBF I have not read their documentation on it).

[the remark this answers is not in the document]
adthrasher: I pulled the return codes from a discussion on the Picard issue tracker where they were discussing introducing them. It's possible that things change or were further refined. 

I misread the `set` previously. I thought you were turning one of the arguments off and on. Agreed, that as is, `-e` isn't going to be overridden by WDL.

workflows/general/alignment-post.wdl
adthrasher: Not any of which I am aware. What sort of documentation are you looking for though? Are you thinking in the tasks themselves? Or something in the XenoCP/workflows metadata to guide a choice?

workflows/chipseq/chipseq-standard.wdl
adthrasher: I would generally sort on basename, but maybe that's not a good approach. I think most of the grouping that you'd want would be handled by internal vs. external. However, it may be desirable to sort the full path, so that everything under `tools/` would be together and separate from something like `structs/`.

For the case of `../foo.wdl` and `../../bar/foo.wdl`, you'd have to rename them internally with an `as` correct? If so I'd sort them with the plain import then the `as` disambiguated one (or on the `as` values if both have them).

PR #98, opened 2023-06-07
feat(qc): Global Phred scores

a-frantz:
Adds a task/script that computes a variety of metrics about PHRED scores across the BAM. These are not currently incorporated into the per-sample MultiQC reports. That integration is waiting on the MultiQC task overhaul I have planned.

### DIFF
--- a/tools/util.wdl
+++ b/tools/util.wdl
@@ -417,3 +417,117 @@ task make_coverage_regions_beds {
         maxRetries: max_retries
     }
 }
+
+task global_phred_scores {
+    meta {
+        description: "Calculates statistics about PHRED scores of the input BAM."
+    }
+
+    input {
+        File bam
+        String outfile_name = basename(bam, ".bam") + ".global_PHRED_scores.txt"
+        Int memory_gb = 4
+        Int modify_disk_size_gb = 0
+        Int max_retries = 1
+    }
+
+    Float bam_size = size(bam, "GiB")
+    Int disk_size_gb = ceil(bam_size) + 10 + modify_disk_size_gb
+
+    command <<<
+        set -euo pipefail
+
+        BAM="~{bam}" OUTFILE="~{outfile_name}" python - <<END
+import os  # lint-check: ignore
+import pysam  # lint-check: ignore
+from collections import defaultdict  # lint-check: ignore
+
+bam_path = os.environ["BAM"]
+bam = pysam.AlignmentFile(bam_path, "rb")
+
+tot_quals = defaultdict(lambda: 0)
+mapped_quals = defaultdict(lambda: 0)
+unmapped_quals = defaultdict(lambda: 0)
+middle_tot_quals = defaultdict(lambda: 0)
+middle_mapped_quals = defaultdict(lambda: 0)
+middle_unmapped_quals = defaultdict(lambda: 0)
+for read in bam:
+    cur_quals = read.query_alignment_qualities  # array of phred scores
+    for i, qual in enumerate(cur_quals):
+        tot_quals[qual] += 1
+        if read.is_unmapped:
+            unmapped_quals[qual] += 1
+        else:
+            mapped_quals[qual] += 1
+
+    middle_pos = len(read.query_alignment_qualities) // 2  # middle base of read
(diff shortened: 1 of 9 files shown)


tools/util.wdl
```diff
@@ -417,3 +417,173 @@ task make_coverage_regions_beds {
...
+    middle_pos = len(cur_quals) // 2  # middle base of read
+    middle_score = cur_quals[middle_pos]
+    middle_tot_quals[middle_score] += 1
+    if read.is_unmapped:
+        middle_unmapped_quals[middle_score] += 1
+    else:
+        middle_mapped_quals[middle_score] += 1
+
+prefix = os.environ["PREFIX"]
+outfile = open(prefix + ".global_PHRED_scores.txt", "w")
```
a-frantz, line 478: Pretty small point, but should this have a `.tsv` ext instead of `.txt`? I think everywhere else we produce TSVs (mainly ngsderive) we use a `.txt` extension. I think I'd prefer to be more specific and change _all_ those to end in `.tsv`. But in the mean time, using `.txt` here is more consistent with the rest of our tasks.

I just see it as a matter of being more specific about the file contents is better. I don't see any reason (besides history) to stick with `.txt`.

adthrasher: I prefer `.tsv` for tab-delimited data as well.

a-frantz:
This task is slooow. One of the last tasks to complete in the QC workflow. If you see a way to speed this up, please comment on anything I'm doing that's inefficient.

tools/util.wdl
```diff
@@ -417,3 +417,173 @@ task make_coverage_regions_beds {
...
+
+    Float bam_size = size(bam, "GiB")
+    Int disk_size_gb = ceil(bam_size) + 10 + modify_disk_size_gb
+
+    String outfile_name = prefix + ".global_PHRED_scores.tsv"
+
+    command <<<
+        set -euo pipefail
+
+        BAM="~{bam}" PREFIX="~{prefix}" python3 - <<END
```
adthrasher, line 442: Just a general comment, should we be inlining python scripts? Is there some limit to complexity/length that should be a cutoff?

adthrasher:
> This task is slooow. One of the last tasks to complete in the QC workflow. If you see a way to speed this up, please comment on anything I'm doing that's inefficient.

I'm not sure there is much to do without going to something like multi-threading. You're iterating over each read and for 90%+ of the reads iterating over every base/quality score. Even then without timing the various components, I don't know how much speedup you would achieve. The time is likely to set up the Python data structures and then iterate over the quality scores. If computing on the quality scores is the slow component, then you could potentially read the chromosomes in different threads and then process each chunk independently.

tools/util.wdl
a-frantz: Good question. Historically, these inline Python scripts (of which there's a few now) were just implemented because it was a fast and easy. Now that I'm proliferating the practice into different places, I think it's a good idea to check whether it's a practice we want to maintain.

The most obvious alternative is to package these scripts with the `util` image. That should be an easy copy+paste without any significant changes to the Docker. Except I believe there's a script that runs on a `gtfparse` image pulled from BioContainers. That one would be trickier to change. IMO we wouldn't want to start maintaining another Docker image just to avoid some inlined Python. We could add `gtfparse` to the util image's dependencies, but then we run the risk of turning that image into a bloated "kitchen sink".

I think it would be a more maintainable and "prettier" solution to package these as scripts with `util`, but it does have a downside. These WDL tasks would become black boxes, unless you go and track down the source Docker directory.

I think the complexity/length angle could be a good way to decide what belongs inline VS packaged in Docker.

I have to chew on this problem some more, so why don't we push this question off to another PR/discuss offline.

TLDR; I don't think inlining Python is _terrible_, but we should avoid it just for "best practice" reasons. The question is what's the best alternative, and I don't have an immediate answer.

adthrasher: Agreed that it doesn't need to be resolved here. I'm not sure that I'm opposed to inline Python. I just think we may want to consider some guidance on length/complexity that would guide splitting it out into a standalone script. If the code is general purpose (such as this), I could see having a top-level `scripts` in the repo and the script lives there.

PR #99, opened 2023-06-22
feat: STAR rewrite

a-frantz:
2 major changes, and a few minor ones. Most importantly, there have been **NO CHANGES** to STAR's alignment or DB building behavior. (_Many_) more options have been exposed, but the defaults have been maintained as they were.

1st major change is exposing as many STAR options as reasonable in both DB building and Alignment. "As reasonable" is loosely defined as "doesn't require substantial changes to the WDL to support/doesn't give the user the option to disobey SAM specifications" (there are a surprisingly large amount of ways that STAR supports breaking the SAM spec. None of these were exposed.)
STAR uses strange parameter syntax and this required some awkward WDL. If anyone sees a more straightforward way to handle any of these parameters, please comment.
Documenting all this strange parameter syntax required getting a bit creative with the `parameter_meta` section. Thanks to much feedback from @adthrasher, I'm pretty happy with how the final documentation came out. But of course accepting feedback on that as well.
To preempt a criticism of the documentation I can see coming, while there are a handful of exceptions, most `parameter_meta` does _not_ include "what is the default?" This is for a two main reasons:
1a) We decided during the style guide writing process, it would be best not to include defaults which might change in the `parameter_meta` sections (which in theory are less likely to change). This prevents a situation where the documentation has default `foo` and the actual WDL has default `bar`.
1b) Most tasks have a limited number of options, and it's likely the WDL code with the default will be on the same screen as the documentation without the default. i.e. it's easy enough in most cases to just scroll down a few lines to find out what the default is. Unfortunately, the vast number of options for STAR breaks that assumption. I'll be the first to admit it is onerous to track down the default for parameter `X` while reading its `parameter_meta`. Which brings me to the next point:
2) We use `wdldoc` to prettify our in-code-file documentation. wdldoc parses the default from the WDL code and brings it up to the parameter description. If we start inserting defaults into the parameter_meta, defaults will appear twice in our MarkDown files. I've attached wdldoc's parsing of the new STAR file so we can see how it appears "prettified".

The wdldoc MarkDown is not so good IMO. There's a lot of nesting happening in the `parameter_meta` which gets flattened in a pretty ugly way.
[star.md](https://github.com/stjudecloud/workflows/files/12335397/star.md)
I think that's a problem to tackle another day. wdldoc just isn't quite sophisticated enough to handle these documentation cases elegantly.

2nd major change: the FASTQ entrypoint to `rnaseq-standard` has been made more user friendly. Instead of accepting a String with read group information (in STAR's bizarre format), the user now supplies an `Array[ReadGroup]`, where `ReadGroup` is a Struct.
I considered using this new Struct more prolifically, but that would be a decent amount of work for unclear gains. e.g. using `ReadGroup` as an input to `rnaseq-core`: `rnaseq-core` is not meant to be called directly by users. It's meant to be purely internal (should the name reflect this? `_rnaseq-core`?) so what's the point of making it more accessible? No one benefits from that accessibility. Different rationale for not using it in the STAR task; this tricky formatting is what's used by the underlying tool, and if using STAR from the commandline, you would be beholden to this formatting. So it's ok for our WDL task to do the same.
For those curious, here's how wdldoc documents the new entrypoint: just kidding. wdldoc won't parse it for some reason. I guess that is also on the docket for fixing now.


tools/star.wdl
```diff
@@ -87,17 +125,382 @@ task alignment {
...
+        outSJfilterDistToOtherSJmin: "minimum allowed distance to other junctions' donor/acceptor for: (1) non-canonical motifs, (2) GT/AG and CT/AC motif, (3) GC/AG and CT/GC motif, (4) AT/AC and GT/AT motif. Does not apply to annotated junctions."
+        alignSJstitchMismatchNmax: "maximum number of mismatches for stitching of the splice junctions (-1: no limit) for: (1) non-canonical motifs, (2) GT/AG and CT/AC motif, (3) GC/AG and CT/GC motif, (4) AT/AC and GT/AT motif"
+        clip3pAdapterSeq: {
+            description: "adapter sequences to clip from 3p of each mate. `left` applies to read one and `right` applies to read two.",
+            choices: {
+                None: "No 3p adapter trimming will be performed",
+                sequence: "Any nucleotide sequence of any length, matching the regex `/[ATCG]+/`",
```
adthrasher, line 141: When I first read this, it wasn't clear that the user provides a string matching `/[ATCG]+/` rather than the literal `sequence`.

```suggestion
                sequence: "A nucleotide sequence string of any length, matching the regex `/[ATCG]+/`",
```

tools/star.wdl
```diff
@@ -87,17 +125,382 @@ task alignment {
...
+            choices: {
+                None: "not used",
+                intronMotif: "strand derived from the intron motif. This option changes the output alignments: reads with inconsistent and/or non-canonical introns are filtered out."
+            }
+        }
+        outSAMattributes: {
+            description: "a string of desired SAM attributes, in the order desired for the output SAM. Tags can be listed in any combination/order. **[STAR defaults]**: `NH HI AS nM`. **[WDL default]**: `NH HI AS nM NM MD XS`.",
+            choices: {
+                NH: "number of loci the reads maps to: =1 for unique mappers, >1 for multimappers. Standard SAM tag.",
+                HI: "ultiple alignment index, starts with --outSAMattrIHstart (=1 by default). Standard SAM tag.",
```
adthrasher, line 176:
```suggestion
                HI: "multiple alignment index, starts with --outSAMattrIHstart (=1 by default). Standard SAM tag.",
```

tools/star.wdl
```diff
@@ -87,17 +125,382 @@ task alignment {
...
+                AS: "local alignment score, +1/-1 for matches/mismateches, score* penalties for indels and gaps. For PE reads, total score for two mates. Standard SAM tag.",
+                nM: "number of mismatches. For PE reads, sum over two mates.",
+                NM: "edit distance to the reference (number of mismatched + inserted + deleted bases) for each mate. Standard SAM tag.",
+                MD: "string encoding mismatched and deleted reference bases (see standard SAM specifications). Standard SAM tag.",
+                jM: "intron motifs for all junctions (i.e. N in CIGAR): 0: non-canonical; 1: GT/AG, 2: CT/AC, 3: GC/AG, 4: CT/GC, 5: AT/AC, 6: GT/AT. If splice junctions database is used, and a junction is annotated, 20 is added to its motif value.",
+                jI: "start and end of introns for all junctions (1-based).",
+                XS: "alignment strand according to --outSAMstrandField.",
+                MC: "mate's CIGAR string. Standard SAM tag.",
+                ch: "marks all segment of all chimeric alingments for --chimOutType WithinBAM output.",
```
adthrasher, line 185:
```suggestion
                ch: "marks all segments of all chimeric alignments for --chimOutType WithinBAM output.",
```

tools/star.wdl
```diff
@@ -87,17 +125,382 @@ task alignment {
...
+                cN: "number of bases clipped from the read ends: 5' and 3'"
+            }
+        }
+        outSAMunmapped: {
+            description: "output of unmapped reads in the SAM format.",
+            choices: {
+                None: "no output **[STAR default]**",
+                Within: "output unmapped reads within the main SAM file (i.e. Aligned.out.sam) **[WDL default]**"
+            }
+        }
```
adthrasher, line 195: I don't know that we strictly need to document the difference in the defaults here. I think `None` is nearly always the wrong option anyway.

tools/star.wdl
```diff
@@ -87,17 +125,382 @@ task alignment {
...
+            description: "2-pass mapping mode",
+            choices: {
+                None: "1-pass mapping **[STAR default]**",
+                Basic: "basic 2-pass mapping, with all 1st pass junctions inserted into the genome indices on the fly **[WDL default]**"
+            }
+        }
+        use_all_cores: "Use all cores? Recommended for cloud environments. Not recommended for cluster environments."
+        outFilterMismatchNoverLmax: "alignment will be output only if its ratio of mismatches to *mapped* length is less than or equal to this value"
+        outFilterMismatchNoverReadLmax: "alignment will be output only if its ratio of mismatches to *read* length is less than or equal to this value"
+        outFilterScoreMinOverLread: "same as outFilterScoreMin, but  normalized to read length (sum of mates' lengths for paired-end reads)"
```
adthrasher, line 301:
```suggestion
        outFilterScoreMinOverLread: "same as outFilterScoreMin, but normalized to read length (sum of mates' lengths for paired-end reads)"
```

tools/star.wdl
```diff
@@ -87,17 +125,382 @@ task alignment {
...
+        outFilterMultimapScoreRange: "the score range below the maximum score for multimapping alignments"
+        outFilterMultimapNmax: "maximum number of loci the read is allowed to map to. Alignments (all of them) will be output only if the read maps to no more loci than this value. Otherwise no alignments will be output, and the read will be counted as 'mapped to too many loci' in the Log.final.out. **[STAR default]**: `10`. **[WDL default]**: `20`."
+        outFilterMismatchNmax: "alignment will be output only if it has no more mismatches than this value"
+        outFilterScoreMin: "alignment will be output only if its score is higher than or equal to this value"
+        outFilterMatchNmin: "alignment will be output only if the number of matched bases is higher than or equal to this value"
+        scoreGap: "splice junction penalty (independent on intron motif)"
+        scoreGapNoncan: "non-canonical junction penalty (in addition to scoreGap)"
+        scoreGapGCAG: "GC/AG and CT/GC junction penalty (in addition to scoreGap)"
+        scoreGapATAC: "AT/AC and GT/AT junction penalty  (in addition to scoreGap)"
```
adthrasher, line 327:
```suggestion
        scoreGapATAC: "AT/AC and GT/AT junction penalty (in addition to scoreGap)"
```

tools/star.wdl
```diff
@@ -130,67 +530,191 @@ task alignment {
...
-             --outFilterScoreMinOverLread 0.66 \
-             --outFileNamePrefix ~{prefix + "."} \
-             --twopassMode Basic \
-             --limitBAMsortRAM ~{memory_limit_bytes} \
-             --outSAMattrRGline $(cat read_groups_sorted.txt)
+        read -ra read_one_args < read_one_fastqs_sorted.txt
+        read -ra read_two_args < read_two_fastqs_sorted.txt
+        read -ra read_group_args < read_groups_sorted.txt
+        STAR --readFilesIn "${read_one_args[@]}" "${read_two_args[@]}" \
+            --readFilesCommand "gunzip -c" \
```
adthrasher, line 554: Not to add complications, but should we be assuming that `read_one_fastqs` contains gzipped files? We could rename it to `read_one_fastqs_gz` for consistency.

adthrasher: Or we could parameterize the read command.

tools/star.wdl
```diff
@@ -13,13 +13,41 @@ task build_star_db {
...
+        use_all_cores: "Use all cores? Recommended for cloud environments. Not recommended for cluster environments."
+        genomeChrBinNbits: "=log2(chrBin), where chrBin is the size of the bins for genome storage: each chromosome will occupy an integer number of bins. For a genome with large number of contigs, it is recommended to scale this parameter as min(18, log2[max(GenomeLength/NumberOfReferences,ReadLength)])."
+        genomeSAindexNbases: "length (bases) of the SA pre-indexing string. Typically between 10 and 15. Longer strings will use much more memory, but allow faster searches. For small genomes, the parameter `--genomeSAindexNbases` must be scaled down to `min(14, log2(GenomeLength)/2 - 1)`."
+        genomeSAsparseD: "suffux array sparsity, i.e. distance between indices: use bigger numbers to decrease needed RAM at the cost of mapping speed reduction."
+        genomeSuffixLengthMax: "maximum length of the suffixes, has to be longer than read length. -1 = infinite."
+        sjdbOverhang: "length of the donor/acceptor sequence on each side of the junctions, ideally = (mate_length - 1). **[STAR default]**: `100`. **[WDL default]**: `125`."
```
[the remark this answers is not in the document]
a-frantz: I'm open to alternatives, but IMO we should be explicit about where we differ from a tool default (particularly with the assumption we'll be moving to a "tool resource" model at some point in the future).

I don't like the idea of silently changing tool behavior. If I was a well-versed STAR user and new to using this repository, I'd be annoyed to find that the default behavior I assumed was wrong

tools/star.wdl
a-frantz: This was a good catch. Path of least resistance is go back to using the `_gz` suffix and specifying we expect gzipped files. That would be easier (and less error prone) than dealing with multiple/arbitrary read commands.

adthrasher (approved):

PR #102, opened 2023-07-11
feat: MultiQC refactor

a-frantz:
All of the dynamic MultiQC config has been stripped away. Instead, users will be expected to:
a) run the workflow (and the `make_qc_reference.wdl` wf) with defaults
b) supply a custom written YAML file (easy to use the new one uploaded to this repo as a template/base)

While this loses some "ease of use" by stripping the dynamic configuration (mainly of coverage region labels), a great amount of new functionality is now possible. Users can supply custom analysis result files with a custom YAML config and get all our analyses reported side by side.


tools/samtools.wdl
```diff
@@ -217,38 +217,43 @@ task subsample {
...
-            true
+
+            # Assumes being called by one of our workflows
+            # or the default 'prefix' was used.
+            # If not, 'sample_name' will equal 'prefix'
+            sample_name=$(basename ~{prefix} '.subsampled')
+            {
+                echo -e "sample\toriginal read count"
+                echo -e "$sample_name\t-"
+            } > "$sample_name".orig_read_count.tsv
```
a-frantz, line 251: There's a problem here with this task, which I believe requires this ugly assumption.
The workflows which call this task (at least QC (which is the only one using this little TSV report), I haven't checked the others), take a string input of `prefix`, which can be arbitrary and unrelated to the input BAM file name.
If the BAM gets subsampled, `.subsampled` will be added to the prefix for _every_ subsequent file. That's controlled at the workflow level: in the call to `subsample` the input `prefix = prefix + '.subsampled'` is passed. So if we subsample the BAM, we just use `prefix` for the filename.
But if we _don't_ subsample, then the `.subsampled` suffix isn't added to subsequent files. So we have to `basename ~{prefix} '.subsampled'` here, so that the filename matches the rest of the WF outputs.

If a user calls this task directly, the `.subsampled` suffix in `~{prefix}` may not be present. I _think_ in that case it's appropriate to use the same filename regardless of whether the subsampling happened. My assumption would be "if I'm a user supplying some arbitrary prefix, I expect the output file to use that prefix regardless". But I feel uneasy making that assumption. I don't how to get around it though, without complicating the input section: e.g. adding a `subsampled_prefix` and ` nonsubsampled_prefix` parameters. That seems awfully ugly.

I think this is the best (only?) way to handle this. I don't know how to handle it without this assumption. Comment if you have any other ideas 🤷

adthrasher:
> @adthrasher Very minor point I don't think matters very much: in your test PR you've named all YML/YAML files with a `.yml` extension. The docs for MultiQC use a `.yaml` extension and I was consistent with that. I don't know if there's a community consensus. Do we care about having consistent YML/YAML extensions? If so which should it be?
> 
> MultiQC definitely auto-picks-up the file `multiqc_config.yaml`. I don't know about `multiqc_config.yml`. But I guess that doesn't really matter because we point to the file explicitly using the `--config /path/` param.
> 
> My preference for this file would be to use `.yaml` so we're consistent with what MultiQC uses. I'm ok with using `.yml` in our test files if that's the community consensus or if `pytest` has a preference.

https://yaml.org/faq.html
```
Is there an official extension for YAML files?

Please use ".yaml" when possible.
```

That suggests I should update the files in the test PR.

tools/samtools.wdl
adthrasher: Should we just alter this task? Remove ".subsampled" from the prefix declaration. Then the usages of `~{prefix}` would become `~{prefix}.subsampled`. 
```
        String prefix = basename(bam, ".bam") + ".subsampled"
```

adthrasher: That seems reasonable to me. Our opinion is that the output should always include `.subsampled` to indicate the subsampling. This change would enforce that.

adthrasher: Or for simplicity, after the input block do something like 
```
String suffixed = ~{prefix} + ".subsampled"
```
Then use `prefix` and `suffixed` where appropriate.

workflows/qc/inputs/multiqc_config.yaml
```diff
@@ -0,0 +1,53 @@
+extra_fn_clean_exts:
+    - '.ValidateSamFile'
+    - '.whole_genome'
+    - '.exon'
+    - '.CDS'
+    - '.MarkDuplicates'
+log_filesize_limit: 1000000000
+fastqc_config:
+  fastqc_theoretical_gc: "hg38_genome"
```
adthrasher, line 9: So we're assuming that we're always using GRCh38. It'd be nice to make this configurable, but I don't know if there is a better way to do that. Maybe we just name the file `multiqc_config_hg38.yaml` and then we could have preconfigured choices for appropriate genomes. I wouldn't actually create the others at this point.

adthrasher (approved):

workflows/qc/inputs/multiqc_config.yaml
a-frantz: I like renaming this and hosting multiple in the future. We're going to want target specific configs in the near future (me and David have discussed differentially altering how the coverage sections appear for our 3 targets (WGS/WES/RNA)).

Biggest problem with this approach is that the number of configs here might get out of hand. I'd say we can plan to host `multiqc_config_[WGS,WXS,RNA-Seq]_hg38.yaml` in the near future. I think hosting 3 configs is reasonable. But do we really want that number increasing further down the line?

I haven't actually tried visualizing it yet, but in the docs there's `fastqc_theoretical_gc: "hg38_transcriptome"` (that might not be the correct keyword). Not sure what it looks like but it's been on my agenda to check out. That could present another config option to host. Then we could be looking at hosting 4 nearly identical configs.

I think statically hosting these is fine for now. But it's worth considering for the future what our line will be. At what point do we stop putting these different configs up?

tools/samtools.wdl
a-frantz: > That seems reasonable to me. Our opinion is that the output should always include .subsampled to indicate the subsampling. This change would enforce that.

I was initially hesitant to force the use of `.subsampled` but I think you're right. It's appropriate for us to enforce and makes the code a lot cleaner. A user can always rename the files if they have a problem with our schema.

workflows/qc/inputs/multiqc_config.yaml
adthrasher: I think we would host any configs that we're using. Beyond that, unless there is some glaring omission that would be generally relevant, I wouldn't host others.

PR #103, opened 2023-07-12
Task test cases

adthrasher:
Adding test cases for each of our WDL tasks. This initial pass executes the task and checks that the appropriate output exists. For plain text outputs, some tasks have additional content checking. A future effort will be required to add content checking to additional output formats, such as BAM.


tests/tools/test_kraken.yml
a-frantz, line 1: Trying to future proof: we could conceivably have a name conflict on `build_db_miniwdl`. I think we want to prefix all these test names with `kraken_*` just to be safe.

tests/tools/input_json/multiqc.json
```diff
@@ -0,0 +1,4 @@
+{
+    "input_files": ["tests/tools/input/test.bwa_aln_pe.bam", "tests/tools/input/test.bwa_aln_pe.readlength.txt"],
```
a-frantz, line 2: Are we intentionally passing in a file that gets ignored by MultiQC? (any `*.bam` is just being passed over)

tests/tools/test_ngsderive.yml
```diff
@@ -0,0 +1,55 @@
...
+- name: encoding_miniwdl
+  tags:
+    - miniwdl
+    - ngsderive
+  command: >-
+    miniwdl run -d test-output/. --copy-input-files --task encoding -i tests/tools/input_json/ngsderive_encoding.json tools/ngsderive.wdl
+  files:
+    - path: test-output/out/encoding_file/test.bwa_aln_pe.encoding.txt
+      contains:
+        - "ASCII range: 74-74"
```
a-frantz, line 44: The only PHRED encoding present here is `74`? Where did this test BAM come from?

tests/tools/test_samtools.yml
```diff
@@ -0,0 +1,94 @@
...
+
+- name: split_miniwdl
+  tags:
+    - miniwdl
+    - samtools
+  command: >-
+    miniwdl run -d test-output/. --copy-input-files --task split tools/samtools.wdl bam="https://github.com/stjude/CICERO/raw/master/test/data/input/test.bam"
+  files:
+    - path: test-output/out/split_bams/0/
+    - path: test-output/out/split_bams/1/
```
a-frantz, line 18: Are these supposed to be just directories and not full paths?

tests/tools/test_util.yml
```diff
@@ -0,0 +1,107 @@
...
+  stdout:
+    contains_regex:
+      - '"download.downloaded_file": ".*test-output/out/downloaded_file/license.txt"'
+  files:
+    - path: test-output/out/downloaded_file/license.txt
+      md5sum: cf3575bd84ab3151c7e9700b5f1a9746
+      contains:
+        - "MIT License"
+
+- name: get_read_groups_miniwdl
```
a-frantz, line 17: OOS for this PR, but just wondering "why is this task in `util.wdl` instead of `samtools.wdl`"?

Tangent: I'm not a fan of how long either of these files are (`util` and `samtools`). I'd consider breaking both of them up into more specific or specialized files. The length of these files is making it annoying to find things IMO.

tools/multiqc.wdl
```diff
@@ -39,8 +39,11 @@ task multiqc {
         echo "~{sep('\n', input_files)}" > file_list.txt
 
         # Start YAML generation
-        echo "extra_fn_clean_exts:" > multiqc_config.yaml
-        
+        touch multiqc_config.yaml
+        if [ "~{(length(extra_fn_clean_exts) > 0)}" = "true" ] || [ "~{(length(mosdepth_labels) > 0)}" = "true" ]; then
```
a-frantz, line 43: Nice catch on this. But bad timing on your fix, because I'm opening a PR which removes _all_ this logic haha.

a-frantz (changes requested):
First off, great job. I'm super excited about having these tests.

* Thanks to the nifty rendering of Github, it was obvious to me that the new files are inconsistent in regards to whether they have a `\n` at the EOF. I prefer "with newline at EOF", but I'd defer to whatever the filetype standard is. For example, off the top of my head, JSON is discouraged to have a newline. I don't know about YML. Just looking for consistency, my preference for "with" is pretty weak.
* We talked about this offline, but I'm reiterating that I'm not a fan of ending every task name with `_miniwdl`. What's your rationale for putting it after every task name? It just seems pointlessly verbose to me, and I'd rather not have it.
* This could wait till later, but I'd like to get a bit fancier with our tagging system. For example I'd like all the reference building tasks to get a `reference` tag. QC tasks get a `qc` tag. etc. There are some other useful tags we could add; we could have a `slow`/`long_running` tag. Lots of stuff we could do with the tagging system. Doesn't need to happen here, just throwing it out as a future TODO.
* Speaking of TODOs, you left a few comments about things that are currently broken. Fine with leaving them broken for now, but I'd like those comments to get prefixed with `TODO`. This is entirely to facilitate my laziness and bad memory 😂  Sometimes when I'm looking to switch tasks I'll just ctrl-f "TODO" on a repo and tackle something random that pops up. Prefixing those comments with "TODO" makes it more likely I remember to fix them. Otherwise they might drop off my radar.

adthrasher:
> * We talked about this offline, but I'm reiterating that I'm not a fan of ending every task name with `_miniwdl`. What's your rationale for putting it after every task name? It just seems pointlessly verbose to me, and I'd rather not have it.

I agree with you. Since each has a `miniwdl` tag, I don't see any need for the suffix `_miniwdl`, unless we were ever to add tests with another runner, such as `cromwell`, but we can deal with that in the future if need be.

adthrasher:
> * Speaking of TODOs, you left a few comments about things that are currently broken. Fine with leaving them broken for now, but I'd like those comments to get prefixed with `TODO`. This is entirely to facilitate my laziness and bad memory 😂  Sometimes when I'm looking to switch tasks I'll just ctrl-f "TODO" on a repo and tackle something random that pops up. Prefixing those comments with "TODO" makes it more likely I remember to fix them. Otherwise they might drop off my radar.

I can add that. I didn't add `TODO`s here, because the `TODO` task is actually in the WDL, but I suppose you'd have to update the tests anyway to remove the broken notice.

tests/tools/test_kraken.yml
adthrasher: I was trying to avoid the redundant prefixes of the tool name, but it is definitely needed in some tasks.

tests/tools/test_ngsderive.yml
adthrasher: It's synthetic data from `ngs generate`. Apparently, it doesn't simulate quality scores realistically by default.

tests/tools/test_samtools.yml
adthrasher: Good catch, I changed the input file right before I opened the PR. Apparently, I left off the updated file names.

tests/tools/test_util.yml
adthrasher: The underlying task has some special logic on how it formats the output. It seems more appropriate to be in `util`, because it is manipulating the `samtools` output so much. The reformatting portion is also `star`-specific.

```
        if ~{format_for_star}; then
            samtools view -H ~{bam} \
                | grep "@RG" \
                | cut -f 2- \
                | sed -e 's/\t/ /g' \
                | awk '{print}' ORS=' , ' \
                | sed 's/ , $//' > read_groups.txt
        else
            samtools view -H ~{bam} | grep "@RG" > read_groups.txt
        fi
```

tests/tools/test_ngsderive.yml
a-frantz: That's fine then. I think this is the first time `encoding` has flagged something other than PHRED+33 encoding (that I didn't synthesize for testing). Nice to see it catching something "real" for once!

a-frantz (approved):
LGTM! 🚀

a-frantz (approved):

PR #104, opened 2023-07-19
feat: fast mode for global PHRED score task

a-frantz:
By default, only compute metrics based on the middle base of each read (as opposed to old behavior of additionally calculating for _every_ base in the read)


tools/util.wdl
```diff
@@ -460,11 +465,12 @@ for read in bam:
 
     cur_quals = read.query_alignment_qualities  # array of phred scores
     for qual in cur_quals:
-        tot_quals[qual] += 1
-        if read.is_unmapped:
-            unmapped_quals[qual] += 1
-        else:
-            mapped_quals[qual] += 1
+        if not only_mid:
```
adthrasher, line 468: It may just be the GitHub browser, but assuming this is the entire `for` loop, shouldn't the `if` go outside?

adthrasher (approved):

tools/util.wdl
adthrasher: That is:
```
if not only_mid:
    for qual in cur_quals:
```

a-frantz: You are correct! Logically equivalent but **not** equivalent for efficiency. Fixing now.

PR #105, opened 2023-07-19
tests: Additional tests

adthrasher:
Adding more in-depth checks of file contents, particularly for BAM and FASTQ format output.


tests/tools/test_samtools.py
```diff
@@ -0,0 +1,51 @@
...
+    assert len(read_groups) == 2
+    assert read_groups[0] == "test2"
+    assert read_groups[1] == "test.bwa_aln_pe"
+
+@pytest.mark.workflow('samtools_collate', 'samtools_collate_to_fastq')
+def test_samtools_collate(workflow_dir):
+    bam = pathlib.Path(workflow_dir, 'test-output/out/collated_bam/test.bwa_aln_pe.collated.bam')
+    samfile = pysam.AlignmentFile(bam, "rb")
+
+    reads = list(samfile.head(n=100))
```
adthrasher, line 40: I think this BAM is small enough that we could read the entire thing. Is that worth a little runtime penalty? If some rare case caused it to be not correctly `collated`, we probably wouldn't catch it anyway with a BAM this size.

tools/samtools.wdl
```diff
@@ -102,12 +107,34 @@ task split {
...
+                    -h 0 \
+                    -n 1 \
+                    "$out_bam" \
+                    > first_read.sam
+
+                if [ ! -s first_read.sam ]; then
+                    >&2 echo "No reads are in output BAM $out_bam!"
+                    >&2 echo "This is likely caused by malformed RG records."
+                    rm first_read.sam
+                    exit 42
```
adthrasher, line 133: Can you modify this to wait and exit after checking all BAMs? I doubt there would be multiple empty read groups, but we should probably catch that case in a single check.

a-frantz: Good idea

tests/tools/test_samtools.py
a-frantz: We might as well read the whole thing. I doubt it will be noticeable given the ~15min this already takes.

a-frantz:
@adthrasher the rollback did not fix the CI unfortunately

a-frantz (approved):
This looks great @adthrasher ! I'll let you merge. We've still got more tests to add, but I think iteratively merging bits when we have bandwidth is going to be better than waiting for all our desired testing to happen at once.

PR #106, opened 2023-08-03
feat(fq): expose lint options

a-frantz:
In addition to exposing `fq lint` options, there are 2 other minor changes.
1) All `read_*_fastq_gz` input variables to `fq` tasks have had the `_gz` suffix dropped. `fq` accepts raw input as well, and auto detects which the file is.
2) The `basename()` regex for FASTQ files that is used in multiple files has been updated to catch lowercase `_r1`/`_r2` extensions.

### DIFF
--- a/tools/fq.wdl
+++ b/tools/fq.wdl
@@ -12,8 +12,8 @@ task fqlint {
     }
 
     parameter_meta {
-        read_one_fastq: "Input FASTQ with read one"
-        read_two_fastq: "Input FASTQ with read two"
+        read_one_fastq: "Input FASTQ with read one. Can be gzipped or uncompressed."
+        read_two_fastq: "Input FASTQ with read two. Can be gzipped or uncompressed."
         disable_validator_codes: {
             description: "Array of codes to disable specific validators",
             choices: {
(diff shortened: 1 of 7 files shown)


tools/fq.wdl
```diff
@@ -8,24 +8,58 @@ version 1.1
...
     meta {
-        description: "This WDL task performs quality control on the input FASTQ pairs to ensure proper formatting."
+        description: "This WDL task performs quality control on the input FASTQs to ensure proper formatting."
     }
 
     parameter_meta {
-        read_one_fastq_gz: "Input FASTQ with read one"
-        read_two_fastq_gz: "Input FASTQ with read two"
+        read_one_fastq: "Input FASTQ with read one"
+        read_two_fastq: "Input FASTQ with read two"
```
adthrasher, line 16: I would just note in the documentation here that gzipped or uncompressed inputs are supported.

tools/fq.wdl
```diff
@@ -8,24 +8,58 @@ version 1.1
...
+            ]
+        }
+        paired_read_validation_level: {
+            description: "Only use paired read validators up to a given level",
+            choices: [
+                "low",
+                "medium",
+                "high"
+            ]
+        }
```
adthrasher, line 45: I didn't look at the fqlint documentation, but these aren't exclusive, right? You can specify `high` and then disable specific validators.

adthrasher (approved):

tools/fq.wdl
adthrasher: I don't think we need to specify in our documentation. I was just making sure.

PR #108, opened 2023-08-18
feat: endedness and ngsderive update

a-frantz:
title


adthrasher (approved):

adthrasher:
> @adthrasher re: this comment [#110 (comment)](https://github.com/stjudecloud/workflows/pull/110#discussion_r1316271013)
> 
> How do we like the voice I used for these `meta.description` and `outputs`? i.e.
> 
> * `meta.description` starts with a verb. All tasks are _doing_ something, and should be phrased in the active tense (grammar's not my strong suit, I think these are examples of active tense).
> * `outputs` are a little trickier. For this, I'd say "First 'sentence' should be a sentence fragment (just the subject)."
>   
>   * `strandedness_file: "TSV file containing the ngsderive strandedness report"`
>     
>     * This is an incomplete sentence. I think it's all we need for most outputs. Can be optionally expanded with a full sentence(s) after the initial fragment.
> 
> IMO this 2 rules (could be formalized with better language) would sufficiently differentiate these strings that appear next to each other.
> 
> Adopting these two styles does beg a question: "do we need to formalize the grammar used for `parameter_meta`?" I think the parameter meta is fine as is. But thought I'd rise the question.

I do think using the active voice for the descriptions is a good approach. I'm less convinced about the `outputs`. I think it is fine to write a fragment for those, but I don't think that should be a requirement.

At this point, I would avoid formalizing this, but I think it's a good approach and we can continue to refine it.

tools/fq.wdl
```diff
@@ -62,7 +62,7 @@ task fqlint {
     Float read2_size = size(read_two_fastq, "GiB")
 
     Int memory_gb = (
-        ceil((read1_size + read2_size) * 0.2) + 4 + modify_memory_gb
+        ceil((read1_size + read2_size) * 0.25) + 4 + modify_memory_gb
```
adthrasher, line 65: That seems like a high memory requirement for this task. I assume you've encountered a failure. Looking at the list of validators, the only one that should require much memory is the duplicate name check.

tools/ngsderive.wdl
```diff
@@ -32,14 +55,19 @@ task infer_strandedness {
...
-            -g ~{gtf} \
+            -g ~{gene_model} \
             "$CWD_BAM" \
             > ~{outfile_name}
 
-        awk 'NR > 1' ~{outfile_name} | cut -d$'\t' -f5 > strandedness.txt
+        if split_by_rg; then
+            awk 'NR > 1' ~{outfile_name} | cut -d$'\t' -f6 > strandedness.txt
+        else
+            awk 'NR > 1' ~{outfile_name} | cut -d$'\t' -f5 > strandedness.txt
```
adthrasher, line 69: Do you need to specify `-d$'\t`? The tab delimiter should be the default for `cut`.

tools/ngsderive.wdl
```diff
@@ -190,16 +268,38 @@ END
...
         disk: "~{disk_size_gb} GB"
-        docker: 'quay.io/biocontainers/ngsderive:2.4.0--pyhdfd78af_0'
+        docker: 'quay.io/biocontainers/ngsderive:3.3.2--pyhdfd78af_0'
         maxRetries: max_retries
     }
 }
 
 task junction_annotation {
+    meta {
+        description: "Annotates junctions found in an RNA-Seq BAM as known, novel, or partially novel"
```
adthrasher, line 278: Do you think it is worth defining `known`, `novel`, and `partially novel` here or in a `help` text?

adthrasher: To add to that, I think the definitions are missing in the `ngsderive` docs. It might be better there. RSeQC states: 

> 
> Annotated (known): The junction is part of the gene model. Both splice sites, 5’ splice site (5’SS) and 3’splice site (3’SS) are annotated by reference gene model.
> 
> Complete_novel: Both 5’SS and 3’SS are novel.
> 
> Partial_novel: One of the splice site (5’SS or 3’SS) is novel, and the other splice site is annotated

adthrasher (approved):

tools/fq.wdl
a-frantz: Yes it is definitely overkill for the majority of cases. But I've run into a _very_ small number of failures at the `0.2` scale (which is also more than what most cases need). So following the ethos for our default resource requirement settings (set defaults so we don't have failures in production/never need to think about it), I've been forced to set it this high.

I don't think we have much of a choice in the matter, unless we revise our resource policies. This is how much memory we need to assign so we guarantee no failures in production.

We could scale this back to a flat ~10gb which would be suitable for roughly 90% of cases? That's a gut estimate, no data to back it up. But then we'd need to override and set it higher to ~50gb for those corner cases. Do we want to deal with that? We have to keep in mind that there's no way to predict which samples are the problem samples ahead of time. Size is a very loose corollary.

tools/ngsderive.wdl
a-frantz: Good point. We can _cut_ that extra argument

a-frantz: Ah. Also just now realizing this won't work for the first case no matter what. We're going to need to add a `grep 'overall'` _after_ v4 of ngsderive is released (in progress [here](https://github.com/stjudecloud/ngsderive/pull/139)). So this will just be broken till that release. There's no sensical value to put in `strandedness.txt` with the current latest release of ngsderive when `--split-by-rg` is true. I think that's fine for now. We aren't using the problem configuration in prod.

I will make a note of the broken state though.

tools/ngsderive.wdl
a-frantz: I'll add to the ngsderive docs, and link to those docs here using the `external_help` key 👍

a-frantz: Definitions have been added to the ngsderive v4 PR. [specific commit with those definitions](https://github.com/stjudecloud/ngsderive/pull/139/commits/625210e30ad2aea2b3ca8bcf0185db223a966981)

PR #111, opened 2023-09-20
Chipseq updates

adthrasher:
Updates to reflect changes to SEAseq and add support for SEAseq's paired-end mode.


tools/samtools.wdl
```diff
@@ -462,9 +463,10 @@ task bam_to_fastq {
         String f = "0"
         String F = "0x900"
         String G = "0"
-        Boolean paired_end = true
+        Boolean append_read_number = true
         Boolean interleaved = false
         Boolean output_singletons = false
+        Boolean paired_end = true
```
a-frantz, line 469: Just curious, did you reorder these to be alphabetical? The style-guide leaves it up to the author how to order parameters of the same type. My process for ordering things is typically:
1) follow the order of the underlying tools documentation
2) what feels most important for a user to see should be towards the top
    * obviously incredibly subjective here. But with this I'd say `paired_end` should be first in this block.

I'm fine with this being alphabetical, but I don't think any other inputs are. This will likely be the odd one out if left alphabetical, but we intentionally punted the question of "order once type order is satisfied", so I can't complain about anything you do here.

tools/samtools.wdl
```diff
@@ -637,3 +652,61 @@ task collate_to_fastq {
...
 }
+
+task fixmate {
+    meta {
+        description: "This WDL task runs Samtools fixmate on the input BAM file. This fills in mate coordinates and insert size fields."
+    }
+
+    parameter_meta {
+        bam: "Input BAM format file to add mate information. Must be name-sorted or name-collated."
+        prefix: "Prefix for the output file. The extension will be added."
```
a-frantz, line 663:
```suggestion
        prefix: "Prefix for the output file. The extension `.bam` will be added."
```

tools/samtools.wdl
```diff
@@ -637,3 +652,61 @@ task collate_to_fastq {
...
+        extension: "File format extension to use for output file. Allowable types: [.sam, .bam, .cram]"
+        add_cigar: "Add template cigar ct tag"
+        add_mate_score: "Add mate score tags. These are used by markdup to select the best reads to keep."
+        disable_proper_pair_check: "Disable FR proper pair check"
```
a-frantz, line 667: IDK what this means. Can we elaborate or add an `external_help` link?

tools/samtools.wdl
a-frantz: Suggested this before seeing the next line. What I suggested is wrong; we need to a bit more verbose here

tools/samtools.wdl
```diff
@@ -637,3 +652,61 @@ task collate_to_fastq {
...
+
+task fixmate {
+    meta {
+        description: "This WDL task runs Samtools fixmate on the input BAM file. This fills in mate coordinates and insert size fields."
+    }
+
+    parameter_meta {
+        bam: "Input BAM format file to add mate information. Must be name-sorted or name-collated."
+        prefix: "Prefix for the output file. The extension will be added."
+        extension: "File format extension to use for output file. Allowable types: [.sam, .bam, .cram]"
```
a-frantz, line 664: IDK how I feel about allowing SAM. IMO we should only be dealing with compressed forms of this data type. SAM just tends to be absurdly large and ultimately impractical.

And by eliminating `.sam`, this can be changed to a Boolean. Makes life a little easier.

a-frantz: If you disagree and want to support `.sam`, this line should be updated to use `choices: {}` syntax.

workflows/chipseq/chipseq-standard.wdl
```diff
@@ -105,15 +107,17 @@ workflow chipseq_standard {
         max_retries=max_retries
     }
 
-    scatter (pair in zip(bam_to_fastqs.read1s, read_groups)){
+    scatter (tuple in zip(zip(bam_to_fastqs.read1s, bam_to_fastqs.read2s), read_groups)){
```
a-frantz, line 110: Are you positive this works for all combinations of inputs? I thought WDL enforced "every array in a zip must be of equal length". I think this line would break under certain circumstances. Or maybe my understanding of `zip` is wrong.

workflows/chipseq/chipseq-standard.wdl
```diff
@@ -159,7 +163,16 @@ workflow chipseq_standard {
...
     }
-    call picard.validate_bam { input: bam=markdup.mkdupbam, max_retries=max_retries }
+    call picard.validate_bam { input: 
+        bam=markdup.mkdupbam,
+        ignore_list=["MISSING_PLATFORM_VALUE",
+                    "INVALID_PLATFORM_VALUE",
+                    "INVALID_MAPPING_QUALITY",
+                    "MATES_ARE_SAME_END",
+                    "MISMATCH_FLAG_MATE_NEG_STRAND",
+                    "MISMATCH_MATE_ALIGNMENT_START"],
```
a-frantz, line 173: These highlighted 4 feel like things that shouldn't be ignored. Are BAMs really coming out of SEAseq with all these problems? Should that be addressed instead of ignored here?

workflows/general/bam-to-fastqs.wdl
```diff
@@ -54,18 +54,39 @@ workflow bam_to_fastqs {
...
+        }
+    }
+
+    Array[File?] r1 = select_first([bam_to_fastq.read_one_fastq_gz, bam_to_fastq_se.interleaved_reads_fastq_gz])
+    Array[File] read1s_ = select_all(r1)
+
+    Array[File?] read2s_ = select_first([bam_to_fastq.read_two_fastq_gz, bam_to_fastq_se.read_two_fastq_gz])
+
+     scatter (reads in 
+        zip(read1s_, read2s_)
```
a-frantz, line 89: Same question about `zip`
https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#arraypairxy-ziparrayx-arrayy

Seems to me this would break in SE mode

a-frantz (changes requested):

tools/samtools.wdl
adthrasher: I did choose alphabetical. That said, in this case, I think it makes sense as `paired_end` is an input for the task, where the other arguments more closely correspond with inputs to the underlying tool.

tools/samtools.wdl
adthrasher: I'll have to look it up. I also am not sure what it means. I assume that it requires there to be one forward mapped read and one reverse mapped read to make a proper pair, but that isn't clear from the help text from `samtools`.

workflows/chipseq/chipseq-standard.wdl
adthrasher: I need to run some additional test data through, but for both single-end and paired-end data so far, that argument works.

workflows/chipseq/chipseq-standard.wdl
adthrasher: It's an issue with `bowtie`. It looks like `bowtie` is emitting multiple alignments, that have equal alignment scores. Instead of picking one to be primary, it just outputs all of them, without marking any as secondary. Then `picard` fails to match the read records properly. It should be able to use the mate start position information to find the proper mate, but it doesn't. So you get these errors.

workflows/general/bam-to-fastqs.wdl
adthrasher: It was already using a `zip`. I just changed the parameters. I think the underlying bam-to-fastq logic emits an `Array[File?]`. The array would have the correct length, it would just be full of empty elements.

tools/samtools.wdl
a-frantz: Ok. The style-guide leaves it up to individual authors, so if you want to write tasks using alphabetic sort at this level, that's fine with me. But I'm going to continue using my (highly subjective) method of sorting inputs, so there will be some degree of inconsistency. I think that's fine. The fine-grain sorting here is fairly inconsequential.

workflows/chipseq/chipseq-standard.wdl
a-frantz: Interesting. Is this a known and logged bug in `bowtie`? Is the SEAseq team aware this is happening? And is this why you started implementing `samtools fixmate`? Would that fix these errors? (sorry for the bombardment of questions)

I'd much rather fix these errors over simply ignoring them. If I'm understanding correctly, this update (and possible the current version) are outputting BAMs which don't follow the SPEC. I'm not a fan of passing the buck off to `bowtie` and saying "not a problem with our code". This BAMs will likely fail downstream. I don't think I'm comfortable just glossing over these errors.

My two cents: we either need an update/rollback on `bowtie` so this isn't a problem in the first place, or we need a tool like `samtools fixmate` to address the bowtie problems before we can call these BAMs "harmonized".

workflows/chipseq/chipseq-standard.wdl
a-frantz: Can you point me to an input JSON you're using for both PE and SE? I want to walk myself through some concrete examples, because I'm confused how this `zip` here is working. I'm sure once I can see the "filled in" logic it'll be clear.

workflows/chipseq/chipseq-standard.wdl
adthrasher: I believe the solution is to use `bowtie2`. I don't think there is any support or development of `bowtie` at this point. I started implementing `samtools fixmate` for a different issue, so I didn't try it here.

workflows/chipseq/chipseq-standard.wdl
adthrasher: Sure. I just moved these from my home directory. I did a `sed` to update the paths. Let me know if I missed one.

```
/research_jude/rgs01_jude/groups/zhanggrp/projects/SJCloud/common/athrashe/chipseq_workflow/chipseq_input.json
/research_jude/rgs01_jude/groups/zhanggrp/projects/SJCloud/common/athrashe/chipseq_workflow/chipseq_paired_input.json
```

a-frantz:
@adthrasher can you split the `samtools.wdl` changes off into a stand-alone PR? I think those changes are good and should be merged. They don't have to wait on the ChIP-Seq questions to be resolved.

adthrasher:
> @adthrasher can you split the `samtools.wdl` changes off into a stand-alone PR? I think those changes are good and should be merged. They don't have to wait on the ChIP-Seq questions to be resolved.

Moved to #119

workflows/chipseq/chipseq-standard.wdl
```diff
@@ -105,29 +107,31 @@ workflow chipseq_standard {
...
         call util.add_to_bam_header { input:
             bam=chosen_bam,
-            additional_header=pair.right,
+            additional_header=tuple.right,
             max_retries=max_retries
         }
-        String rg_id_field = sub(sub(pair.right, ".*ID:", "ID:"), "\t.*", "") 
+        String rg_id_field = sub(sub(tuple.right, ".*ID:", "ID:"), "\t.*", "") 
         String rg_id = sub(rg_id_field, "ID:", "")
         call samtools.addreplacerg as single_end { input:
```
adthrasher, line 136: Rename `single_end` to something else

workflows/chipseq/chipseq-standard.wdl
```diff
@@ -133,8 +143,19 @@ workflow chipseq_standard {
...
-    call picard.validate_bam { input: bam = markdup.mkdupbam }
+    call picard.validate_bam { input:
+        bam = markdup.mkdupbam,
+        ignore_list = [
+            "MISSING_PLATFORM_VALUE",
+            "INVALID_PLATFORM_VALUE",
+            "INVALID_MAPPING_QUALITY",
+            "MATES_ARE_SAME_END",
+            "MISMATCH_FLAG_MATE_NEG_STRAND",
+            "MISMATCH_MATE_ALIGNMENT_START"
```
a-frantz, line 156: The first two seem "ok" to ignore. Less comfortable ignoring the last four. Are all these still being produced by seaseq?

workflows/chipseq/chipseq-standard.wdl
a-frantz: poke

PR #112, opened 2023-10-02
feat: librarian

a-frantz:
The `librarian` tool is still under development (see https://github.com/DesmondWillowbrook/Librarian/issues/17), but there are passenger-changes I added to this branch I want to see in `main`. So I'm voting for a preemptive merge, and we can update `librarian` later.

The passenger change in question: changing how analyses are toggled on/off in QC. By this, I mean we now present a "base set of analyses" with a number of "extra analyses" that can be **toggled on**, but are **off by default**. This is more-or-less how it currently works, though we had a variety of "toggling methods" and it was awkward IMO.

In practice, the above means dropping the `"DNA"/"RNA"` `String molecule` parameter and replacing it with `Boolean rna = false`. And some tweaking of phrasing in the docs.

### DIFF
--- a/tools/librarian.wdl
+++ b/tools/librarian.wdl
@@ -23,7 +23,7 @@ task librarian {
     )
 
     command <<<
-        /app/librarian --local -o ~{prefix} ~{read_one_fastq} ~{read_two_fastq}
+        RUST_BACKTRACE=1 /app/librarian --local -o ~{prefix} ~{read_one_fastq} ~{read_two_fastq}
     >>>
 
     # output {
--- a/tools/librarian.wdl
+++ b/tools/librarian.wdl
@@ -30,6 +30,7 @@ task librarian {
 
     output {
         File report = "~{prefix}.tar.gz"
+        File raw_data = "~{prefix}/librarian_heatmap.txt"
     }
 
     runtime {
--- a/tools/librarian.wdl
+++ b/tools/librarian.wdl
@@ -23,7 +23,7 @@ task librarian {
     )
 
     command <<<
-        RUST_BACKTRACE=1 /app/librarian --local -o ~{prefix} ~{read_one_fastq} ~{read_two_fastq}
+        RUST_LOG=trace /app/librarian --local -o ~{prefix} ~{read_one_fastq} ~{read_two_fastq}
     >>>
 
     # output {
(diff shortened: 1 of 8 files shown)


tools/librarian.wdl
```diff
@@ -0,0 +1,55 @@
...
+    Float read1_size = size(read_one_fastq, "GiB")
+    Int disk_size_gb = (
+        ceil(read1_size) + 10 + modify_disk_size_gb
+    )
+
+    command <<<
+        set -euo pipefail
+
+        mkdir ~{prefix}
+        RUST_LOG=trace /app/librarian --local --raw -o ~{prefix} ~{read_one_fastq}
```
adthrasher, line 39: Do we need `RUST_LOG=trace` in production?

tools/librarian.wdl
```diff
@@ -0,0 +1,55 @@
...
+    >>>
+
+    output {
+        File report = "~{prefix}.tar.gz"
+        File raw_data = "~{prefix}/librarian_heatmap.txt"
+    }
+
+    runtime {
+        memory: "4 GB"
+        disk: "~{disk_size_gb} GB"
```
adthrasher, line 51:
```suggestion
        disks: "~{disk_size_gb} GB"
```

workflows/qc/quality-check-standard.wdl
```diff
@@ -402,21 +409,21 @@ task parse_input {
...
+            check: "Dummy output to indicate success and to enable call-caching",  # TODO: is this still needed with labels?
             labels: "An array of labels to use on the result coverage files associated with each coverage BED"
         }
     }
 
     parameter_meta {
         coverage_labels: "An array of equal length to `coverage_beds_len` which determines the prefix label applied to coverage output files. If an empty array is supplied, defaults of `regions1`, `regions2`, etc. will be used."
-        input_molecule: "Must be `DNA` or `RNA`"
-        gtf_provided: "Was a GTF supplied by the user? Must be `true` if `input_molecule = RNA`."
+        rna: "Is the input molecule RNA?"
```
adthrasher, line 419: What about `input` -> `sequenced`? I think that's more accurate. It is input to this workflow though, but the molecule is what was sequenced.

adthrasher (changes requested):

tools/librarian.wdl
a-frantz: Good catch.
(For the record, the librarian team has ignored my suggestion to put the binary in the image's PATH. Which is why it's accessed `/app/librarian`)

PR #114, opened 2023-10-10
docs: More docs

a-frantz:
All our task+workflow files are dang near perfectly SPEC compliant now! Except the CellRanger/SCRNA stuff which we plan to move to it's own internal repo. The `style-guide` does specify that meta descriptions should be in the Active Voice, which has not been conformed to universally. (EDIT: that has been taken care of!) I think that's it though.

There were some small code tweaks that got included along the way, mostly renaming vars or shuffling the order of things to be compliant.

As usual, I added a fair number of `TODO`s along the way, for problems I saw but decided were OOS for this PR (or I was too lazy to deal with in the moment and decided they could wait).


tools/md5sum.wdl
```diff
@@ -1,10 +1,11 @@
...
-## This WDL file wraps the md5sum tool from the [GNU core utilities](https://github.com/coreutils/coreutils).
-## md5sum is a utility for generating and verifying MD5 hashes.
-
+## [Homepage](https://github.com/coreutils/coreutils)
+#
+# SPDX-License-Identifier: MIT
+# Copyright St. Jude Children's Research Hospital
 version 1.1
 
+# TODO why isn't this in util?
```
adthrasher, line 7: I actually prefer `md5sum` as a separate file. I think if a task is a straightforward invocation of a tool, then it should go in a per-tool WDL file. Otherwise `util` will become quite unwieldy.

adthrasher: In my mind, `util` is a place for utility tasks that are more complex than a single command invocation. I don't know that we've exactly followed that and there is some room for flexibility, but that's how I tend to think about it.

tools/star.wdl
```diff
@@ -120,13 +122,26 @@ task build_star_db {
...
+        outputs: {
+            star_log: "Summary mapping statistics after mapping job is complete. The statistics are calculated for each read (single- or paired-end) and then summed or averaged over all reads. Note that STAR counts a paired-end read as one read. Most of the information is collected about the UNIQUE mappers. Each splicing is counted in the numbers of splices, which would correspond to summing the counts in SJ.out.tab. The mismatch/indel error rates are calculated on a per base basis, i.e. as total number of mismatches/indels in all unique mappers divided by the total number of mapped bases."
+            star_bam: "STAR aligned BAM"
+            star_junctions: {
+                description:"File contains high confidence collapsed splice junctions in tab-delimited format. Note that STAR defines the junction start/end as intronic bases, while many other software define them as exonic bases. See `external_help` for file specification."
+                external_help: "https://github.com/alexdobin/STAR/blob/2.7.10a/doc/STARmanual.pdf"  # keep this up to date with container updates
```
adthrasher, line 131: Is this going to be difficult to maintain? I wonder if for STAR `external_help` should just be specified once at the top level, rather than per-output.

tools/star.wdl
```diff
@@ -712,6 +727,7 @@ task alignment {
     }
 }
 
+# TODO does this need documentation?
```
adthrasher, line 730: Probably just a short comment and a reference to the STAR docs.

tools/util.wdl
```diff
@@ -83,20 +96,35 @@ task get_read_groups {
     >>>
 
     output {
-        File read_groups_file = "read_groups.txt"
+        File read_groups_file = "read_groups.txt"  # TODO change output to String type (or Array[String])
```
adthrasher, line 99: Regarding the `TODO`, I would probably take a similar approach to other tools. We can generate a `File` output and a `Array[String]` output.

tools/util.wdl
```diff
@@ -301,12 +279,30 @@ task compression_integrity {
...
         memory: "~{memory_gb} GB"
         disk: "~{disk_size_gb} GB"
-        docker: 'quay.io/biocontainers/samtools:1.16.1--h6899075_1'
+        docker: 'quay.io/biocontainers/samtools:1.17--h00cdaf9_0'
         maxRetries: max_retries
     }
 }
 
 task add_to_bam_header {
+    # Should this be in samtools.wdl instead of util.wdl?
```
adthrasher, line 288: I suppose it could go there. However, it isn't a straightforward invocation of a `samtools` command. It serves a more specialized purpose. So to me, that suggests `util` is the place for it.

tools/util.wdl
```diff
@@ -301,12 +279,30 @@ task compression_integrity {
...
+    meta {
+        description: "Adds another line of text to the bottom of a BAM header"
+        outputs: {
+            updated_header: "The new header after modifications"  # TODO is this a lie? It would be missing a PG line from `reheader`. Why do we need this output?
+            header_lines: "Each line of the header is an entry in this array"  # Same question about needing this output
```
adthrasher, line 293: Both are probably artifacts of development. It was easier to check the input file to the `samtools reheader` command when implementing. It would be missing a `@PG` record from the reheader, but I would consider that an artifact of the processing, rather than a meaningful component of the header update.

workflows/chipseq/chipseq-standard.wdl
```diff
@@ -1,32 +1,5 @@
...
-## substantial portions of the Software.
-##
-## THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
-## BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
-## NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
-## DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-## OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
-
+# SPDX-License-Identifier: MIT
+# Copyright St. Jude Children's Research Hospital
```
adthrasher, line 2: I'm not sure about the shortened header for workflows. I'm not sure our earlier headers were good, but I think a brief overview of the workflow and purpose is a reasonable inclusion.

adthrasher (approved):

tools/md5sum.wdl
a-frantz: IMO, as a `coreutil`, `md5sum` belongs in `util`. Or maybe `md5sum.wdl` could be renamed `coreutils.wdl`? Group it with our `wget` implementation? Is there anything else that would go in there?
However, I'm not going to fight over it, because I think `util` is _already_ unwieldy enough. So preventing another addition to that file is ok by me, even if I don't really agree with the reasoning for keeping it separate.

I would like to explore ways to break up some of our larger files, `util` being among them. I often find it difficult to find the specific task I'm looking for in our long files. Just something I'd like on the back burner, I actually think I'm going to open an Issue for it so we can discuss. I won't list out my full rationale here, but I think there should be at least a soft-cap, if not a hard-cap, on the number of tasks in a WDL file. Don't have an exact number in mind though.

We can continue this conversation when I log that Issue. For now, let's remove this TODO and leave `md5sum.wdl` as is. Unless you like the `coreutils.wdl` rename?

tools/star.wdl
a-frantz: I think that's probably a good idea. If we do that, I think it would be appropriate to change the last sentences here to something along the lines of "See `meta.external_help` for file specification."
Question though, is `meta.external_help` the correct way to refer to that section (from here)? I can't think of a good alternative.

tools/util.wdl
a-frantz: I guess it's trivial to output the file as well, but what's the use case? I'd think we'd always prefer to work with the String directly, rather than need calls to `read_file()`. What scenario do you envision where having a file output is useful?

tools/util.wdl
a-frantz: A little unclear: it sounds like it would be safe to remove both these outputs?

workflows/chipseq/chipseq-standard.wdl
a-frantz: I'm open to the possibility. My counter-argument would be that the "brief overview of the workflow and purpose" would better belong in `meta.description`. I kept those brief, probably too brief. I'd rather rewrite those to be more verbose than go back to the headers.

What do you think of that?

tools/util.wdl
a-frantz: By that argument, we should move `subsample` out of `samtools.wdl` and into `util`. I disagree.

I would include `samtools` in my list of "files that need to be shortened/split up", so between having this task in `util` or `samtools`, I feel bothered either way lol. For simplicity (or you could probably say laziness), let's leave it as is. We can come back to this if my movement to shorten our files gets traction. This is a candidate for moving _somewhere_, I'm just not sure where yet.

tools/md5sum.wdl
adthrasher: I'm fine with deferring a decision for now.

tools/star.wdl
adthrasher: I think that's about the best I can come up with. If WDLdoc can handle that specially and just link to it, that would make it easier when looking at the rendered docs.

tools/util.wdl
adthrasher: I think the only purpose for the file would be storing the value as an output of the pipeline. I suppose it is easy enough for a use to dump the `@RG`s or look in the output `json`.

For the output, I'm fine with having a single `Array` object and in the STAR case, it just outputs a single entry. I think that's the easiest from a usability perspective.

tools/star.wdl
a-frantz: That would be a neat feature. Will try and remember that for when I get to the WDLdoc update.

tools/util.wdl
adthrasher: Yes, I think we're fine to remove these outputs.

workflows/chipseq/chipseq-standard.wdl
adthrasher: I'm fine with that as well. I think many people are conditioned to look at the start of a file for some information, but since WDL provides a mechanism for documentation elsewhere, I don't think we need to repeat it.

adthrasher (approved):

tools/util.wdl
a-frantz: IMO I'd rather have this streamlined by removing the `File` output than supporting a potential edge-case.
1) for this task specifically, what pipeline would want read groups as an output?
2) as a general rule for our `String` tasks, I would discourage ever outputting a generic file like `read_groups.txt`. I think all our outputs should follow `prefix` conventions, and I'd rather not complicate this task by needing a `prefix` argument.

> I'm fine with having a single Array object and in the STAR case, it just outputs a single entry. I think that's the easiest from a usability perspective.

Fair enough. The "usability perspective" makes a strong case. Optional outputs are annoying. I'll make a quick commit addressing this.

a-frantz: > 2\. as a general rule for our `String` tasks, I would discourage ever outputting a generic file like `read_groups.txt`. I think all our outputs should follow `prefix` conventions, and I'd rather not complicate this task by needing a `prefix` argument.

I didn't explain why I think this: When running batches, it's nice being able to aggregate all output files into the same directory. `prefix` conventions allow that to be a safe operation (unless you for some reason specify the same prefix in the same batch, which would be your fault IMO). Having any "generically named" files like `read_groups.txt` would cause clobbering, complicating the output process.

adthrasher (approved):

PR #118, opened 2023-11-03
Iterative samtools merge

adthrasher:
For a BAM with many read groups, the current workflow fails when passing the read groups to `samtools merge`. This is due to limitations on the command line argument length. This new workflow takes the input BAMs and if greater than the supplied limit, iteratively merges the BAMs.


adthrasher:
> If the command for `samtools merge` is exceeding `ARG_MAX`/`MAX_ARG_STRLEN` limits, an alternative approach would be to write the paths to a file and use the `-b` flag[1](#user-content-fn-b-39480e6937be907604197e46cb955f75) as the input.
> 
> ## Footnotes
> 1. `-b FILE    List of input BAM filenames, one per line [null]` [↩](#user-content-fnref-b-39480e6937be907604197e46cb955f75)

Thanks @zaeleus for the suggestion. I realize that my description above is imprecise. The error is actually upstream from the `samtools merge` invocation. When running `miniwdl` with `singularity` as the backend, `miniwdl` bind mounts each individual input file. So this is where the command line exceeds the limit.

a-frantz (approved):
I'm fine with neither of these files having complete meta info, as both are noted as being experimental.
Otherwise it looks good!

a-frantz (approved):

PR #119, opened 2023-11-07
chore: move from pr #111

adthrasher:
Samtools changes split off from PR #111

### DIFF
--- a/tools/samtools.wdl
+++ b/tools/samtools.wdl
@@ -540,7 +540,7 @@ task collate {
 
 task bam_to_fastq {
     meta {
-        description: "Runs `samtools fastq` on the input BAM file. Splits the BAM into FASTQ files. Assumes either a name sorted or collated BAM. For splitting a position sorted BAM see `collate_to_fastq`."
+        description: "Runs `samtools fastq` on the input BAM file. Converts the BAM into FASTQ files. If paired-end = false, then all reads in the BAM will be output to a single FASTQ file. Use filtering arguments to remove any unwanted reads. Assumes either a name sorted or collated BAM. For splitting a position sorted BAM see `collate_to_fastq`."
         outputs: {
             read_one_fastq_gz: "Gzipped FASTQ file with 1st reads in pair"
             read_two_fastq_gz: "Gzipped FASTQ file with 2nd reads in pair"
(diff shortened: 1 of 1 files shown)


tools/samtools.wdl
```diff
@@ -776,3 +791,68 @@ task collate_to_fastq {
...
+    }
+
+    runtime {
+        cpu: ncpu
+        memory: "~{memory_gb} GB"
+        disk: "~{disk_size_gb} GB"
+        container: 'quay.io/biocontainers/samtools:1.17--h00cdaf9_0'
+        maxRetries: max_retries
+    }
+}
```
a-frantz, line 858: style-guide requires a newline at EOF

a-frantz (changes requested):
For some reason GH isn't letting me leave comments on lines outside the scope of changes, so I'll use links I guess.

[This comment](https://github.com/stjudecloud/workflows/pull/119/files#diff-0da90800755d389152bdfeff3692a26dacb373da7f46dad9f87ebda52b808e07R650-R651) is no longer true.
We also want to come up with a better name for `File? interleaved_reads_fastq_gz = "~{prefix}.fastq.gz"`. This filename (and therefore output) is now used for both Paired-End interleaved FASTQs and all Single-End FASTQs. That's a bit tricky to name...
I wonder if there's anything stopping us from pointing at the same file with 2 different output variables. I bet that would work, but haven't tested it. Then we could change the output block to:
```
File? interleaved_reads_fastq_gz = "~{prefix}.fastq.gz"
File? single_end_reads_fastq_gz = "~{prefix}.fastq.gz"
```
Or is that confusing? My rationale is that it let's us re-use the file-name, but calling workflows will get to use a var name that's appropriate to them. What do you think? Any better ideas?

Lastly, this is the cost of our code duplication. Can you port these `bam_to_fastq` changes over to `collate_to_fastq` so they are in sync?

adthrasher:
> For some reason GH isn't letting me leave comments on lines outside the scope of changes, so I'll use links I guess.
> 
> [This comment](https://github.com/stjudecloud/workflows/pull/119/files#diff-0da90800755d389152bdfeff3692a26dacb373da7f46dad9f87ebda52b808e07R650-R651) is no longer true. We also want to come up with a better name for `File? interleaved_reads_fastq_gz = "~{prefix}.fastq.gz"`. This filename (and therefore output) is now used for both Paired-End interleaved FASTQs and all Single-End FASTQs. That's a bit tricky to name... I wonder if there's anything stopping us from pointing at the same file with 2 different output variables. I bet that would work, but haven't tested it. Then we could change the output block to:
> 
> ```
> File? interleaved_reads_fastq_gz = "~{prefix}.fastq.gz"
> File? single_end_reads_fastq_gz = "~{prefix}.fastq.gz"
> ```
> 
> Or is that confusing? My rationale is that it let's us re-use the file-name, but calling workflows will get to use a var name that's appropriate to them. What do you think? Any better ideas?
> 
> Lastly, this is the cost of our code duplication. Can you port these `bam_to_fastq` changes over to `collate_to_fastq` so they are in sync?

I've removed the two comments. I added the additional output specification. `miniwdl check` didn't report any issues, but I didn't run real data through.

I copied the changes to `collate_to_fastq` as well.

a-frantz (approved):

PR #120, opened 2023-11-10
fix: calc_gene_length algorithm. And better docs for gene length and TPM

a-frantz:
See title


a-frantz:
@adthrasher This is ready for review.
But I wonder if while we're at it, we should make the `gene_name` key customizable. `gene_name` isn't even guaranteed by the SPEC, but `gene_id` is. I think offering a bool toggle between `gene_name` and `gene_id` covers the vast majority of use cases.

We could also make the key arbitrary. Just glancing at the gencode defined attributes, I can see others serving a purpose: `havana_gene` for sure; not so sure about the following: `transcript_id`, `transcript_name`, `havana_transcript`, `protein_id`

adthrasher:
> @adthrasher This is ready for review. But I wonder if while we're at it, we should make the `gene_name` key customizable. `gene_name` isn't even guaranteed by the SPEC, but `gene_id` is. I think offering a bool toggle between `gene_name` and `gene_id` covers the vast majority of use cases.
> 
> We could also make the key arbitrary. Just glancing at the gencode defined attributes, I can see others serving a purpose: `havana_gene` for sure; not so sure about the following: `transcript_id`, `transcript_name`, `havana_transcript`, `protein_id`

I would suggest we make it arbitrary. I think that's how we implement it elsewhere (e.g. `htseq`). Maybe with a `choices` meta.

adthrasher (approved):
All comments were discussed via slack.

adthrasher (approved):

PR #122, opened 2023-11-21
Style updates

a-frantz:
Makes use of `runtime` overriding added by the WDLv1.1 SPEC. This means moving our statically allocated resources out of `input{}` and directly into `runtime`. 
Note that tasks which must be aware of their runtime attributes (mainly memory and nCPU) outside of the `runtime` block have kept their static allocations in the `input` block. We may revisit this in the future, especially with some of the changes slated for WDLv1.2.

This PR also adds the `common: true` tag to many parameter metas.

It also adds `allowNestedInputs: true` to all workflows' meta section.


tools/mosdepth.wdl
```diff
@@ -21,10 +21,11 @@ task coverage {
         coverage_bed: "BED file to pass to the `-b` flag of `mosdepth`. This will restrict coverage analysis to regions defined by the BED file."
```
adthrasher, line 21: Do we think this would be `common`? I think I would lean towards "yes".

adthrasher: Or are we assuming in the future it would surface due to being 'optional'?

tools/samtools.wdl
```diff
@@ -814,19 +887,25 @@ task fixmate {
         extension: {
             description: "File format extension to use for output file.",
             choices: [
-                ".sam",
+                ".sam",  # TODO are we sure we want to encourage the use of .sam?
```
adthrasher, line 890: I think it's more a question of fully supporting the underlying tool options. It is also possible that in the future, you could stream output between tasks, rather than passing files around (maybe?). SAM may make sense in that application.

workflows/general/samtools-merge.wdl
```diff
@@ -12,20 +12,18 @@ workflow samtools_merge {
...
         prefix: "Prefix for output BAM."
         use_all_cores: "Use all cores? Recommended for cloud environments. Not recommended for cluster environments."
-
     }
     input {
         Array[File] bams
         Int max_length = 100
-        Int? max_retries
-        String prefix = basename(bams[0], ".bam")
+        String prefix = basename(bams[0], ".bam")  # TODO is this a sane default? Or should be required?
```
adthrasher, line 26: That's a good question. It likely should just be a required parameter. But it's fine to leave the `TODO` for now.

adthrasher (approved):
That was a very large PR, in terms of file count.

tools/mosdepth.wdl
a-frantz: My assumption was that we would automatically surface all `optional` parameters. I kept my eye out for them while going through all these files, and I think it always makes sense to give them special treatment.
There aren't many places we use optional parameters. We typically default to an empty array or other such "null-esque" value.

tools/samtools.wdl
a-frantz: Fair point on the future application. I hadn't thought of that, but skipping [de]compression makes sense in that case.

> I think it's more a question of fully supporting the underlying tool options

I disagree that this is necessary. There's precedence for us saying "we're not exposing XYZ parameter because it's silly/bad/etc." We've done it with STAR, Qualimap RNAseq, probably others I'm blanking on at the moment.

As I see it, there are 2 philosophies conflicting here:
1) we should expose all parameters we can. **Trust the end user to not break things.**
2) **we should provide a curated set of parameters to users** (because we can't really trust the end user to not break things)
    * In most cases that is "all parameters" and this looks like philosophy (1)
    * in certain cases we should hide "bad" options from users. Defining "bad" is tough, but I would call "passing around uncompressed SAM data and storing it on disk at least once if not more times" an objectively "bad" idea.
    * As I/O streaming is not yet a feature of WDL, I do not see _any_ case where SAM output is appropriate.

I vote we remove `".sam"` from `choices` and call it a day. A user could still pass in `.sam` and get the "desired" behavior, but we don't broadcast it as a possibility. If streaming is added in the future, we revisit the best way to expose this.

adthrasher: I'm not sure I agree on SAM being bad in this context. The default is BAM. If a user wants SAM, it's up to them to understand the storage issue. I don't have a concrete idea for why someone would want to do that, but they might. I suppose removing it from `choices` would be fine. It doesn't remove the functionality, but does make it a little more hidden.

Issue #88, opened 2023-04-13
investigate appropriate `disk_size`s


a-frantz:

a-frantz:
@adthrasher can you think of a decent way to test how much disk space each task uses? The difficult part is that many of our tasks clean up after themselves (a good thing), but that means we need to snapshot disk usage during runtime. A more complicated task than just running `du` after completion.

We could just run everything on the cloud with the current allocations and see what fails. That sounds the most straightforward, and will be "real-world" in that the host OS will be using some of the disk unlike on the cluster.

Does HPCF have any tools that would help here? I've always been under the impression their disk space tools have a hefty lag, so maybe not.

PR #123, opened 2023-12-07
RNA-Seq germline variant calling pipeline

adthrasher:
Add a variant calling pipeline for RNA-Seq. Implements the GATK RNA-Seq short variant discovery best practices workflow (https://gatk.broadinstitute.org/hc/en-us/articles/360035531192-RNAseq-short-variant-discovery-SNPs-Indels-). 

Adds an implementation of `sambamba` in WDL.


tools/gatk4.wdl
```diff
@@ -0,0 +1,333 @@
...
+    >>>
+
+    output {
+        File split_bam = "~{prefix}.bam"
+        File split_bam_index = "~{prefix}.bam.bai"
+        File split_bam_md5 = "~{prefix}.bam.md5"
+    }
+
+    runtime {
+        cpu: 8
```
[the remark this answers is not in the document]
a-frantz: 3. please use an `ncpu` param, just for consistency with all other tasks. Hypothetical: a user needs to single-thread all our tasks. This one task would have `<>.runtime.cpu: 1` syntax while everything else is `<>.ncpu: 1`

tools/picard.wdl
```diff
@@ -865,3 +865,180 @@ task bam_to_fastq {
...
+        modify_disk_size_gb: "Add to or subtract from dynamic disk space allocation. Default disk size is determined by the size of the inputs. Specified in GB."
+    }
+
+    input {
+        File fasta
+        String outfile_name = basename(fasta, ".fa") + ".dict"
+        String? assembly_name
+        String? fasta_url
+        String? species
+        Int memory_gb = 16
```
a-frantz, line 1013: Can just put `memory: "16 GB"` directly in the runtime block instead of this.

tools/picard.wdl
```diff
@@ -865,3 +865,180 @@ task bam_to_fastq {
...
+
+    parameter_meta {
+        fasta: "Input FASTA format file from which to create dictionary"
+        outfile_name: "Name for the CreateSequenceDictionary dictionary file"
+        memory_gb: "RAM to allocate for task, specified in GB"
+        modify_disk_size_gb: "Add to or subtract from dynamic disk space allocation. Default disk size is determined by the size of the inputs. Specified in GB."
+    }
+
+    input {
+        File fasta
```
a-frantz, line 1008: I figure that in conjunction with the optional params this should also be optional?

a-frantz (commented):
I haven't yet looked at the new sambamba file (or anything "below" that in the GH interface). Will get to it today, but not immediately. Wanted to kick off the feedback I did write.

tools/gatk4.wdl
```diff
@@ -0,0 +1,333 @@
...
+        File bam_index
+        File interval_list
+        File fasta
+        File fasta_index
+        File dict
+        String prefix = basename(bam, ".bam") + ".split"
+        Int modify_disk_size_gb = 0
+    }
+
+    Int disk_size_gb = ceil(size(bam, "GB") + 1) * 5 + ceil(size(fasta, "GB")) + modify_disk_size_gb
```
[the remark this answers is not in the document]
adthrasher: I'm not certain. I took that directly from GATK.

tools/gatk4.wdl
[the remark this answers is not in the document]
adthrasher: GATK SplitNCigarReads doesn't offer any explicit way to control parallelism. However, when I look at the LSF results, it spawns 8 processes. The number of threads varies. I will also note that I haven't done any profiling to see how fully it utilizes the cores.

```
     Max Processes :                              8
     Max Threads :                                65
```

tools/gatk4.wdl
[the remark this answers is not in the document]
a-frantz: Hm. That's fishy to me. But considering our exec. env. ignore `disks` I guess it doesn't matter. I'm fine pushing this off till #88 gets addressed (although who knows when that will creep up the priority list enough to actually see progress).

tools/gatk4.wdl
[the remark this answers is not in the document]
a-frantz: Well I don't like that, but I guess we're stuck with it.

The repo-wide approach is generally to be conservative on `cpu` allocation. Unless this task takes an extreme amount of time, I would roll it back to `4`(?)

tools/gatk4.wdl
```diff
@@ -0,0 +1,333 @@
...
+            --spark-master local[4] \
+            -I ~{bam} \
+            ~{if use_original_quality_scores then "--use-original-qualities" else "" } \
+            -O ~{prefix}.bam \
+            --bqsr-recal-file ~{recalibration_report}
+       # GATK is unreasonable and uses the plain ".bai" suffix.
+       #mv ~{prefix}.bai ~{prefix}.bam.bai
+    >>>
+
+    output {
```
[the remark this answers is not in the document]
adthrasher: No and I see no option to create it...Who needs consistency

tools/gatk4.wdl
```diff
@@ -0,0 +1,333 @@
...
+    input {
+        File bam
+        File bam_index
+        File recalibration_report
+        String prefix = basename(bam, ".bam") + ".bqsr"
+        Int modify_disk_size_gb = 0
+        Boolean use_original_quality_scores = true
+    }
+
+    Int disk_size_gb = ceil(size(bam, "GB") * 4) + 30 + modify_disk_size_gb
```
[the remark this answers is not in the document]
adthrasher: I took this directly from the GATK implementation, but I agree. That seems extreme, even if the original quality scores are added to the BAM.

tools/picard.wdl
adthrasher: The `fasta` file is required. The URL is optional because it is only added to the `UR` field of the sequence dictionary.

tools/picard.wdl
```diff
@@ -865,3 +865,180 @@ task bam_to_fastq {
...
+    }
+
+    Float fasta_size = size(fasta, "GiB")
+    Int disk_size_gb = ceil(fasta_size * 2) + 10 + modify_disk_size_gb
+    Int java_heap_size = ceil(memory_gb * 0.9)
+
+    command <<<
+        set -euo pipefail
+
+        rc=0
```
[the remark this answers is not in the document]
adthrasher: No idea...

tools/picard.wdl
adthrasher: I use that value to compute `java_heap_size` which gets passed to the JVM. If WDL gets support for accessing runtime values in the `command` block eventually, we can remove this.

a-frantz: Ah yes. We've used this elsewhere too. Looks good then 👍

tools/gatk4.wdl
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    parameter_meta  {
+        vcf: "Input VCF format file to filter"
+        vcf_index: "VCF index file corresponding to the input VCF"
+        fasta: "Reference genome in FASTA format"
+        fasta_index: "Index for FASTA format genome"
+        dict: "Dictionary file for FASTA format genome"
+        prefix: "Prefix for the output filtered VCF. The extension `.vcf.gz` will be added."
+        filter_name: "Name of the filter to apply"
```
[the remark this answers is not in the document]
adthrasher: It's free text. The name gets inserted into the VCF `FILTER` field, but it is completely user-dependent.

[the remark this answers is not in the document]
a-frantz: Ahhh ok. That does make sense but it wasn't clear to me at first. So these are just the labels that go along with `filter_expression`? Just an idea, IDK if it's the right approach, but should these two params be a struct?

tools/gatk4.wdl
```diff
@@ -0,0 +1,333 @@
...
+# SPDX-License-Identifier: MIT
+# Copyright St. Jude Children's Research Hospital
+version 1.1
+
+task split_n_cigar_reads {
+    meta {
+        description: "Splits reads that contain Ns in their CIGAR strings into multiple reads."
+        external_help: "https://gatk.broadinstitute.org/hc/en-us/articles/360036858811-SplitNCigarReads"
+        outputs: {
+            split_bam: "BAM file with reads split at N CIGAR elements and updated CIGAR strings."
```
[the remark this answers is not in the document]
adthrasher: That's a good question. I could go with the simpler `output_bam`, but we generally disallow that and `bam` seems a little imprecise. Something like `split_n_bam` would work, but is that clearer?

tools/gatk4.wdl
```diff
@@ -0,0 +1,333 @@
...
+        modify_memory_gb: "Add to or subtract from dynamic memory allocation. Default memory is determined by the size of the inputs. Specified in GB."
+    }
+
+    input {
+        File bam
+        File bam_index
+        File interval_list
+        File fasta
+        File fasta_index
+        File dict
```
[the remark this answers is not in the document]
adthrasher: `miniwdl` definitely links everything into a single directory.
```
$ ls -l 20240118_104404_rnaseq_variant_calling/call-split_n_cigar_reads/work/_miniwdl_inputs/0/
total 0
-rwxrwx--- 1 athrashe Domain Users 0 Jan 18 14:31 GRCh38_no_alt.dict
-rwxrwx--- 1 athrashe Domain Users 0 Jan 18 14:31 GRCh38_no_alt.fa
-rwxrwx--- 1 athrashe Domain Users 0 Jan 18 14:31 GRCh38_no_alt.fa.fai
-rwxrwx--- 1 athrashe Domain Users 0 Jan 18 14:31 HG005_GM24631_replicate1_totalRNAseq_ILLUMINA_UNC_20220518_ACTCGTGT-ATTGGAAC.main_from_bam.Aligned.out.sorted.MarkDuplicates.bam
-rwxrwx--- 1 athrashe Domain Users 0 Jan 18 14:31 HG005_GM24631_replicate1_totalRNAseq_ILLUMINA_UNC_20220518_ACTCGTGT-ATTGGAAC.main_from_bam.Aligned.out.sorted.MarkDuplicates.bam.bai
-rwxrwx--- 1 athrashe Domain Users 0 Jan 18 14:31 wgs_calling_regions.hg38.interval_list
```

That said, my reading of the spec (https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#task-input-localization), particularly this part: 
```
Two input files that originated in the same storage directory must also be localized into the same directory for task execution (see the special case handling for Versioning Filesystems below).
```

Suggests that localizing index files manually should not be necessary.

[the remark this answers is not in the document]
a-frantz: Ok. I'm fine discontinuing the localization step.

tools/gatk4.wdl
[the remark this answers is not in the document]
a-frantz: I'm ok with being verbose. My vote is `n_reads_split_bam`. Or `split_n_reads_bam`? IDK which is clearer.

workflows/rnaseq/rnaseq-variant-calling.wdl
```diff
@@ -0,0 +1,132 @@
...
+import "../../tools/picard.wdl"
+import "../../tools/gatk4.wdl" as gatk
+
+workflow rnaseq_variant_calling {
+    meta {
+        description: "Call variants from RNA-Seq data. Produces a VCF file of variants. Based on GATK RNA-Seq short variant calling best practices pipeline."
+        outputs: {
+            recalibrated_bam: "BAM that has undergone recalibration of base quality scores"
+            recalibrated_bam_index: "Index file for recalibrated BAM file"
+            merged_vcf: "VCF of raw RNA-Seq variant calls"
```
a-frantz, line 14: Does the user care it's been merged? I wouldn't care. I supplied one bam, I expect one VCF. I would rename this, but IDK what. Plain `vcf` is a little terse, and we've gone on record saying output names should be "descriptive". `unfiltered_vcf`? Does that make it sound like an intermediate file? `raw_vcf`?

workflows/rnaseq/rnaseq-variant-calling.wdl
```diff
@@ -0,0 +1,132 @@
...
+            merged_vcf_index: "Index for merged VCF file"
+            variant_filtered_vcf: "VCF file after variant filters have been applied"
+            variant_filtered_vcf_index: "Index for filtered variant VCF file"
+        }
+    }
+
+    parameter_meta {
```
a-frantz, line 21: snake_case all these please.

workflows/rnaseq/rnaseq-variant-calling.wdl
```diff
@@ -0,0 +1,132 @@
...
+        bam: "BAM file of aligned RNA-Seq reads"
+        bam_index: "Index file for BAM file"
+        fasta: "Reference FASTA file"
+        fasta_index: "Index file for reference FASTA file"
+        dict: "Sequence dictionary for reference FASTA file"
+        callingIntervalList: "Interval list of regions from which to call variants. Used for parallelization."
```
a-frantz, line 27: Good call putting `Used for parallelization`. I was looking at this a little quizzically until I saw that lol.

workflows/rnaseq/rnaseq-variant-calling.wdl
```diff
@@ -0,0 +1,132 @@
...
+        knownVcfs: "Array of known indels VCF files"
+        knownVcfIndexes: "Array of index files for known indels VCF files"
+        dbSNP_vcf: "dbSNP VCF file"
+        dbSNP_vcf_index: "Index file for dbSNP VCF file"
+        scatterCount: "Number of intervals to scatter over"
```
a-frantz, line 32: Should we have guidance on setting this? `6` seems like a magic number to me. (i.e. I do not understand why it was picked)

workflows/rnaseq/rnaseq-variant-calling.wdl
```diff
@@ -0,0 +1,132 @@
...
+        output_vcf_name: "Name of output VCF file"
```
a-frantz, line 33: IMO this should WF should `prefix` and force all the suffixes/extensions. I think if I'm reading correctly, the user doesn't have a say in the output BAM name.

workflows/rnaseq/rnaseq-variant-calling.wdl
```diff
@@ -0,0 +1,132 @@
...
+        File callingIntervalList
+        Array[File] knownVcfs
+        Array[File] knownVcfIndexes
+        File dbSNP_vcf
+        File dbSNP_vcf_index
+        Int scatterCount = 6
+        String output_vcf_name = basename(bam, '.bam') + '.vcf.gz'
+    }
+    
+    call picard.mark_duplicates {
```
a-frantz, line 51: What if the BAM's already been dup marked? I guess since this is RNA specific that's unlikely. Do we need to handle that?

workflows/rnaseq/rnaseq-variant-calling.wdl
```diff
@@ -0,0 +1,132 @@
...
+        input: 
+            bam = bam,
+            create_bam = true
+    }
+
+    call gatk.split_n_cigar_reads {
+        input:
+            bam = select_first([mark_duplicates.duplicate_marked_bam, '']),
```
a-frantz, line 59: I think inconsistency on this is _ok_, so won't hold up the PR on this change. But I think the literal in a select_first should be `"undefined"`. That's clearer to me than an empty string. Also ok with `error` or something like that. I just like it to be more explicit.

Also we've switched to using double quotes everywhere. Please double check if I missed any single quotes elsewhere.

workflows/rnaseq/rnaseq-variant-calling.wdl
```diff
@@ -0,0 +1,132 @@
...
+            recalibration_report = base_recalibrator.recalibration_report
+    }
+
+    call picard.scatter_interval_list {
+        input:
+            interval_list = callingIntervalList,
+            scatter_count = scatterCount
+    }
+
+    scatter (i in range(scatter_interval_list.interval_count)) {
```
a-frantz, line 93:
```suggestion
    scatter (list in scatter_interval_list.out) {
```
I think that would be more straight-forward, right?

tools/picard.wdl
```diff
@@ -865,3 +865,182 @@ task bam_to_fastq {
...
+           dir=\$(dirname \$list)
+           name=\$(basename \$list)
+           mv \$list \${dir}/\${I}\${name}
+        done
+        echo \$I > interval_count.txt
+        CODE
+    >>>
+
+    output {
+        Array[File] out = glob("out/*/*.interval_list")
```
a-frantz, line 979:
```suggestion
        Array[File] interval_lists = glob("out/*/*.interval_list")
```
Or is that too close to `interval_list` input?

tools/gatk4.wdl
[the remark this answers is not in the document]
adthrasher: I reduced it to 3x BAM size. I'm sure it needs 2x since it stores the original and the updated. 3x gives some extra overhead.

tools/gatk4.wdl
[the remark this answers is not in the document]
adthrasher: I went with `split_n_reads_bam`.

tools/gatk4.wdl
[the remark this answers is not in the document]
adthrasher: My concern is in stricter cluster environments, if the job uses 8 cores, it may get killed if it only requests 4 cores.

tools/gatk4.wdl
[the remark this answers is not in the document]
adthrasher: I dropped it to 2x BAM + 30GB. That should be sufficient to store both copies of the BAM with some overhead.

tools/gatk4.wdl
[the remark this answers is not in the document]
adthrasher: We could add a `struct`, but that would be unlikely to be broadly useful. I've chosen filters and names that mirror the reference best practices workflow, so I would hope that these would not need to be commonly modified. That said, specifying the WDL parameters is not much different from using the command line tool, so most users are probably familiar. Not necessarily the ideal case though.

tools/samtools.wdl
```diff
@@ -1249,3 +1249,58 @@ task markdup {
...
+    }
+
+    input {
+        File fasta
+        Boolean use_all_cores = false
+        Int modify_disk_size_gb = 0
+    }
+
+    Float fasta_size = size(fasta, "GiB")
+    Int disk_size_gb = ceil(fasta_size * 1.2) + 10 + modify_disk_size_gb
```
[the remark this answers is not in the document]
adthrasher: Good point. I suspect the `+ 10` was doing a lot of work in this equation.

workflows/rnaseq/rnaseq-variant-calling.wdl
```diff
@@ -0,0 +1,132 @@
...
+version 1.1
+
+import "../../tools/picard.wdl"
+import "../../tools/gatk4.wdl" as gatk
+
+workflow rnaseq_variant_calling {
+    meta {
+        description: "Call variants from RNA-Seq data. Produces a VCF file of variants. Based on GATK RNA-Seq short variant calling best practices pipeline."
+        outputs: {
+            recalibrated_bam: "BAM that has undergone recalibration of base quality scores"
```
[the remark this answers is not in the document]
adthrasher: Maybe. I output it because there may be other GATK steps that someone would want to run later. For example, somatic variant calling or SV analysis.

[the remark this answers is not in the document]
a-frantz: I could see it going either way. So as you have it is fine (and a bit more straight-forward).

workflows/rnaseq/rnaseq-variant-calling.wdl
a-frantz: Another thought on this. Isn't this VCF strictly a subset of the information present in the filtered VCF? Is there any use for it? If a user doesn't care about the annotations/filters applied, they can always be ignored (is "filtered" the right word? Variants which fail get annotated, not removed, right?).

tools/picard.wdl
adthrasher: I went with `interval_lists_scatter`. It's still close to the task name, but better than `out`.

adthrasher: I went with `interval_lists_scatter`. It's still close to the task name, but better than `out`.

workflows/rnaseq/rnaseq-variant-calling.wdl
adthrasher: According to the tool documentation, it should be possible to remove the variants. However, I don't see a command line flag that actually enables that feature. Also, I definitely don't expose such a parameter. So you're correct, we could just output the filtered VCF. It's a little more cumbersome to ignore the filters in analysis, but it is possible.

a-frantz: > it should be possible to remove the variants. However, I don't see a command line flag that actually enables that feature.

🙄 

> we could just output the filtered VCF

My thinking is that it's simpler to only have one output VCF. No user is going to ask us "which should I use?" Worst case someone down the line complains (with a valid use case for the "unfiltered" VCF) and we add this line back 🤷

workflows/rnaseq/rnaseq-variant-calling.wdl
adthrasher: It is definitely a magic number. I copied it from the GATK workflow. I can't find any real guidance for setting it. In fact in many places, the Broad pipelines leave it unset and completely up to the user. [This](https://gatk.broadinstitute.org/hc/en-us/community/posts/360058099652-Scatter-Count-recommendations) is about the only thing I can find and it isn't overly helpful. The [documentation](https://gatk.broadinstitute.org/hc/en-us/articles/360037592651-IntervalListTools-Picard#--SCATTER_COUNT) is similarly vague.

adthrasher: I think it is also dependent on the value of `subdivision_mode`.

workflows/rnaseq/rnaseq-variant-calling.wdl
adthrasher: Agreed. I've made that change. Now `prefix` will be used to name the recalibrated BAM and the filtered VCF.

workflows/rnaseq/rnaseq-variant-calling.wdl
adthrasher: I've added a param to enable this.

workflows/rnaseq/rnaseq-variant-calling.wdl
a-frantz: Just bringing in 2 quotes from that forum answer:
> We usually use 10-20 for exomes. For our purposes this is a good compromise between wall clock time and the overhead of starting VMs and delocalizing. 

> If you're not using the cloud you could increase the scatter count, but there's no real need unless you're really in a hurry.

So the balance here is greater `scatter_count` leads to quicker turnaround (up to a point) but at higher costs for spinning up instances. That's fairly straightforward. Should we add a sentence saying something to the effect "This should typically be set to 5-20."?

tools/sambamba.wdl
```diff
@@ -0,0 +1,293 @@
...
+        File bam
+        String sort_order = "coordinate"
+        String prefix = basename(bam, ".bam") + ".sorted"
+        Int memory_gb = 25
+        Int modify_disk_size_gb = 0
+        Int ncpu = 2
+    }
+
+    Float bam_size = size(bam, "GiB")
+    Int disk_size_gb = ceil(bam_size * 4) + 10 + modify_disk_size_gb
```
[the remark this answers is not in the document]
adthrasher: I dropped it to `* 3`. It does store partial files on disk in order to do the final merge, so I would think, as you say, that 3x would be the worst case.

workflows/reference/gatk-reference.wdl
```diff
@@ -0,0 +1,108 @@
+# SPDX-License-Identifier: MIT
+# Copyright St. Jude Children's Research Hospital
+version 1.1
+
+import "../../tools/picard.wdl"
+import "../../tools/samtools.wdl"
+import "../../tools/util.wdl"
+
+workflow gatk_reference {
```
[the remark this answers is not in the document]
adthrasher: Everything but the dbSNP download finished in ~2m on the cluster. Total time is ~3.5m. The dbSNP file is uncompressed and is ~10GB.

tools/gatk4.wdl
```diff
@@ -0,0 +1,333 @@
...
+        Int modify_disk_size_gb = 0
+        Boolean use_original_quality_scores = true
+    }
+
+    Int disk_size_gb = ceil(size(bam, "GB") + 1) * 3 + ceil(size(fasta, "GB")) + modify_disk_size_gb
+
+    command <<<
+        gatk \
+            --java-options "-XX:GCTimeLimit=50 -XX:GCHeapFreeLimit=10 -XX:+PrintFlagsFinal \
+            -Xloggc:gc_log.log -Xms4000m" \
```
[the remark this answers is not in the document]
adthrasher: `GCHeapFreeLimit` is the amount of heap space that must be freed in a GC run (defaults to 2%). 
`GCTimeLimit` is the amount of time spent garbage collecting as a percentage of the total time (defaults to 98%). I suspect that these were introduced by the Broad to avoid cases where the process gets swamped by excessive GC runs that don't actually find much space to free. In those cases, it's probably better to throw an error and have the user resubmit with additional memory.

workflows/reference/gatk-reference.wdl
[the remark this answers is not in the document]
a-frantz: Ok good to know. I was considering asking for a `parse_input` task, but that would be overkill when the cost of a crashed workflow is so negligible.

workflows/reference/gatk-reference.wdl
```diff
@@ -0,0 +1,108 @@
...
+            fasta = fasta_download.downloaded_file
+    }
+
+    call util.download as dbsnp {
+        input:
+            url = dbSNP_vcf_url,
+            outfile_name = dbSNP_vcf_name
+    }
+
+    if (defined(dbSNP_vcf_index_url)) {
```
a-frantz, line 75:
```suggestion
    if (defined(dbSNP_vcf_index_url) && defined(dbSNP_vcf_index_name)) {
```
Currently, if a user skips passing in a name (but does pass in a URL), the file will be called `undefined`.

workflows/reference/gatk-reference.wdl
```diff
@@ -0,0 +1,108 @@
...
+        call util.download as dbsnp_index {
+            input:
+                url = select_first([dbSNP_vcf_index_url, "undefined"]),
+                outfile_name = select_first([dbSNP_vcf_index_name, "undefined"])
+        }
+    }
+
+    if (defined(interval_list_url)) {
```
a-frantz, line 83:
```suggestion
    if (defined(interval_list_url) && defined(interval_list_name)) {
```

workflows/rnaseq/rnaseq-variant-calling.wdl
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    call gatk.apply_bqsr { input:
+        bam = split_n_cigar_reads.split_n_reads_bam,
+        bam_index = split_n_cigar_reads.split_n_reads_bam_index,
+        recalibration_report = base_recalibrator.recalibration_report,
+        prefix = prefix
+    }
+
+    call picard.scatter_interval_list { input:
```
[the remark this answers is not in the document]
a-frantz: I agree 👍

tools/gatk4.wdl
```diff
@@ -0,0 +1,358 @@
...
+        File fasta_index
+        File dict
+        File dbSNP_vcf
+        File dbSNP_vcf_index
+        Array[File] known_indels_sites_VCFs
+        Array[File] known_indels_sites_indices
+        Int memory_gb = 25
+        Int modify_disk_size_gb = 0
+        Int ncpu = 4
+        Boolean use_original_quality_scores = true
```
a-frantz, line 112: If I recall prior discussion correctly, this makes the most sense as defaulting to `false` (and that's what the parameter meta says is the default), but the GATK WDL defaults it to `true`? I think we ignore the GATK WDL and set this as `false`.

tools/gatk4.wdl
```diff
@@ -0,0 +1,358 @@
...
+        Int modify_disk_size_gb = 0
+        Int ncpu = 8
+    }
+
+    Int disk_size_gb = ceil(size(bam, "GB") + 1) * 3 + ceil(size(fasta, "GB")) + modify_disk_size_gb
+
+    command <<<
+        set -euo pipefail
+
+        gatk \
```
a-frantz, line 47: Is there a reason this command doesn't have `--java-options` like the other `gatk` commands?

tools/gatk4.wdl
```diff
@@ -0,0 +1,358 @@
+## [Homepage](https://software.broadinstitute.org/gatk)
+#
+# SPDX-License-Identifier: MIT
+# Copyright St. Jude Children's Research Hospital
+version 1.1
```
a-frantz, line 5:
```suggestion
## [Homepage](https://software.broadinstitute.org/gatk)

version 1.1
```

tools/gatk4.wdl
```diff
@@ -0,0 +1,358 @@
...
+    runtime {
+        cpu: ncpu
+        memory: "~{memory_gb} GB"
+        disk: "~{disk_size_gb} GB"
+        container: "quay.io/biocontainers/gatk4:4.4.0.0--py36hdfd78af_0"
+        maxRetries: 1
+    }
+}
+
+task apply_bqsr {
```
a-frantz, line 144: Just an idea: does it make sense to combine this and the prior task into one? Is there any context where you would run `base_recalibrator` but **not** `apply_bqsr`?

tools/picard.wdl
```diff
@@ -866,3 +866,182 @@ task bam_to_fastq {
...
+        Array[File] vcfs
+        Array[File] vcfs_indexes
+        String output_vcf_name
+        Int modify_disk_size_gb = 0
+    }
+
+    Int disk_size_gb = ceil(size(vcfs, "GiB") * 2) + 10 + modify_disk_size_gb
+
+    command <<<
+        picard -Xms2000m \
```
a-frantz, line 897: It probably doesn't practically matter, but shouldn't this be linked to the `memory: "3 GB"` line?

Side note, why 3gb? I'm fairly sure that all our "low-mem" tasks use 4gb. 4 is a magic number in those cases, so 3 is fine here. We may just want to be consistent.

tools/picard.wdl
```diff
@@ -866,3 +866,182 @@ task bam_to_fastq {
...
+            SCATTER_COUNT=~{scatter_count} \
+            SUBDIVISION_MODE=~{subdivision_mode} \
+            UNIQUE=~{unique} \
+            SORT=~{sort} \
+            INPUT=~{interval_list} \
+            OUTPUT=out
+
+        bash <<CODE
+        I=0
+        for list in \$(ls out/*/*.interval_list)
```
a-frantz, line 968:
```suggestion
        for list in out/*/*.interval_list
```

tools/picard.wdl
```diff
@@ -866,3 +866,182 @@ task bam_to_fastq {
...
+
+        mkdir out
+        picard -Xms1g \
+            IntervalListTools \
+            SCATTER_COUNT=~{scatter_count} \
+            SUBDIVISION_MODE=~{subdivision_mode} \
+            UNIQUE=~{unique} \
+            SORT=~{sort} \
+            INPUT=~{interval_list} \
+            OUTPUT=out
```
a-frantz, line 964: One of the many annoying things about picard. It has two CL input formats. `ALL_CAPS=<>` and `--ALL_CAPS <>`. The former used to be the only supported format. Than they added the later [and bizarrely don't allow them to mix]. Our repo used to be all "former" style like you use here. I went and did a mass update of the Picard tasks to use the later style. Would appreciate keeping that consistent.

tools/picard.wdl
```diff
@@ -866,3 +866,182 @@ task bam_to_fastq {
...
+        modify_disk_size_gb: "Add to or subtract from dynamic disk space allocation. Default disk size is determined by the size of the inputs. Specified in GB."
+    }
+
+    input {
+        File interval_list
+        String subdivision_mode = "BALANCING_WITHOUT_INTERVAL_SUBDIVISION_WITH_OVERFLOW"
+        Boolean unique = true
+        Boolean sort = true
+        Int scatter_count
+        Int modify_disk_size_gb = 0
```
a-frantz, line 948:
```suggestion
```

tools/picard.wdl
```diff
@@ -866,3 +866,182 @@ task bam_to_fastq {
...
+    >>>
+
+    output {
+        Array[File] interval_lists_scatter = glob("out/*/*.interval_list")
+        Int interval_count = read_int("interval_count.txt")
+    }
+
+    runtime {
+        memory: "2 GB"
+        disk: "~{disk_size_gb} GB"
```
a-frantz, line 986:
```suggestion
        disk: "1 GB"
```
In the rest of the repo we use the magic number of 10gb for the lowest disk size. But I think you were running and testing this on the dnanexus, right? So if 1gb works in that environment, that's a good argument we're being _way_ over generous with our 10gb magic number.

tools/sambamba.wdl
```diff
@@ -0,0 +1,289 @@
...
+
+    output {
+        File duplicate_marked_bam = "~{prefix}.markdup.bam"
+        File duplicate_marked_bam_index = "~{prefix}.markdup.bam.bai"
+        File markdup_log = "~{prefix}.markdup_log.txt"
+    }
+
+    runtime {
+        cpu: ncpu
+        memory: "50 GB"
```
a-frantz, line 227: This should probably be made dynamic. I don't want to hold up the PR with a series of tests to find the right formula though. So really just making a comment, not expecting any change right now.

workflows/reference/gatk-reference.wdl
```diff
@@ -0,0 +1,108 @@
...
+import "../../tools/samtools.wdl"
+import "../../tools/util.wdl"
+
+workflow gatk_reference {
+    meta {
+        description: "Fetches reference files for GATK."
+        outputs: {
+            fasta: "FASTA file for the reference genome."
+            fasta_index: "Index for the FASTA file for the reference genome."
+            fasta_dict: "Picard Sequence Dictionary for the reference genome."
```
a-frantz, line 15: Are `.dict`s a Picard specific format?

tools/picard.wdl
adthrasher: I bumped it to 4 for consistency. `Xms` sets the initial heap size.

tools/gatk4.wdl
adthrasher: Good catch. Somehow I missed swapping that over.

tools/gatk4.wdl
adthrasher: Not a good one. I've added it.

tools/gatk4.wdl
adthrasher: It probably would be fine and likely the use case ~100% of the time. However, if someone is used to GATK, I think it may be confusing to only have one step where they expect two. Also, I suspect if the resource requirements were optimized, you'd get different node sizes, so you could probably save costs.

tools/sambamba.wdl
adthrasher: ```External sort is not implemented. Thus, memory consumption grows by 2Gb per each 100M reads. Check that you have enough RAM before running the tool.```
I'm curious if that is actually proportional to the read count or if it is actually proportional to the size of the file.

workflows/reference/gatk-reference.wdl
adthrasher: Not anymore. I think it was originally a Picard file, but now you can generate one with `samtools` as well.

tools/sambamba.wdl
a-frantz: Or, the more likely third option: neither lol.
Which is to say I imagine it correlates to both, and tracks neither perfectly.

a-frantz (approved):

PR #124, opened 2023-12-12
docs: doc update

a-frantz:
I'm including a link to these documents as part of the Rust `wdl` announcement. So I brought them up to date with our latest conventions.


template/task-templates.wdl
```diff
@@ -28,34 +24,32 @@ task static_disk_and_ram_task {
     }
 
     runtime {
-        memory: "~{memory_gb} GB"
-        disk: "~{disk_size_gb} GB"
+        memory: "<> GB"
+        disk: "<> GB"
         container: ""
```
adthrasher, line 29: Should we put a reasonable default for `container`? Likely it will almost always be overridden.

best-practices.md
```diff
@@ -23,13 +23,6 @@ All rules below should be followed by contributors to this repo. Contributors sh
...
       - e.g. `memory_gb` and `modify_memory_gb` cannot be present in the same task.
-- All tasks and workflows should have a `max_retries` input.
-  - This should be defaulted to `1` for nearly all tasks
-  - Some tasks are particularly error prone and can have a higher default `max_retries`
-  - **rule specific to workflows:** `max_retries` should be an optional `Int?`
-    - This allows each task to have it's own specific default `max_retries`
-      - If a user does not supply `max_retries`, those task level defaults will get used
-      - If a user does supply `max_retries`, it should override the default for *every* task called
 - multi-core tasks should *always* follow the conventions laid out in the `use_all_cores_task` example (see `template/task-templates.wdl`)
   - this is catering to cloud users, who may be allocated a machine with more cores than are specified by the `ncpu` parameter
```
adthrasher, line 27: I wonder if we should note that `ncpu` will be deprecated when a task body can access `runtime` fields.

template/task-templates.wdl
```diff
@@ -69,29 +63,23 @@ task dynamic_disk_and_ram_task {
...
-        max_retries: "Number of times to retry in case of failure"
     }
 
     input {
         Boolean use_all_cores = false
-        Int ncpu = 1
-        Int memory_gb = <>
-        Int disk_size_gb = <>
-        Int max_retries = 1
+        Int ncpu = <>
```
adthrasher, line 82: I suppose this comes down to how one thinks about templates, but should we leave a reasonable default, e.g. `Int ncpu = 1`. I would think it would be clear that one might want to update this, but it would have a valid value. The same may be true for `memory` and `disk` in the `runtime` section.

adthrasher (approved):

template/task-templates.wdl
adthrasher: I think it depends. Pretty much all of our `util` tasks could run in a default container, such as `ubuntu:22.04`. So it isn't wrong in all cases. Are our templates designed to be just a fillable form or do we want to provide a as close to working as possible starting point?

best-practices.md
a-frantz: I don't think it hurts. But we _are_ planning to keep the `use_all_cores` convention regardless, correct?

template/task-templates.wdl
a-frantz: I think a reasonable setting for all of them would be our lower bound for the repo (which should also be documented). So 4gb mem, 10gb disk, 2cpu. I'll go and make those updates.

best-practices.md
adthrasher: I think so, but we've only retained the `ncpu` input parameter because we can't access the runtime. Hopefully with 2.0 
```
        n_cores=~{ncpu}
        if ~{use_all_cores}; then
            n_cores=$(nproc)
        fi
```
Will become something like
```
        n_cores=~{task.runtime.cpu}
        if ~{use_all_cores}; then
            n_cores=$(nproc)
        fi
```
And we can eliminate the `input` parameter in favor of querying the `runtime` directly.

template/task-templates.wdl
adthrasher: I think setting these argues in favor of providing a default for `container` as well, but I'm interested to hear your opinion as well.

template/task-templates.wdl
a-frantz: Here's my hold-up on providing a default container. We want the container selected for a task to be _targeted_. We have had internal discussions about how much we dislike the "kitchen sink" convention encouraged by certain set ups. A container should be as minimal as possible, right? That seems a good convention to stick by. And by that philosophy, a "default container" isn't very sensible, is it?

But we do have the `util` image, which is admittedly a bit of a kitchen sink container. We aren't always following our own best practices here. (I _DO NOT_ want to split `util` into more targeted containers. That is way more work than it's worth, but I think it's worth acknowledging that touch of hypocrisy we have going on).

Hypothetical scenario: You are an external collaborator looking to add some tasks to this repo as a good samaritan. You have not been privy to all our internal discussions about containers and our philosophy. You start developing with our templates, which have `util` already filled in.
You realize that `util` doesn't cut it for your use case. What's your first instinct here? To shop around for a different extant container that meets your needs? Or modify util? It's the default, right? Shouldn't it meet your needs?
Contrast to the hypothetical where no default container is supplied. First thing you do as a developer would be find a container to suit your use case.

IMO the latter is the correct development pattern. Pick a container first, write your task around what is available in your image. Write a new container (or modify an extant one) only as a last resort.

So I still think "no default container". But I am close to being persuaded here. Do you have any more thoughts?

adthrasher: You make good points. I wouldn't use the `util` image as the default, but something public like `ubuntu` or `alpine` (open to others). I'd probably suggest an in-line comment in the template to reinforce the idea that that value should likely be changed by the dev. That said, I can see leaving it empty as well. 

Do you happen to know if an empty string in the `container` field is valid WDL? Or if it is included, it must be a valid container path?

adthrasher (approved):

PR #126, opened 2023-12-13
docs: gut the README of outdated information. Fill back in sparsely.

a-frantz:
title. As stated in Slack, I am opting to err on the side of understating VS overstating. IMO the goal should be to remove all the bad info (✅ ) and fill back in the bare minimum needed for the README to not be embarrassing.

IMO we remove the whole `conf` and `bin` directories, but I didn't pull the trigger on that yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,10 +42,18 @@ The repository is laid out as follows:
 ## Tests
 
 Every task in this repository is covered by at least one test (see all of our tests in `tests/tools/`). These are run using [pytest-workflow](https://pytest-workflow.readthedocs.io/en/stable/).
+The command for running our tests should be executed at the root of the repo: `python -m pytest --kwdof --git-aware`
 
 ## 🤝 Contributing
 
 Contributions, issues and feature requests are welcome!<br />Feel free to check [issues page](https://github.com/stjudecloud/workflows/issues). You can also take a look at the [contributing guide](https://github.com/stjudecloud/workflows/blob/master/CONTRIBUTING.md).
+
+## Links worth checking out
+
+[The OpenWDL GitHub](https://github.com/openwdl)
+Our preferred WDL runner: [miniwdl](https://github.com/chanzuckerberg/miniwdl)
+Most of our tasks are run inside a [BioContainers image](https://github.com/BioContainers/containers)
+Our tasks are validated using [pytest-workflow](https://pytest-workflow.readthedocs.io/en/stable/)
 
 ## 📝 License
 
--- a/README.md
+++ b/README.md
@@ -18,51 +18,18 @@ Resources requirements have been optimized to minimize failures in our computing
 
 ### 🏠 [Homepage](https://stjude.cloud)
 
-## Getting Started
-
-At the time of writing, all workflows are written in [WDL][wdl] and are tested
-using [Cromwell][cromwell]. We use [Oliver][oliver] to easily interact with the
-Cromwell server to perform various tasks. Although we do not test outside of Cromwell, we
-expect that the workflows will work just as well using other runners.
-
-The easiest way to get started is to install [bioconda][bioconda] and the run the following commands:
-
-```bash
-conda create -n workflows-dev -c conda-forge cromwell -y
-conda activate workflows-dev
-git clone git@github.com:stjudecloud/workflows.git
-cd workflows
-```
-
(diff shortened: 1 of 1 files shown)


adthrasher:
> IMO we remove the whole `conf` and `bin` directories, but I didn't pull the trigger on that yet.
I think I'd leave them for now. I think we can mark them unused as you have done in this PR, but I think there may still be others using them. I believe Modupe was using them at one point. They're also documented and linked from some of our internal wiki docs on WDL.

a-frantz:
> > IMO we remove the whole `conf` and `bin` directories, but I didn't pull the trigger on that yet.
> 
> I think I'd leave them for now. I think we can mark them unused as you have done in this PR, but I think there may still be others using them. I believe Modupe was using them at one point. They're also documented and linked from some of our internal wiki docs on WDL.

Fair enough. No disagreement here. We can leave them for now.

README.md
```diff
@@ -84,8 +51,3 @@ Contributions, issues and feature requests are welcome!<br />Feel free to check
 
 Copyright © 2020-Present [St. Jude Cloud Team](https://github.com/stjudecloud).<br />
 This project is [MIT](https://github.com/stjudecloud/workflows/blob/master/LICENSE.md) licensed.
-
-[wdl]: http://openwdl.org/
-[cromwell]: https://github.com/broadinstitute/cromwell
-[bioconda]: https://bioconda.github.io/
-[oliver]: https://github.com/stjudecloud/oliver
```
adthrasher, line 91: I actually like having this, but I'd make it a section. Link to the WDL spec, `miniwdl`, `pytest-workflows`, and anything else we think is an important part of the WDL ecosystem.

adthrasher: To be clear, I'd remove these links and do a replacement with more relevant ones.

README.md
```diff
@@ -18,51 +18,18 @@ Resources requirements have been optimized to minimize failures in our computing
...
-| Build BWA References   | N/A             | Builds reference files used by the [BWA aligner](https://github.com/lh3/bwa).                                             | None                                                                                                  | [Workflow](./workflows/general/bwa-db-build.wdl)                                                                               | ![In Production](https://img.shields.io/static/v1?label=Status&message=Production&color=green&style=flat-square)    |
-| BAM to FASTQs          | v1.0.0          | Split a BAM file into read groups, then read 1 FASTQs and  read 2 FASTQs.                                                 | None                                                                                                  | [Workflow](./workflows/general/bam-to-fastqs.wdl)                                                                              | ![In Production](https://img.shields.io/static/v1?label=Status&message=Production&color=green&style=flat-square)    |
```
adthrasher, line 65: Are you planning to reintroduce this section later with up-to-date information? Or is this intended as a permanent deletion?

a-frantz: Yes, I intend for this section to come back. IDK if this format is the best format we can present the information in (maybe it is), but we will definitely bring this info back.

README.md
```diff
@@ -74,7 +41,7 @@ The current workflows exist in this repo with the following statuses:
 
 ## Tests
 
-Given that this repo is still new, there are no tests. When we add tests, we will update the README.
+Every task in this repository is covered by at least one test (see all of our tests in `tests/tools/`). These are run using [pytest-workflow](https://pytest-workflow.readthedocs.io/en/stable/).
```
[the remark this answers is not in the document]
a-frantz: We probably should, but that would also need environment bootstrapping instructions. Probably want to include miniwdl config instructions bc miniwdl is hardcoded. IDK. Sounds like a hassle to fully document.

Do you think we can get away with just the following?
```
Tests can be run with the following command: `python -m pytest-workflow --git-aware`
```
And _*only*_ that? Can we let the user figure out the environment on their own? It's not _that_ complicated? (Yes I am just being lazy on this PR 😅 )

README.md
a-frantz: Alright. No disagreement. Will add that first thing in the AM.

README.md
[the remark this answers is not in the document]
adthrasher: I think that's fine for now. We should probably have a dev environment definition eventually. Since we're using Docker containers to do the actual runs, it should be relatively simple.

adthrasher (approved):

PR #128, opened 2024-01-10
fix: correct extension

adthrasher:


a-frantz (approved):
Good catch

adthrasher:
> Good catch

Actually, one of my workflows failed, which is how I found this.

PR #129, opened 2024-01-31
Add test runner CI workflow

adthrasher:
Adds a pytest-workflow runner step to our CI workflows.

Fixes broken tests.

Removes `docker://` prefixes from `container` entries.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -22,4 +22,8 @@ jobs:
         find tests -name '*.yaml' -exec yq --output-format yaml -i 'del(.[] | select(.tags[] | test("reference|deprecated|cellranger") ) )' {} \;
     - name: Run pytest-workflow
       run: |
-        pytest --git-aware --workflow-threads 4
+        pytest --version
+        miniwdl --version
+        pip show pytest-workflow
+        python --version
+        pytest --git-aware --workflow-threads 2
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -17,7 +17,7 @@ jobs:
         pip install miniwdl pytest-workflow
     - name: filter tests
       run: | 
-        find tests -name '*.yaml' -exec yq -i '. - map(select(.tags[] | contains ("reference") or contains ("deprecated") or contains ("cellranger")))' {} \;
+        find tests -name '*.yaml' -exec yq -y -i '. - map(select(.tags[] | contains ("reference") or contains ("deprecated") or contains ("cellranger")))' {} \;
     - name: Run pytest-workflow
       run: |
         pytest --git-aware --workflow-threads 4
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -1,0 +1,23 @@
+name: Pytest-Workflows Check
+
+on: [push]
+
+jobs:
+  miniwdl_check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v1
+      with:
+        python-version: '3.10'
+    - name: Install miniwdl and pytest-workflow
+      run: |
+        python -m pip install --upgrade pip
+        pip install miniwdl pytest-workflow
+    - name: filter tests
+      run: | 
+        find tests -name '*.yaml' -exec yq -i '. - map(select(.tags[] | contains ("reference") or contains ("deprecated") or contains("cellranger")))' {} \;
+    - name: Run pytest-workflow
+      run: |
+        pytest --git-aware --workflow-threads 4
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -6,7 +6,7 @@ jobs:
   pytest_check:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
       with:
         lfs: true
     - name: Set up Python
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -22,9 +22,4 @@ jobs:
         find tests -name '*.yaml' -exec yq --output-format yaml -i 'del(.[] | select(.tags[] | test("reference|deprecated|cellranger") ) )' {} \;
     - name: Run pytest-workflow
       run: |
-        df -h
-        pytest --version
-        miniwdl --version
-        pip show pytest-workflow
-        python --version
         pytest --git-aware --basetemp /home/runner/work/pytest --symlink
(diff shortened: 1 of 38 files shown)


.github/workflows/pytest.yaml
```diff
@@ -0,0 +1,25 @@
+name: Pytest-Workflows Check
```
a-frantz, line 1:
```suggestion
name: pytest-workflow Check
```

tools/util.wdl
```diff
@@ -353,7 +353,7 @@ task unpack_tarball {
 
         mkdir unpacked_tarball
         tar -C unpacked_tarball -xzf ~{tarball} --no-same-owner
-        find unpacked_tarball/ -type f > file_list.txt
+        find unpacked_tarball/ -type f | LC_ALL=C sort > file_list.txt
```
a-frantz, line 356: When the "find order" bug was first identified, we discussed this exact fix. And we opted not to implement it. Not 100% certain this was the original rationale, but I think we decided it "wasn't the task's job" to have sorted output.  Or rather, having a `sort` statement would raise questions about "why". I don't think I still agree with that though.

I'm fine with this, but just documenting the historical rationale. Please delete the comment about this in the `yaml` file though.

a-frantz (changes requested):
See comments for requested changes.

I've asked this before, why are we specifying `--copy-input-files` all the time? You said you had disk size issues, wouldn't this be the biggest culprit? Are we sure we need that flag?

tools/util.wdl
adthrasher: I don't fully recall that conversion, but I will note that one of the `multiqc` tasks explicitly sets `LC_ALL`. So this wouldn't be the only one. 

That said, I dug into this and it seems that `find` has no inherent sorting. It is fully up to the order that the filesystem returns items. For all practical purposes (other than this test), that should be fine. I would hope that no downstream task would `untar` many files and then try to address them by index. 

For example, here is the output of untarring on the cluster.

```
$ tar -C unpacked_tarball -vxzf test.tar.gz --no-same-owner
test/
test/b
test/a

$ find unpacked_tarball/ -type f
unpacked_tarball/test/b
unpacked_tarball/test/a
```

We can see that `tar` processes `b` and then `a` and the `find` command respects that ordering. 

For another example, I ran this on my Mac in a Ubuntu Docker container.

```
# tar -C unpacked_tarball -vxzf /data/tests/tools/input/test.tar.gz --no-same-owner
test/
test/b
test/a

# find unpacked_tarball/ -type f
unpacked_tarball/test/a
unpacked_tarball/test/b
```

Again `tar` processes `b` and then `a`, but the `find` command reverses that ordering.

Looking at `stat` output on the cluster, the inode and timestamps (except `modify`) all make sense for `b` preceding `a`.

```
$ stat unpacked_tarball/test/a
  File: unpacked_tarball/test/a
  Size: 4               Blocks: 1          IO Block: 4194304 regular file
Device: 34h/52d Inode: 14406740294  Links: 1
Access: (0770/-rwxrwx---)  Uid: (1018972/athrashe)   Gid: (76937/Domain Users)
Access: 2024-01-31 15:37:04.258968000 -0600
Modify: 2023-09-20 11:15:13.000000000 -0500
Change: 2024-01-31 15:37:04.259195840 -0600
 Birth: -

$ stat unpacked_tarball/test/b
  File: unpacked_tarball/test/b
  Size: 4               Blocks: 1          IO Block: 4194304 regular file
Device: 34h/52d Inode: 14406740293  Links: 1
Access: (0770/-rwxrwx---)  Uid: (1018972/athrashe)   Gid: (76937/Domain Users)
Access: 2024-01-31 15:37:04.258462000 -0600
Modify: 2023-09-20 11:15:24.000000000 -0500
Change: 2024-01-31 15:37:04.258663107 -0600
 Birth: -
```

On my mac, the stat is similar (other than timezone). 

```
# stat unpacked_tarball/test/a
  File: unpacked_tarball/test/a
  Size: 4         	Blocks: 8          IO Block: 4096   regular file
Device: 88h/136d	Inode: 3676512     Links: 1
Access: (0750/-rwxr-x---)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2024-01-31 21:38:23.650214005 +0000
Modify: 2023-09-20 16:15:13.000000000 +0000
Change: 2024-01-31 21:38:23.651214005 +0000
 Birth: 2024-01-31 21:38:23.650214005 +0000

# stat unpacked_tarball/test/b
  File: unpacked_tarball/test/b
  Size: 4         	Blocks: 8          IO Block: 4096   regular file
Device: 88h/136d	Inode: 3676511     Links: 1
Access: (0750/-rwxr-x---)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2024-01-31 21:38:23.649214005 +0000
Modify: 2023-09-20 16:15:24.000000000 +0000
Change: 2024-01-31 21:38:23.650214005 +0000
 Birth: 2024-01-31 21:38:23.649214005 +0000
```

Yet the output order of `find` is reversed. So the difference must be something in the implementation of GPFS vs APFS. I've tried options such as `tar`'s `--preserve-order`, but it doesn't seem to be `tar` output ordering, it is the underlying file system ordering.

adthrasher: Looking at this again, I'm less convinced that omitting a `sort` is proper. We're not returning the structure of the `tar` file. We're explicitly deconstructing it and returning only a listing of files that were contained. Without the directory structure, does that make sense to return an unsorted list?

a-frantz: Ahh I remembered more of the original conversation thanks to this sentence:

> I would hope that no downstream task would untar many files and then try to address them by index.

The rationale was "if we guarantee an output order, regardless of what that order is, we might encourage users to try indexing into the produced array looking for file X. To prevent that, let's leave the order up to the filesystem which is inherently unreliable, preventing that use case".

I don''t agree with that anymore. My current thinking: if we _don't_ pass it through `sort`, we need to write a custom test. If we _do_ pass it through sort, our current test works. I do not want to write a custom test. Let's just pass it through `sort` and call it a day.

But I'd make one last change request. Can we put a comment which says `# pipe through sort because otherwise order is random (dependent on filesystem)` or something similar?

adthrasher:
> See comments for requested changes.
> 
> I've asked this before, why are we specifying `--copy-input-files` all the time? You said you had disk size issues, wouldn't this be the biggest culprit? Are we sure we need that flag?

I'll make the change and see what happens. When I ran locally initially, it failed without `--copy-input-files`.

tools/util.wdl
a-frantz: > Please delete the comment about this in the yaml file though.

https://github.com/stjudecloud/workflows/blob/main/tests/tools/test_util.yaml#L85

^that comment should be deleted now that we've settled this behavior.

After that I'll approve the changes

a-frantz (approved):

PR #132, opened 2024-02-16
chore: update versions

a-frantz:


a-frantz:
If I am not mistaken, the tests are about to fail due to STAR trying to pull non-existent container images. Those links will be corrected once this gets merged into `main`. We should maybe rethink this set-up? No decent alternatives are coming to mind though.

adthrasher (approved):

adthrasher:
> If I am not mistaken, the tests are about to fail due to STAR trying to pull non-existent container images. Those links will be corrected once this gets merged into `main`. We should maybe rethink this set-up? No decent alternatives are coming to mind though.

That's an interesting problem. I don't have any great thoughts on how to get around it for now.

PR #133, opened 2024-02-17
feat: opt-in to run kraken twice, second time on only mapped reads

a-frantz:
This might be a dud of an analysis. I've only run it on one sample, a "Cellular Organism" case from the latest RTCG batch, and this extra Kraken run didn't reveal anything interesting.
[multiqc_report.html.txt](https://github.com/stjudecloud/workflows/files/14323333/multiqc_report.html.txt)

Do we want to merge this now? It works as expected. The analysis is sound in theory. Remains to be seen if it's useful in practice. There's been a test with `n` of `1` which doesn't tell us much of anything.
I vote merge now, with the possibility we revert the commit sometime down the line when we've used it more. I want to see these result before and after XenoCP. That might be neat. Probably some other use cases we will run into in the future(?)

Voting against leaving this PR open, because I want to use it and I don't want to deal with keeping it up to date with `main`, which has other work going on. If `main` were stagnant I'd feel differently.


adthrasher (approved):
I think this could be quite useful for some cases. It's useful to know when mapped reads are better allocated to a different genome. I agree with this being disabled by default.

a-frantz:
@adthrasher I was about to hit merge when I got an idea.

Should the `F = "0x904"` line be at the workflow input level? (With a different param name) Perhaps all the `samtools fastq` filters should be brought up to the workflow level? (I think we'd want to obfuscate them in a struct, but I'm still in the "spit-ball" idea phase here.)

My thinking: there are a few tangentially related questions you can ask by filtering-pre-Kraken. I don't want to support them all individually, but we have a mechanism to abstract here and allow the user to construct their own comparative filter.

This PR currently allows you to ask "what species do the mapped reads best map to?" I can certainly see someone wanting to know "what species do the _unmapped_ reads best map to?" A little more niche question, but one I see a use case for.

I haven't worked through all the details yet, but LMK if you think this is worth pursuing. A difficulty here will be sensibly constructing the docs. But we can get through that.

adthrasher:
I think having a `struct` to wrap the flag filters is worth pursuing. Right now, you've documented in comments, but if you had the right `struct`, you wouldn't need the comments. It would also be potentially useful in other filtering.

a-frantz:
@adthrasher Please pay special attention to `data_structures/flag_filter.wdl`. Of course we don't need to feel tied to any precedent set here, but it would be nice to get the whole "`data_structures` structure" right the first time. So let's dot our 'i's and all that to make sure we're happy with this direction.

data_structures/flag_filter.wdl
```diff
@@ -0,0 +1,157 @@
+## # FlagFilter
+##
+## A struct to represent the flags used in the `samtools fastq` command.
```
adthrasher, line 3: These are also supported by `view`. There may be others as well.

data_structures/flag_filter.wdl
```diff
@@ -0,0 +1,157 @@
...
+version 1.1
+
+struct FlagFilter {
+    String include_if_all  # samtools -f
+    String exclude_if_any  # samtools -F
+    String include_if_any  # samtools --rf
+    String exclude_if_all  # samtools -G
+}
+
+task validate_string_is_oct_dec_or_hex {
```
adthrasher, line 71: I'm not sure that we'd ever use this anywhere else, but this doesn't necessarily have to be specific to this data structure.

tests/data_structures/test_flag_filter.yaml
```diff
@@ -0,0 +1,61 @@
...
+  command: >-
+    miniwdl run -d test-output/. --task validate_string_is_oct_dec_or_hex data_structures/flag_filter.wdl number="01"
+
+- name: flag_filter_0x1000
+  tags:
+    - miniwdl
+    - flag_filter
+  command: >-
+    miniwdl run -d test-output/. --task validate_string_is_oct_dec_or_hex data_structures/flag_filter.wdl number="0x1000"
+  exit_code: 42
```
adthrasher, line 21: In the `pytest-workflow` docs, the author uses a `should fail` tag for tasks like this. Should we adopt something similar? Arguably, specifying the non-zero exit code may be sufficient.

a-frantz: Yes, I saw that in the `pytest-workflow` docs and thought it was odd. My understanding of the tag feature is to enable/disable groups of tests from the command line. I don't see when you'd ever want to toggle all the "should fail" tests at once.

I think it's safe to skip that tag.

data_structures/flag_filter.wdl
a-frantz: That's a good point. Although I'm worried that the order of precedence might not be consistent across subcommands. That's just a hunch. I'll look into it tomorrow and see if we can broaden the scope a bit on what other commands can use this struct.

data_structures/flag_filter.wdl
a-frantz: I started typing a disagreement with you, but I came around. This can be abstracted some so it can be re-used (hypothetically) and I think that's a good idea, but where would it belong? I feel like it kinda has to go into a file of its own, which seems a little silly. And I also wouldn't call it a data structure, which means we'd need... `validation/` in the root???
I'm not sure what to do with it 🤷

tools/samtools.wdl
```diff
@@ -285,7 +294,14 @@ task subsample {
...
+            --threads "$n_cores" \
+            -f ~{filter.include_if_all} \
+            -F ~{filter.exclude_if_any} \
+            --rf ~{filter.include_if_any} \
+            -G ~{filter.exclude_if_all} \
+            -c ~{bam} \
+        )"
 
         if [[ "$read_count" -gt "~{desired_reads}" ]]; then
             # the BAM has at least ~{desired_reads} reads, meaning we should
```
a-frantz, line 307: I just pushed these changes and then immediately realized the logic was busted.

If a filter is enabled and the resulting count is smaller than `desired_count`, the unfiltered bam will be output.
I'm not entirely sure how this logic _should_ work, but I know it isn't that.

The most straightforward fix is to drop the `-c` flag here, pipe through `tee tmp.bam | wc -l` and then if `read_count < desired_reads` then `mv tmp.bam $bam` and output the filtered BAM. (That won't work bc the input BAM is read-only, but whatever you get the gist.)

My issue with this is that with an arbitrarily large `desired_reads` and an active filter, this task becomes `samtools view $filter > output.bam` with a whole lot of complicated stuff wrapped around it doing nothing. Which makes me ask the question, shouldn't we just have a `samtools view $filter > output.bam` task by itself? I think the answer is "yes" and that would be easy for me to add. But then that new task is overlapping with this task under certain criteria. But maybe that's ok?

OR. Maybe this task needs a rebrand. Spit balling: but call it `samtools.filter_and_subsample`. Expose all the `samtools view` options. Keep all our subsampling logic. But instead of writing this task with the assumption that the BAM should be subsampled, that's just an option.
Problem with that idea: it's freakin complicated. That's gonna be messy documentation. But IDK. I think it's the right call. Gonna take lunch and come back to this problem.

tools/samtools.wdl
```diff
@@ -332,6 +387,101 @@ task subsample {
...
+task filter {
+    # TODO expose more `view` options
+    meta {
+        description: "Filters a BAM based on its bitwise flag value."
+        help: "This task is a wrapper around `samtools view`. This task will fail if there are no reads in the output BAM. This can happen either because the input BAM was empty or because the supplied `bitwise_filter` was too strict. This task is not suitable for random subsampling. If you want to subsample a BAM, use the `subsample` task instead."
+    }
+
+    parameter_meta {
+        bam: "Input BAM format file to filter"
+        bitwise_filter: "A set of 4 possible read filters to apply. This is a `FlagFilter` object (see ../data_structures/flag_filter.wdl for more information)."
```
a-frantz, line 399: I don't want to have a detailed explanation of this struct every task it's used, but this I imagine this documentation leads to a bad user experience (particularly as this task doesn't have a default, so _all_ users _must_ go and read another file).

I don't have a solution, I guess I'm just complaining. Any thoughts on this issue?

tests/data_structures/test_flag_filter.yaml
adthrasher: I think the use of `tags` is a little overloaded. It could be useful as a documentation field as well as a togglable set of tests to run.

tools/samtools.wdl
adthrasher: I actually don't mind this. I also do not want to copy the explanation to every usage. Filtering is a fairly manual process anyway and it seems reasonable to require users to read the docs first.

tools/samtools.wdl
```diff
@@ -334,6 +388,101 @@ task subsample {
...
+        # TODO should there be a check that the output
+        # BAM is different from the input?
+        # It would only need to be a simple `wc -l` check.
+        # Ideally this would be done _without_ an extra pass through the BAM.
+        # An easier check is that the supplied filters aren't all zeroes.
+        # However it's still possible to have an active filter
+        # that doesn't catch anything. What's the ideal
+        # behavior in that case? The user provided something valid,
+        # so failing seems wrong. But (essentially) duplicating the BAM is also
+        # (most likely) wrong.
```
adthrasher, line 470: I don't think we should add this check. I can see a use case where someone wants to drop all `duplicate` reads (or `supplemental` alignments), but not every BAM their pipeline ingests will have them. In that case, the task might often produce an identical BAM.

adthrasher (approved):

tools/samtools.wdl
a-frantz: That's a good point. I haven't made a ton of headway on `sprocket doc` yet, but I've been toying with the idea of internal links. That would be really cool to add, and this is a prime use case. So I guess I'll leave this as is for now. And I'll revisit if/when I add the capability to link between Markdown files.

PR #134, opened 2024-02-17
fix(tools/star): default MAPQ of 254 instead of 255

a-frantz:


adthrasher (approved):
My only feedback would be, I'm not sure that Alex did mean 254. In this thread (https://github.com/alexdobin/STAR/issues/1069), I proposed that change and I'm not sure if it was rejected or if there was a misunderstanding.

PR #137, opened 2024-02-22
Arriba

adthrasher:
Adding support for Arriba. The most controversial thing is likely to be the changes I made to the STAR defaults. Some are required to output chimeric alignments. Others are likely more debatable. The changes reflect the parameters that Arriba uses when running STAR.


tools/arriba.wdl
```diff
@@ -0,0 +1,315 @@
+## [Homepage](https://arriba.readthedocs.io/en/latest/)
+#
+# SPDX-License-Identifier: MIT
+# Copyright St. Jude Children's Research Hospital
+version 1.1
+
+task arriba {
+    meta {
+        description: "Run Arriba structural variant caller on a STAR-produced BAM file."
```
a-frantz, line 9: Question: Does it _have_ to be STAR-produced? How exhaustive are Arriba's assumption that it's STAR-produced? I would assume to get data from another aligner through Arriba you'd have to do some post-processing. Is it enough post-processing that we can say "no one in their right mind would do it that way"?

Note: I didn't check if Arriba docs specify the STAR assumption. If they do, the above is fine as it is and you don't have to dig for more information. But I would still be curious about the answer to those questions haha.

tools/arriba.wdl
```diff
@@ -0,0 +1,315 @@
...
+        outputs: {
+            fusions: "Output file of fusions in TSV format"
+            discarded_fusions: "Output file of discarded fusions in TSV format"
+        }
+    }
+
+    parameter_meta {
+        bam: "Input BAM format file on which to call fusions"
```
a-frantz, line 17:
```suggestion
        bam: "Input BAM format file from which to call fusions"
```

tools/arriba.wdl
```diff
@@ -0,0 +1,315 @@
...
+        gtf: "GTF features file. Gzipped or uncompressed."
+        reference_fasta_gz: "Gzipped reference genome in FASTA format"
+        chimeric_sam: "Optional input file of chimeric reads in SAM format, from older versions of STAR"
+        exclude_list: "Optional input file of regions to exclude from analysis in tab delimited format"
+        known_fusions: "Optional input file of known fusions in tab delimited format"
+        annotate_fusions: "Optional input file in tab delimited format of fusions to annotate with tags"
+        protein_domains: "Optional input file of protein domains coordinates in GFF3 format"
+        wgs_svs: "Optional input file of structural variants found by WGS in tab delimited or VCF format"
```
a-frantz, line 25: Can we get a link to how these should be formatted? I think just saying `in tab delimited format` is not enough. Reading that, **I** do not know how to write/supply these extra files. So we need more detail or instruction. But I don't want these to become super-long because it's detailing a whole file specification. That would be a pain to write and a pain to read.

I think adding `external_help` to each of these would be good. Would it be weird if all of them have the same link? Maybe instead, add `See 'meta.external_help' for more information.` (and obviously put a link at that location)

tools/arriba.wdl
```diff
@@ -0,0 +1,315 @@
...
+        min_anchor_length: "Minimum anchor length for split reads. Default is 23."
+        many_spliced_events: "Recover fusions with at least this many spliced breakpoints. Default is 4."
+        fragment_length: "For single-end data, this is the fragment length. With paired-end reads, this is ignored and determined automatically. Default is 200."
+        max_reads: "Subsample fusions with more than this number of reads. Default is 300."
+        top_n: "Only report the top N most highly expressed viral integration sites. Default is 5."
+        max_itd_length: "Maximum length of internal tandem duplications. Default is 100."
+        min_itd_supporting_reads: "Minimum number of supporting reads for internal tandem duplications. Default is 10."
+        modify_memory_gb: "Add to or subtract from dynamic memory allocation. Default memory is determined by the size of the inputs. Specified in GB."
+        modify_disk_size_gb: "Add to or subtract from dynamic disk space allocation. Default disk size is determined by the size of the inputs. Specified in GB."
+    }
```
a-frantz, line 65: This applies to a bunch of above parameters:
It's best to not refer to the default of a setting in the parameter_meta (unless there's a special circumstance*). This is because `wdldoc`/`sprocket doc` (which will hopefully be live sometime soon) pull the default value out of the WDL code and put it right next to the meta information.
So there are 2 problems with writing the default in the meta: 1) it could be wrong or out-of-date 2) it's going to be specified right there anyway so its redundant.

*a "special circumstance" for when to redundantly write the default is mostly a gut judgement. I can think of 2 cases where it makes sense. One would be where the type of the parameter is "weird" and talking about the default is a bit of translation from code to English (like the FlagFilter objects in #133 ). Two would be if _our_ default is different from _the tool's_ default.
For two, we can have a conversation if we want to keep this habit up, but I think it can prevent headaches for us, the developers to have that documented. Maybe it should move to a comment in the code rather than the meta? IDK. But I'd argue "good practice to note where we deviate from a tool's default behavior".

tools/arriba.wdl
```diff
@@ -0,0 +1,315 @@
...
+        Int top_n = 5
+        Int max_itd_length = 100
+        Int min_itd_supporting_reads = 10
+        Int modify_memory_gb = 0
+        Int modify_disk_size_gb = 0
+    }
+
+    Int input_size_gb = ceil(size(bam, "GiB"))
+    Int disk_size_gb = ceil(input_size_gb ) + ceil(size(gtf, "GiB")) + ceil(size(reference_fasta_gz, "GiB")) + modify_disk_size_gb
+    Int memory_gb = ceil(input_size_gb) + modify_memory_gb
```
a-frantz, line 111:
```suggestion
    Int bam_size_gb = ceil(size(bam, "GiB"))
    Int disk_size_gb = bam_size_gb + ceil(size(gtf, "GiB")) + ceil(size(reference_fasta_gz, "GiB")) + modify_disk_size_gb
    Int memory_gb = bam_size_gb + modify_memory_gb
```

tools/arriba.wdl
```diff
@@ -0,0 +1,315 @@
...
+            -Z ~{min_itd_supporting_reads} \
+            ~{if mark_duplicates then "" else "-u"} \
+            ~{if report_additional_columns_discard then "-X" else ""} \
+            ~{if fill_gaps then "-I" else ""}
+    >>>
+
+    output {
+        File fusions = "fusions.tsv"
+        File discarded_fusions = "fusions.discarded.tsv"
+    }
```
a-frantz, line 159: Are these considered intermediate files no one cares about? If not, can we rename them to share a basename with the BAM? This just looks weird to me, unless there's something I'm missing.

tools/arriba.wdl
```diff
@@ -0,0 +1,315 @@
...
+        outputs: {
+            fusion_bams: "Array of BAM files corresponding with fusions in the input file"
+            fusion_bam_indexes: "Array of BAM indexes corresponding with the BAMs in the 'fusion_bams'"
+        }
+    }
+
+    parameter_meta {
+        bam: "Input BAM format file on which to call fusions"
+        bam_index: "BAM index file corresponding to the input BAM"
+        fusions: "Input fusions in TSV format to convert to VCF"
```
a-frantz, line 229: I think this meta is from another task

tools/arriba.wdl
```diff
@@ -0,0 +1,315 @@
...
+        Int modify_disk_size_gb = 0
+    }
+
+    Int input_size_gb = ceil(size(gtf, "GiB"))
+    Int disk_size_gb = ceil(input_size_gb) + 5 + modify_disk_size_gb
+
+    String gene_model = basename(gtf, ".gz")
+
+    command <<<
+        if [ "~{gene_model}" != "~{gtf}" ]
```
a-frantz, line 293: This won't work as expected. `gtf` will have an absolute path, `gene_model` will only have a basename.

Is there a reason you can't copy+paste from one of our other tasks that does this? Using `ln -sf`?

a-frantz (commented):
This PR was bigger than I expected. I haven't finished yet, but I don't want to lose my progress so posting now.

tools/arriba.wdl
adthrasher: Let me take another look at this in the morning. Clearly, I missed a few things. I'll reping you when I've updated.

tools/arriba.wdl
adthrasher: Some of these (`exclude_list`, `known_fusions`, `protein_domains`) are included in the Arriba release. The others are optional.

tools/arriba.wdl
a-frantz: Interesting, I'm glad I asked! How I would document this, (though I don't think this is the only right way), is to leave the `description` vague and say it requires an RNA-Seq BAM. Then I'd use the `help` key to explain a summary of what you found. ("Typical is STAR, but DRAGEN and anything SPEC compliant for features XYZ will work")

tools/arriba.wdl
```diff
@@ -0,0 +1,325 @@
...
+        gtf: "GTF features file. Gzipped or uncompressed."
+        reference_fasta_gz: "Gzipped reference genome in FASTA format"
+        chimeric_sam: "Optional input file of chimeric reads in SAM format, from older versions of STAR"
+        exclude_list: {
+            description: "Optional input file of regions to exclude from analysis in tab delimited format"
+            external_help: "https://arriba.readthedocs.io/en/latest/input-files/#blacklist"
+        }
+        known_fusions: {
+            description: "Optional input file of known fusions in tab delimited format"
+            extenral_help: "https://arriba.readthedocs.io/en/latest/input-files/#known-fusions"
```
a-frantz, line 25:
```suggestion
            external_help: "https://arriba.readthedocs.io/en/latest/input-files/#known-fusions"
```

tools/arriba.wdl
```diff
@@ -0,0 +1,325 @@
...
+    meta {
+        description: "Annotate fusions with exon numbers."
+        outputs: {
+            fusion_bams: "Array of BAM files corresponding with fusions in the input file"
+            fusion_bam_indexes: "Array of BAM indexes corresponding with the BAMs in the 'fusion_bams'"
+        }
+    }
+
+    parameter_meta {
+        fusions: "Input fusions in TSV format to convert to VCF"
```
a-frantz, line 289: This is from another task

tools/arriba.wdl
```diff
@@ -0,0 +1,325 @@
...
+        gtf: "GTF features file. Gzipped or uncompressed."
+        modify_disk_size_gb: "Add to or subtract from dynamic disk space allocation. Default disk size is determined by the size of the inputs. Specified in GB."
+    }
+
+    input {
+        File fusions
+        File gtf
+        String prefix = basename(fusions, ".tsv") + ".annotated.tsv"
```
a-frantz, line 297:
```suggestion
        String outfile_name = basename(fusions, ".tsv") + ".annotated.tsv"
```

tools/arriba.wdl
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    runtime {
+        cpu: 1
+        memory: "4 GB"
+        disks: "~{disk_size_gb} GB"
+        container: "quay.io/biocontainers/arriba:2.4.0--h0033a41_2"
+        maxRetries: 1
+    }
+}
```
a-frantz, line 325:
```suggestion
}

```

tools/arriba.wdl
```diff
@@ -0,0 +1,325 @@
...
+    meta {
+        description: "Convert Arriba TSV format fusions to VCF format."
+        outputs: {
+            fusions_vcf: "Output file of fusions in VCF format"
+        }
+    }
+
+    parameter_meta {
+        fusions: "Input fusions in TSV format to convert to VCF"
+        reference_fasta_gz: "Gzipped reference genome in FASTA format"
```
a-frantz, line 194:
```suggestion
        reference_fasta: "Reference genome in FASTA format. Either gzipped or uncompressed."
```
And then switch the `gunzip -c` line to the `ln -sf` convention.

tools/arriba.wdl
```diff
@@ -0,0 +1,325 @@
...
+        File bam
+        File gtf
+        File reference_fasta_gz
+        File? chimeric_sam
+        File? exclude_list
+        File? known_fusions
+        File? annotate_fusions
+        File? protein_domains
+        File? wgs_svs
+        Array[String] interesting_contigs = ["1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12", "13", "14", "15", "16", "17", "18", "19", "20", "21", "22", "X", "Y", "AC_*", "NC_*"]
```
a-frantz, line 90: I would expect all these to have `chr` prefixes. Does Arriba handle both prefix+non-prefix gracefully?

tools/arriba.wdl
```diff
@@ -0,0 +1,325 @@
...
+            description: "Strandedness of the input data."
+            external_help: "https://arriba.readthedocs.io/en/latest/command-line-options/"
+            choices: [
+                "auto",
+                "yes",
+                "no",
+                "reverse"
+            ]
+        }
+        mark_duplicates: "Mark duplicates in the input BAM file with Arriba."
```
a-frantz, line 53: When this is enabled, does it create an output BAM? If so, that should be output. If not, I don't understand this parameter.

tools/arriba.wdl
```diff
@@ -0,0 +1,325 @@
...
+        File? exclude_list
+        File? known_fusions
+        File? annotate_fusions
+        File? protein_domains
+        File? wgs_svs
+        Array[String] interesting_contigs = ["1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12", "13", "14", "15", "16", "17", "18", "19", "20", "21", "22", "X", "Y", "AC_*", "NC_*"]
+        Array[String] viral_contigs = ["AC_*", "NC_*"]
+        Array[String] disable_filters = []
+        String feature_name = "gene_name=gene_name|gene_id,gene_id=gene_id,transcript_id=transcript_id,feature_exon=exon,feature_CDS=CDS"
+        String prefix = basename(bam, ".bam")
```
a-frantz, line 94:
```suggestion
        String prefix = basename(bam, ".bam") + ".fusions"
```
This seems like a nicer filename for downstream than just SJABC123_D.tsv

tools/arriba.wdl
```diff
@@ -0,0 +1,325 @@
...
+    parameter_meta {
+        fusions: "Input fusions in TSV format to convert to VCF"
+        reference_fasta_gz: "Gzipped reference genome in FASTA format"
+        modify_disk_size_gb: "Add to or subtract from dynamic disk space allocation. Default disk size is determined by the size of the inputs. Specified in GB."
+    }
+
+    input {
+        File fusions
+        File reference_fasta_gz
+        String vcf = basename(fusions, ".tsv") + ".vcf"
```
a-frantz, line 201: This is fine, if you understand you're letting users create VCF files with non-`.vcf` extensions. I would guess you'd rather enforce the .vcf filename?
```suggestion
        String prefix = basename(fusions, ".tsv")
```

tools/arriba.wdl
```diff
@@ -0,0 +1,325 @@
...
+        outputs: {
+            fusion_bams: "Array of BAM files corresponding with fusions in the input file"
+            fusion_bam_indexes: "Array of BAM indexes corresponding with the BAMs in the 'fusion_bams'"
+        }
+    }
+
+    parameter_meta {
+        bam: "Input BAM format file from which fusions were called"
+        bam_index: "BAM index file corresponding to the input BAM"
+        fusions: "Input fusions in TSV format to convert to VCF"
```
a-frantz, line 243: Also needs correction

tools/arriba.wdl
adthrasher: No, I think it computes duplicate reads on-the-fly when examining a potential fusion site.

```
-u : Arriba performs marking of duplicates internally based on identical mapping coordinates. When this switch is set, internal marking of duplicates is disabled and Arriba assumes that duplicates have been marked by a preceding program. In this case, Arriba only discards alignments flagged with the BAM_FDUP flag. This makes sense when duplicates cannot be reliably identified solely based on their mapping coordinates, e.g. when unique molecular identifiers (UMIs) are used or when independently generated libraries are merged in a single BAM file and the read group must be interrogated to distinguish duplicates from reads that map to the same coordinates by chance. In addition, when this switch is set, duplicate reads are not considered for the calculation of the coverage at fusion breakpoints (columns coverage1 and coverage2 in the output file).
```

tools/arriba.wdl
adthrasher: Yes, fortunately Arriba handles both.

```
-i CONTIGS : Comma-/space-separated list of interesting contigs. Fusions between genes on other contigs are ignored. Contigs can be specified with or without the prefix chr. Asterisks (*) are treated as wild-cards. Default: 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 X Y AC_* NC_*
```

tools/arriba.wdl
```diff
@@ -0,0 +1,325 @@
...
+task arriba_annotate_exon_numbers {
+    meta {
+        description: "Annotate fusions with exon numbers."
+        outputs: {
+            fusion_bams: "Array of BAM files corresponding with fusions in the input file"
+            fusion_bam_indexes: "Array of BAM indexes corresponding with the BAMs in the 'fusion_bams'"
+        }
+    }
+
+    parameter_meta {
```
a-frantz, line 288: missing meta for `outfile_name`

tools/arriba.wdl
```diff
@@ -0,0 +1,325 @@
...
+    }
+}
+
+task arriba_annotate_exon_numbers {
+    meta {
+        description: "Annotate fusions with exon numbers."
+        outputs: {
+            fusion_bams: "Array of BAM files corresponding with fusions in the input file"
+            fusion_bam_indexes: "Array of BAM indexes corresponding with the BAMs in the 'fusion_bams'"
+        }
```
a-frantz, line 285: bad copy +paste

tools/arriba.wdl
```diff
@@ -0,0 +1,325 @@
...
+            fusions: "Output file of fusions in TSV format"
+            discarded_fusions: "Output file of discarded fusions in TSV format"
+        }
+    }
+
+    parameter_meta {
+        bam: "Input BAM format file from which to call fusions"
+        gtf: "GTF features file. Gzipped or uncompressed."
+        reference_fasta_gz: "Gzipped reference genome in FASTA format"
+        chimeric_sam: "Optional input file of chimeric reads in SAM format, from older versions of STAR"
```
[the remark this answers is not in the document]
adthrasher: My approach is to treat this as a public resource. So I'm exposing all options.

tools/arriba.wdl
```diff
@@ -0,0 +1,372 @@
...
+        File fusions
+        File reference_fasta
+        String prefix = basename(fusions, ".tsv")
+        Int modify_disk_size_gb = 0
+    }
+
+    Int input_size_gb = ceil(size(fusions, "GiB"))
+    Int disk_size_gb = ceil(input_size_gb) + (ceil(size(reference_fasta, "GiB")) * 3) + modify_disk_size_gb
+
+    command <<<
```
a-frantz, line 254:
```suggestion
    command <<<
        set -euo pipefail

```

adthrasher:
@claymcleod - I don't think you need to review the Arriba changes, but I wanted to ping you for the STAR changes. I'm opting to change some parameters here to match Arriba recommendations. Most of the parameters relate to enabling chimeric alignment output, which _I think_ we intended to enable previously. However, there are a couple that are more broadly applicable.

tools/star.wdl
```diff
@@ -529,7 +529,7 @@ task alignment {
         String alignEndsType = "Local"
         String alignSoftClipAtReferenceEnds = "Yes"
         String alignInsertionFlush = "None"
-        String chimOutType = "Junctions"
+        String chimOutType = "WithinBAM HardClip"
```
claymcleod, line 532: I would think that soft clipping is a more reasonable choice for this as a default parameter, right? That way, no information is lost from the `SEQ` field? If Arriba needs it hardclipped, then I would recommend overriding this at the workflow level.

tools/star.wdl
```diff
@@ -503,10 +503,10 @@ task alignment {
             AT_AC_and_GT_AT_motif: 10
         }
         SJ_Motifs alignSJstitchMismatchNmax = SJ_Motifs {
-            noncanonical_motifs: 0,
+            noncanonical_motifs: 5,
```
claymcleod, line 506: Sure, this is probably fine if that's what Arriba recommends.

tools/star.wdl
```diff
@@ -540,7 +540,7 @@ task alignment {
         Float outFilterMatchNminOverLread = 0.66
         Float scoreGenomicLengthLog2scale = -0.25
         Float seedSearchStartLmaxOverLread = 1.0
-        Float alignSplicedMateMapLminOverLmate = 0.66
+        Float alignSplicedMateMapLminOverLmate = 0.5
```
claymcleod, line 543: This has very little information in the manual, so I'm not sure what it's doing. I see the value changes very little, but do you know if Arriba introduces any reasoning as to why they changed this?

tools/star.wdl
```diff
@@ -555,7 +555,7 @@ task alignment {
         Int outSAMflagOR = 0
         Int outSAMflagAND = 65535
         Int outFilterMultimapScoreRange = 1
-        Int outFilterMultimapNmax = 20
+        Int outFilterMultimapNmax = 50
```
claymcleod, line 558: I'm not thrilled about this one: the default is actually quite permissible, so thinking that there is going to be a valid range for 21-49 different loci _per read_ is unlikely IMO. That said, I don't have a sense of how many reads are mapped to `mapped to too many loci` today.

Personally, I think keeping the default selected by Alex and then overriding this for Arriba is also the way to go here.

tools/star.wdl
adthrasher: `HardClip` is actually the default in STAR. This only applies to `supplemental chimeric` alignments. The primary alignment should still have the full sequence, only the supplemental alignments will be hard clipped.


For example for the following read, only the `supplementary` alignments contain hard clipping.

```
A00467:240:HCN7KDRX2:2:2275:20410:25786 2481    chr1    24927   0       107H18M1S       chr14   49586604        0       GACCACCGGCACGCCCCCC     ,:F,:F,,,F,F,F,FF,F     NH:i:7  HI:i:4  AS:i:17 nM:i:0  NM:i:0  MD:Z:18 RG:Z:c949090    SA:Z:chr14,49586609,+,19S107M,0,10;
A00467:240:HCN7KDRX2:2:2275:20410:25786 2481    chr1    195452  0       107H18M1S       chr14   49586604        0       GACCACCGGCACGCCCCCC     ,:F,:F,,,F,F,F,FF,F     NH:i:7  HI:i:3  AS:i:17 nM:i:0  NM:i:0  MD:Z:18 RG:Z:c949090    SA:Z:chr14,49586609,+,19S107M,0,10;
A00467:240:HCN7KDRX2:2:2275:20410:25786 2481    chr1    46341588        0       110H16M chr14   49586604        0       CACCGGCACGCCCCCC        ,:F,,,F,F,F,FF,F        NH:i:7  HI:i:1  AS:i:13 nM:i:1  NM:i:1  MD:Z:0G15       RG:Z:c949090    SA:Z:chr14,49586606,+,16S110M,0,10;
A00467:240:HCN7KDRX2:2:2275:20410:25786 2481    chr1    46341589        0       111H15M chr9    9442086 0       ACCGGCACGCCCCCC :F,,,F,F,F,FF,F NH:i:7  HI:i:7  AS:i:14 nM:i:0  NM:i:0  MD:Z:15 RG:Z:c949090    SA:Z:chr9,9442085,+,15S111M,0,10;
A00467:240:HCN7KDRX2:2:2275:20410:25786 2465    chr2    113588497       0       1S18M107H       chr14   49586604        0       GGGGGGCGTGCCGGTGGTC     F,FF,F,F,F,,,F:,F:,     NH:i:7  HI:i:5  AS:i:17 nM:i:0  NM:i:0  MD:Z:18 RG:Z:c949090    SA:Z:chr14,49586609,+,19S107M,0,10;
A00467:240:HCN7KDRX2:2:2275:20410:25786 419     chr9    9442085 0       15S111M =       9442086 125     GGGGGGCGTGCCGGTGGTCCCAGCAACACGGGCGGCTGAGGCGGGAGGATCGATGGAGTCAAGGAGTTCTGGGGTGTAGTGCGCTATGCCGATCGGGTGTCCGCACTAACTTCGGCATCAAAATGG  F,FF,F,F,F,,,F:,F:,,::,F,F,,F,::,,::,:,F,:,:::F:,,::,,,:F:::,::::::,:F,FF,:F,:F,F:::FF,:,:::,F,:,,:::FFFFF,,F,,,:,:FFF:FF,F,,,  NH
:i:7    HI:i:7  AS:i:213        nM:i:10 NM:i:10 MD:Z:9T2T4A9T9C1T5C12C35G11T4   RG:Z:c949090    SA:Z:chr1,46341589,-,111H15M,0,0;
A00467:240:HCN7KDRX2:2:2275:20410:25786 339     chr9    9442086 0       2S124M  =       9442085 -125    TAGTCCCAGCTACTCGGGAGGCTGAGGCTGGAGGATCGCTTGAGTCCAGGAGTTCTGGGCTGTAGTGCGCTATGCCGATCGGGTGTCCGCACTAAGTTCGGCATCAATATGGTGACCTCCCGGGAG  FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF  NH
:i:7    HI:i:7  AS:i:213        nM:i:10 NM:i:0  MD:Z:124        RG:Z:c949090
A00467:240:HCN7KDRX2:2:2275:20410:25786 2225    chr12   26990   0       107H18M1S       chr14   49586604        0       GACCACCGGCACGCCCCCC     ,:F,:F,,,F,F,F,FF,F     NH:i:7  HI:i:2  AS:i:17 nM:i:0  NM:i:0  MD:Z:18 RG:Z:c949090    SA:Z:chr14,49586609,+,19S107M,0,10;
A00467:240:HCN7KDRX2:2:2275:20410:25786 83      chr14   49586604        0       126M    =       49586609        -121    TAGTCCCAGCTACTCGGGAGGCTGAGGCTGGAGGATCGCTTGAGTCCAGGAGTTCTGGGCTGTAGTGCGCTATGCCGATCGGGTGTCCGCACTAAGTTCGGCATCAATATGGTGACCTCCCGGGAG  FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
FFFFFFFFFFFF    NH:i:7  HI:i:2  AS:i:211        nM:i:10 NM:i:0  MD:Z:126        RG:Z:c949090
A00467:240:HCN7KDRX2:2:2275:20410:25786 339     chr14   49586604        0       126M    =       49586606        -124    TAGTCCCAGCTACTCGGGAGGCTGAGGCTGGAGGATCGCTTGAGTCCAGGAGTTCTGGGCTGTAGTGCGCTATGCCGATCGGGTGTCCGCACTAAGTTCGGCATCAATATGGTGACCTCCCGGGAG  FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
FFFFFFFFFFFF    NH:i:7  HI:i:1  AS:i:214        nM:i:10 NM:i:0  MD:Z:126        RG:Z:c949090
A00467:240:HCN7KDRX2:2:2275:20410:25786 339     chr14   49586604        0       126M    =       49586609        -121    TAGTCCCAGCTACTCGGGAGGCTGAGGCTGGAGGATCGCTTGAGTCCAGGAGTTCTGGGCTGTAGTGCGCTATGCCGATCGGGTGTCCGCACTAAGTTCGGCATCAATATGGTGACCTCCCGGGAG  FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
FFFFFFFFFFFF    NH:i:7  HI:i:3  AS:i:211        nM:i:10 NM:i:0  MD:Z:126        RG:Z:c949090
A00467:240:HCN7KDRX2:2:2275:20410:25786 339     chr14   49586604        0       126M    =       49586609        -121    TAGTCCCAGCTACTCGGGAGGCTGAGGCTGGAGGATCGCTTGAGTCCAGGAGTTCTGGGCTGTAGTGCGCTATGCCGATCGGGTGTCCGCACTAAGTTCGGCATCAATATGGTGACCTCCCGGGAG  FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
FFFFFFFFFFFF    NH:i:7  HI:i:4  AS:i:211        nM:i:10 NM:i:0  MD:Z:126        RG:Z:c949090
A00467:240:HCN7KDRX2:2:2275:20410:25786 339     chr14   49586604        0       126M    =       49586609        -121    TAGTCCCAGCTACTCGGGAGGCTGAGGCTGGAGGATCGCTTGAGTCCAGGAGTTCTGGGCTGTAGTGCGCTATGCCGATCGGGTGTCCGCACTAAGTTCGGCATCAATATGGTGACCTCCCGGGAG  FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
FFFFFFFFFFFF    NH:i:7  HI:i:5  AS:i:211        nM:i:10 NM:i:0  MD:Z:126        RG:Z:c949090
A00467:240:HCN7KDRX2:2:2275:20410:25786 339     chr14   49586604        0       126M    =       49586609        -121    TAGTCCCAGCTACTCGGGAGGCTGAGGCTGGAGGATCGCTTGAGTCCAGGAGTTCTGGGCTGTAGTGCGCTATGCCGATCGGGTGTCCGCACTAAGTTCGGCATCAATATGGTGACCTCCCGGGAG  FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
FFFFFFFFFFFF    NH:i:7  HI:i:6  AS:i:211        nM:i:10 NM:i:0  MD:Z:126        RG:Z:c949090
A00467:240:HCN7KDRX2:2:2275:20410:25786 419     chr14   49586606        0       16S110M =       49586604        124     GGGGGGCGTGCCGGTGGTCCCAGCAACACGGGCGGCTGAGGCGGGAGGATCGATGGAGTCAAGGAGTTCTGGGGTGTAGTGCGCTATGCCGATCGGGTGTCCGCACTAACTTCGGCATCAAAATGG  F,FF,F,F,F,,,F:,F:,,::,F,F,,F,::,,::,:,F,:,:::F:,,::,,,:F:::,::::::,:F,FF,:F,:F,F:::FF,:,:::,F,:,,:::FFFFF,,F,,,:,
:FFF:FF,F,,,    NH:i:7  HI:i:1  AS:i:214        nM:i:10 NM:i:10 MD:Z:8T2T4A9T9C1T5C12C35G11T4   RG:Z:c949090    SA:Z:chr1,46341588,-,110H16M,0,1;
A00467:240:HCN7KDRX2:2:2275:20410:25786 163     chr14   49586609        0       19S107M =       49586604        121     GGGGGGCGTGCCGGTGGTCCCAGCAACACGGGCGGCTGAGGCGGGAGGATCGATGGAGTCAAGGAGTTCTGGGGTGTAGTGCGCTATGCCGATCGGGTGTCCGCACTAACTTCGGCATCAAAATGG  F,FF,F,F,F,,,F:,F:,,::,F,F,,F,::,,::,:,F,:,:::F:,,::,,,:F:::,::::::,:F,FF,:F,:F,F:::FF,:,:::,F,:,,:::FFFFF,,F,,,:,
:FFF:FF,F,,,    NH:i:7  HI:i:2  AS:i:211        nM:i:10 NM:i:10 MD:Z:5T2T4A9T9C1T5C12C35G11T4   RG:Z:c949090    SA:Z:chr12,26990,-,107H18M1S,0,0;
A00467:240:HCN7KDRX2:2:2275:20410:25786 419     chr14   49586609        0       19S107M =       49586604        121     GGGGGGCGTGCCGGTGGTCCCAGCAACACGGGCGGCTGAGGCGGGAGGATCGATGGAGTCAAGGAGTTCTGGGGTGTAGTGCGCTATGCCGATCGGGTGTCCGCACTAACTTCGGCATCAAAATGG  F,FF,F,F,F,,,F:,F:,,::,F,F,,F,::,,::,:,F,:,:::F:,,::,,,:F:::,::::::,:F,FF,:F,:F,F:::FF,:,:::,F,:,,:::FFFFF,,F,,,:,
:FFF:FF,F,,,    NH:i:7  HI:i:3  AS:i:211        nM:i:10 NM:i:10 MD:Z:5T2T4A9T9C1T5C12C35G11T4   RG:Z:c949090    SA:Z:chr1,195452,-,107H18M1S,0,0;
A00467:240:HCN7KDRX2:2:2275:20410:25786 419     chr14   49586609        0       19S107M =       49586604        121     GGGGGGCGTGCCGGTGGTCCCAGCAACACGGGCGGCTGAGGCGGGAGGATCGATGGAGTCAAGGAGTTCTGGGGTGTAGTGCGCTATGCCGATCGGGTGTCCGCACTAACTTCGGCATCAAAATGG  F,FF,F,F,F,,,F:,F:,,::,F,F,,F,::,,::,:,F,:,:::F:,,::,,,:F:::,::::::,:F,FF,:F,:F,F:::FF,:,:::,F,:,,:::FFFFF,,F,,,:,
:FFF:FF,F,,,    NH:i:7  HI:i:4  AS:i:211        nM:i:10 NM:i:10 MD:Z:5T2T4A9T9C1T5C12C35G11T4   RG:Z:c949090    SA:Z:chr1,24927,-,107H18M1S,0,0;
A00467:240:HCN7KDRX2:2:2275:20410:25786 419     chr14   49586609        0       19S107M =       49586604        121     GGGGGGCGTGCCGGTGGTCCCAGCAACACGGGCGGCTGAGGCGGGAGGATCGATGGAGTCAAGGAGTTCTGGGGTGTAGTGCGCTATGCCGATCGGGTGTCCGCACTAACTTCGGCATCAAAATGG  F,FF,F,F,F,,,F:,F:,,::,F,F,,F,::,,::,:,F,:,:::F:,,::,,,:F:::,::::::,:F,FF,:F,:F,F:::FF,:,:::,F,:,,:::FFFFF,,F,,,:,
:FFF:FF,F,,,    NH:i:7  HI:i:5  AS:i:211        nM:i:10 NM:i:10 MD:Z:5T2T4A9T9C1T5C12C35G11T4   RG:Z:c949090    SA:Z:chr2,113588497,+,1S18M107H,0,0;
A00467:240:HCN7KDRX2:2:2275:20410:25786 419     chr14   49586609        0       19S107M =       49586604        121     GGGGGGCGTGCCGGTGGTCCCAGCAACACGGGCGGCTGAGGCGGGAGGATCGATGGAGTCAAGGAGTTCTGGGGTGTAGTGCGCTATGCCGATCGGGTGTCCGCACTAACTTCGGCATCAAAATGG  F,FF,F,F,F,,,F:,F:,,::,F,F,,F,::,,::,:,F,:,:::F:,,::,,,:F:::,::::::,:F,FF,:F,:F,F:::FF,:,:::,F,:,,:::FFFFF,,F,,,:,
:FFF:FF,F,,,    NH:i:7  HI:i:6  AS:i:211        nM:i:10 NM:i:10 MD:Z:5T2T4A9T9C1T5C12C35G11T4   RG:Z:c949090    SA:Z:chr19,66535,-,107H18M1S,0,0;
A00467:240:HCN7KDRX2:2:2275:20410:25786 2481    chr19   66535   0       107H18M1S       chr14   49586604        0       GACCACCGGCACGCCCCCC     ,:F,:F,,,F,F,F,FF,F     NH:i:7  HI:i:6  AS:i:17 nM:i:0  NM:i:0  MD:Z:18 RG:Z:c949090    SA:Z:chr14,49586609,+,19S107M,0,10;
```

This is where Arriba specifies its STAR parameters: https://github.com/suhrig/arriba/blob/786f647a73b8ffb8bbe38b6607460c27e17e8846/run_arriba.sh#L30-L36

tools/star.wdl
adthrasher: For `tMN` with our existing pipeline, there are an average of `260382` reads per sample that are mapped to too many loci. That's between `0.11%` and `0.96%` of reads.

adthrasher: If I expand to all available STAR logs (`/research_jude/rgs01_jude/groups/zhanggrp/projects/SJCloud/common/plaid/work/*/*/TRANSCRIPTOME/*/*/results/*.final.out`), then the average is `338989` with a median of `145909`.

tools/star.wdl
```diff
@@ -585,20 +585,20 @@ task alignment {
         Int alignWindowsPerReadNmax = 10000
         Int alignTranscriptsPerWindowNmax = 100
         Int alignTranscriptsPerReadNmax = 10000
-        Int peOverlapNbasesMin = 0
+        Int peOverlapNbasesMin = 10
```
claymcleod, line 588: This purports to improve accuracy for paired end calling by turning on the "§ 10. Merging and mapping of overlapping paired-end reads." Reading over the algorithm, it seems fine to include in RNA-Seq v2 workflows, though I still might stick with the defaults.

tools/star.wdl
```diff
@@ -585,20 +585,20 @@ task alignment {
...
         Int winFlankNbins = 4
-        Int chimSegmentMin = 0
+        Int chimSegmentMin = 10
         Int chimScoreMin = 0
-        Int chimScoreDropMax = 20
-        Int chimScoreSeparation = 10
-        Int chimScoreJunctionNonGTAG = -1
-        Int chimJunctionOverhangMin = 20
-        Int chimSegmentReadGapMax = 0
+        Int chimScoreDropMax = 30
```
claymcleod, line 595: I stopped reviewing here in the list. My main thought is this: should we stick to the STAR defaults for the WDL task defaults and then only break these out for our Arriba-related workflows? Let me know what you think of that.

tools/star.wdl
claymcleod: Okay, I see. Since we are balancing all of these concerns based on the purpose, see my general comment below on whether the defaults can match STAR defaults for this task but then we can override them for Arriba-based workflows.

tools/star.wdl
claymcleod: Okay, so it disproportionately affects some samples. Interesting.

tools/star.wdl
adthrasher: I think that's reasonable. The downside, and this is a failing of WDL in my opinion, is that overriding at the workflow level is cumbersome. If you override in the `call` block, then the user is blocked from changing the value, even if the workflow otherwise enables `allowNestedInputs`. So you're forced in to explicitly declaring `inputs` at the workflow level for these STAR parameters. We can do that, but it is significantly more cluttered.

tools/star.wdl
```diff
@@ -590,7 +590,7 @@ task alignment {
         Int winBinNbits = 16
         Int winAnchorDistNbins = 9
         Int winFlankNbins = 4
-        Int chimSegmentMin = 0
+        Int chimSegmentMin = 10
```
adthrasher, line 593: @claymcleod and @a-frantz 

I've migrated all of the other overrides up to `rnaseq-core`. I've left these two as `chimSegmentMin` must be > 0 to enable chimeric detection, so I suspect we meant to enable that originally. For `chimOutType`, `WithinBAM` is recommended by the STAR manual (even though `Junctions` is the default).

a-frantz: Just curious, why `10`?

IMO `chimOutType WithinBAM` is the right call. I think it's better to output chimeric alignments by default than to exclude them. That would necessitate having a `chimSegmentMin > 0`. (Weird set-up from STAR.) I'm just not sure where `10` came from.

adthrasher: Valid question. I took `10` from Arriba. I don't know if that is a completely sane choice. I don't think allowing a chimeric segment to be too short (e.g. `1`bp) is wise. Below a certain value, you're not getting any meaningful information and likely just spurious alignments.

workflows/rnaseq/rnaseq-core.wdl
```diff
@@ -55,18 +55,32 @@ workflow rnaseq_core {
...
-        Array[File] read_one_fastqs_gz
-        Array[File] read_two_fastqs_gz
         File gtf
         File star_db
+        Array[File] read_one_fastqs_gz
+        Array[File] read_two_fastqs_gz
         String read_groups
         String prefix
         File? contaminant_db
+        SJ_Motifs alignSJstitchMismatchNmax = SJ_Motifs {
```
a-frantz, line 65: 2 small changes I'd like to see happen here (in `rnaseq-core`)

1. I can be persuaded that this is overkill, but I would prefix all these new params with `star_`. I'd just like it to be clear that these parameters are specific to the STAR alignment task.
2. we're missing parameter meta for all of these. I think a copy+paste from STAR is fine, with two small tweaks.
    - The first sentence needs to provide context by stating that the param impacts STAR alignment
    - please add a `tool_default:` key that makes it clear _this default_ is overriding the _tool/task/STAR default_. IDK if "tool"_default is correct term there.

adthrasher: Based on Slack conversation, we decided to skip the `star_` prefix, but add a `tool: star` meta entry to make it clear what the underlying tool is.

tools/arriba.wdl
```diff
@@ -0,0 +1,374 @@
+## [Homepage](https://arriba.readthedocs.io/en/latest/)
+version 1.1
```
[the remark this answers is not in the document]
adthrasher: No idea how I missed that...

tools/arriba.wdl
```diff
@@ -0,0 +1,374 @@
...
+            description: "Strandedness of the input data."
+            external_help: "https://arriba.readthedocs.io/en/latest/command-line-options/"
+            choices: [
+                "auto",
+                "yes",
+                "no",
+                "reverse"
+            ]
+        }
+        mark_duplicates: "Mark duplicates in the input BAM file with Arriba."
```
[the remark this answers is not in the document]
adthrasher:
```
-u : Arriba performs marking of duplicates internally based on identical mapping coordinates. When this switch is set, internal marking of duplicates is disabled and Arriba assumes that duplicates have been marked by a preceding program. In this case, Arriba only discards alignments flagged with the BAM_FDUP flag. This makes sense when duplicates cannot be reliably identified solely based on their mapping coordinates, e.g. when unique molecular identifiers (UMIs) are used or when independently generated libraries are merged in a single BAM file and the read group must be interrogated to distinguish duplicates from reads that map to the same coordinates by chance. In addition, when this switch is set, duplicate reads are not considered for the calculation of the coverage at fusion breakpoints (columns coverage1 and coverage2 in the output file).
```

If I'm reading that correctly, it should be the same method as Picard/samtools/etc.

tools/arriba.wdl
```diff
@@ -0,0 +1,374 @@
...
+                'isoforms',
+                'intronic',
+                'in_vitro',
+                'intragenic_exonic',
+                'internal_tandem_duplication'
+            ]
+        }
+        feature_name: {
+            description: "List of feature names to use in GTF."
+            help: "The Arriba default it designed to handle RefSeq, GENCODE, or ENSEMBL format annotations. `feature_name` expects a string of space separated options. The required fields are `gene_name`, `gene_id`, `transcript_id`, `feature_exon`, and `feature_CDS`. The fields should space separated. The values should be provided with `field=value`. Mutliple values can be provided and separated by a pipe (`|`), e.g. `=value1|value2`. A complete example is `gene_name=gene_name|gene_id gene_id=gene_id transcript_id=transcript_id feature_exon=exon feature_CDS=CDS`."
```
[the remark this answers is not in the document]
adthrasher: No idea why I wrote space separated. Arriba accepts both space or comma separated for this field.

tools/arriba.wdl
```diff
@@ -0,0 +1,374 @@
...
+        quantile: "Genes with expression above the given quantile are eligible for filtering."
+        exonic_fraction: "Minimum fraction of exonic sequence between breakpoints."
+        coverage_fraction: "Minimum fraction of viral contig transcription."
+        min_itd_allele_fraction: "Minimum supporting read fraction for internal tandem duplications."
+        max_genomic_breakpoint_distance: "With 'wgs_svs', threshold for relating genomic and transcriptomic events."
+        min_supporting_reads: "Minimum number of supporting reads for a fusion."
+        homopolymer_length: "Maximum homopolymer length adjacent to breakpoints."
+        read_through_distance: "Minimum distance between breakpoints for read-through events."
+        min_anchor_length: "Minimum anchor length for split reads."
+        many_spliced_events: "Recover fusions with at least this many spliced breakpoints."
```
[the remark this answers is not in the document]
adthrasher: Should we do something explicit like `common: false`? Not that different to what we're doign now, but maybe makes it clearer that these probably shouldn't be changed.

workflows/rnaseq/rnaseq-core.wdl
```diff
@@ -32,6 +32,16 @@ workflow rnaseq_core {
...
         prefix: "Prefix for output files"
         contaminant_db: "A compressed reference database corresponding to the aligner chosen with `xenocp_aligner` for the contaminant genome"
+        alignSJstitchMismatchNmax: {
+            description: "This overrides the STAR alignment default. Maximum number of mismatches for stitching of the splice junctions (-1: no limit) for: (1) non-canonical motifs, (2) GT/AG and CT/AC motif, (3) GC/AG and CT/GC motif, (4) AT/AC and GT/AT motif",
```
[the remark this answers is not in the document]
adthrasher: I originally had it in there, but with `tool: "star"` I took it out.

tools/arriba.wdl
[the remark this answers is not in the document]
a-frantz: I thought about doing `common: false`, but that is potentially confusing. There would be 3 cases, `common: true`, `common: false`, and the most frequent case where there is no `common` entry. To me, it only makes sense that `null` case should be considered equivalent with the `false` case. I don't want to get fall into the `T/F/Null` trap.

I think it's the same with the `deprecated` key we introduced. In the absence of a `deprecated` key, it's safe to assume `deprecated: false`.

workflows/rnaseq/rnaseq-core.wdl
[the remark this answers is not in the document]
a-frantz: Maybe it's overkill 🤷 but it should be consistent one way or the other

tools/arriba.wdl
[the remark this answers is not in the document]
adthrasher: To me, the absence of `common: true` is an indication that most people shouldn't need to mess with the parameters. They're also sorted to the bottom of the list. I'm concerned that if we move toward a community resource model, where we need to expose all tool parameters, we're going to end up with a ton of these types of situations. It's a lot to read and maintain as well as requiring some consensus on what `advanced` would be.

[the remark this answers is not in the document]
a-frantz: Let's view this through the lens of `sprocket doc`.

* `common: true` [will] mean put this right under the required parameters and highlight it
* `common: false || NULL` means no special treatment

Are those two cases enough?
Going off memory here, but our current convention is to only start labelling params as `common` once a task has >10-or-so parameters. Should we revisit that and start labelling `common` more frequently?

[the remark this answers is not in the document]
a-frantz: I'm fine with only using `common` for now. It's easy enough to add other labels later if we see the need. But if we're going to keep with the one-label/two-conditions situation, I'd like to see us more consistently use the `true` condition. Doesn't have to hold up this PR, but something for us to keep in mind for the future.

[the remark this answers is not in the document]
adthrasher: I just thought of this, but what if we added an additional level/object for `sprocket`?
```
        chimSegmentMin: {
            description: "minimum length of chimeric segment length, if ==0, no chimeric output",
            tool_default: 0,
            sprocket: {
                 category: common
            }
        }
```
I could immediately see `common` and `advanced` for `category` choices. Are there other things we might want to explicitly write for `sprocket`?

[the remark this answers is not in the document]
adthrasher: From Slack, we decided to continue with just `common: true` for now. We may refactor to 
```
 sprocket: {
    category: common
}
```
in the future. This would be a controlled vocabulary. We would also want additional use cases for the `sprocket` object before proceeding.

tools/star.wdl
```diff
@@ -553,7 +555,7 @@ task alignment {
         Int out_sam_flag_OR = 0
         Int out_sam_flag_AND = 65535
         Int out_filter_multimap_score_range = 1
-        Int out_filter_multimap_n_max = 20
+        Int out_filter_multimap_n_max = 10
```
adthrasher, line 558: `10` is the default for 2.7.11.b. Not sure why it was `20` previously.

tools/star.wdl
```diff
@@ -588,7 +590,7 @@ task alignment {
         Int win_bin_n_bits = 16
         Int win_anchor_dist_n_bins = 9
         Int win_flank_n_bins = 4
-        Int chim_segment_min = 0
+        Int chim_segment_min = 10
```
adthrasher, line 593: This enables chimeric detection, so it belongs in `star` itself.

workflows/rnaseq/rnaseq-core.wdl
```diff
@@ -30,6 +30,16 @@ workflow rnaseq_core {
...
         prefix: "Prefix for output files"
         contaminant_db: "A compressed reference database corresponding to the aligner chosen with `xenocp_aligner` for the contaminant genome"
+        align_sj_stitch_mismatch_n_max: {
+            description: "This overrides the STAR alignment default. Maximum number of mismatches for stitching of the splice junctions (-1: no limit) for: (1) non-canonical motifs, (2) GT/AG and CT/AC motif, (3) GC/AG and CT/GC motif, (4) AT/AC and GT/AT motif",
+            tool: "star",
+            tool_default: {
+                noncanonical_motifs: 0,
+                GT_AG_and_CT_AC_motif: -1,
```
adthrasher, line 38: @a-frantz  - Sprocket also doesn't like this line because it doesn't understand the `-` thinks this is a string. I don't really want to quote it, but should we to get sprocket to pass and avoid missing other issues?

a-frantz: Don't really like any option I see here, but I think the best we can do is quote the `-1` and leave a comment to the effect of `# TODO: remove quotes once sprocket parses negative numbers`

tools/arriba.wdl
```diff
@@ -0,0 +1,379 @@
...
+        File fusions
+        File gtf
+        String prefix = basename(fusions, ".tsv")
+        Int modify_disk_size_gb = 0
+    }
+
+    Int input_size_gb = ceil(size(gtf, "GiB"))
+    Int disk_size_gb = ceil(input_size_gb) + 5 + modify_disk_size_gb
+
+    command <<<
```
a-frantz, line 358:
```suggestion
    command <<<
        set -euo pipefail

```

a-frantz (dismissed):
GitHub is acting up and not letting me approve but I am trying to approve.

a-frantz (approved):

PR #138, opened 2024-03-08
feat: don't MarkDuplicates on DNA

a-frantz:
This started as a PR for replacing `picard MarkDuplicates` with `samtools markdup`. Turns out samtools' implementation is prohibitively memory hungry. (Tested allocating 3X bam size in memory and still had failures.) But while doing a bit of a deep dive on the samtools/picard MD implementations, I discovered that the algorithms' distinction between optical and non-optical duplicates is built on shaky ground (IMO at least). I don't think it's a sound analysis for us to be doing in QC. And to my knowledge the portion of optical VS non-optical duplicates has never been used by us during QC, so it should be safe to drop.

Cool side effect, it's going to save a chunk of change on running DNA samples through QC, as MD took a long time with a decent memory allocation (50gb).

Duplicate counts can still be discovered by samtools flagstat output.


tools/ngsderive.wdl
```diff
@@ -69,10 +69,10 @@ task strandedness {
             "$CWD_BAM" \
             > ~{outfile_name}
 
-        if split_by_rg; then
-            awk 'NR > 1' ~{outfile_name} | grep 'overall' | cut -f6 > strandedness.txt # TODO broken until ngsderive v4 is released
+        if ~{split_by_rg}; then
+            echo "N/A" > strandedness.txt
```
adthrasher, line 73: Does `ngs strandedness` not output an overall call when running in per-RG mode?

tools/picard.wdl
```diff
@@ -14,31 +14,60 @@ task mark_duplicates {
...
+                "OpticalOnly",
+                "All"
+            ],
+        }
         create_bam: {
             description: "Enable BAM creation (true)? Or only output MarkDuplicates metrics (false)?",
             common: true
         }
-        memory_gb: "RAM to allocate for task, specified in GB"
+        clear_dt: "Clear the `DT` tag from the input BAM? For increased performance if the input BAM does not have the `DT` tag set to `false`."
```
adthrasher, line 46:
```suggestion
        clear_dt: "Clear the `DT` tag from the input BAM? For increased performance, if the input BAM does not have the `DT` tag, set to `false`."
```

tools/samtools.wdl
```diff
@@ -1006,3 +1018,224 @@ task fixmate {
         maxRetries: 1
     }
 }
+
+task position_sorted_fixmate {
```
adthrasher, line 1022: I hesitate to add this to the repo as a task. I'm guessing that the intent is to avoid creating a number of extra intermediate outputs, but it seems like this should be a workflow that can be included as a subworkflow to get this behavior.

tools/samtools.wdl
```diff
@@ -1006,3 +1018,224 @@ task fixmate {
...
+        disk: "~{disk_size_gb} GB"
+        container: 'quay.io/biocontainers/samtools:1.19.2--h50ea8bc_0'
+        maxRetries: 1
+    }
+}
+
+task markdup {
+    meta {
+        description: "**[DEPRECATED]** Runs `samtools markdup` on the position-sorted input BAM file. This creates a report and optionally a new BAM with duplicate reads marked."
+        help: "This task assumes `samtools fixmate` has already been run on the input BAM. If it has not, then the output may be incorrect. Running `fixmate` may not be necessary, if the aligner used correctly sets the mate information. This includes but is not limited to having correct insert sizes, mate coordinates, `ms` or \"mate score\" tags, and `MC` or \"mate cigar\" tags. A name-sorted or collated BAM can be run through the `fixmate` task (and then position sorted prior to this task) or a position-sorted BAM can be run through the `position_sorted_fixmate` task. Deprecated due to extremely high memory usage for certain RNA-Seq samples when searching for optical duplicates. Use `mark_duplicates` in `./picard.wdl` instead."
```
adthrasher, line 1126:
```suggestion
        help: "This task assumes `samtools fixmate` has already been run on the input BAM. If it has not, then the output may be incorrect. Running `fixmate` may not be necessary, if the aligner correctly sets the mate information. This includes but is not limited to having correct insert sizes, mate coordinates, `ms` or \"mate score\" tags, and `MC` or \"mate cigar\" tags. A name-sorted or collated BAM can be run through the `fixmate` task (and then position-sorted prior to this task) or a position-sorted BAM can be run through the `position_sorted_fixmate` task. Deprecated due to extremely high memory usage for certain RNA-Seq samples when searching for optical duplicates. Use `mark_duplicates` in `./picard.wdl` instead."
```

tests/data_structures/test_flag_filter.yaml
```diff
@@ -3,58 +3,58 @@
     - miniwdl
     - flag_filter
   command: >-
-    miniwdl run -d test-output/. --task validate_string_is_12bit_oct_dec_or_hex data_structures/flag_filter.wdl number="0x900"
+    miniwdl run --verbose -d test-output/. --task validate_string_is_12bit_oct_dec_or_hex data_structures/flag_filter.wdl number="0x900"
```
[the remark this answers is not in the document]
a-frantz: If a task fails in the CI, `--verbose` will print the logs to the GitHub interface. If `--verbose` isn't there, the only way to access those logs is by running the tests locally (or rather IDK another way). So it's useful when things fail.

tools/ngsderive.wdl
a-frantz: Good question! It did not, until I released v4. As you'll see, in WDL we're stuck on v3. This is because there is an issue with the conda build preventing v4 from being released (and by extension, it isn't released on BioContainers either).

I've spent a decent amount of time trying to get to the bottom of the conda build error without luck. I pinged the conda team asking them to look at it, and someone _did_ take a look. They just didn't offer any helpful advice and I gave up TBH. And that was quite a long time ago now.

TBH I expected I could cut the Rust port of `ngsderive` sooner than has materialized. IMO the v4 release isn't important enough to deal with that headache **in light of the Rust port releasing soon.** The PR is in Clay's hands for review. The plan is to archive the entire `ngsderive` repo once that port is merged. So no sense in spending time on addressing that when it's on the verge of archival.

tools/samtools.wdl
a-frantz: Is the fear that this opens the door for having a name-sorted **and** position-sorted version of each task? Because I had not thought of that, and it is a scary idea.

> I'm guessing that the intent is to avoid creating a number of extra intermediate outputs,

Yes. Specifically I wanted to avoid putting a collated BAM on disk. I don't have the numbers in front of me, but collation can cause some astonishingly huge BAMs in the worst case. (And no, I have not figured out how to predict what causes the worst case.)

Flashback to when we first swapped from running `picard SamToFastq` to running `samtools collate` then `samtools fastq` **as separate tasks**. Having collated WGS BAMs for each sample hogged _all_ the disk space wherever I was running QC and the LSF team was upset.

My point is that I think we need to think carefully about how to prevent massive intermediate file build-up.

>  it seems like this should be a workflow that can be included as a subworkflow to get this behavior.

Can you elaborate on this idea? I'm not following, unless you're suggesting we just eat the intermediate file cost.

Good news: we don't need this task. I kept us using `picard MarkDuplicates`, which means this is not called by any of our workflows. We could just delete it. I'm not convinced that's the _best_ choice, but it is a choice.

🤔 🤔 🤔

tools/samtools.wdl
a-frantz: Thank you for bringing my attention back to this. One part here is incorrect:
> Running `fixmate` may not be necessary, if the aligner correctly sets the mate information. This includes but is not limited to having correct insert sizes, mate coordinates, `ms` or \"mate score\" tags,...

No aligner is going to set the `ms` tag. It's not an officially supported tag, it's custom for `fixmate+markdup`. So I'll remove those 2 sentences. `fixmate` _must_ be run. Which is a real pain and downside of the samtools implementation. But that's a tangent.

adthrasher (approved):

PR #140, opened 2024-03-26
fix(tools/picard): make VALIDATION_STRINGENCY SILENT the default

a-frantz:
^Tiitle

Brought to my attention by a user with files that were failing during Picard steps due to a MAPQ>0 for unmapped reads. This should fix that.

### DIFF
--- a/tools/picard.wdl
+++ b/tools/picard.wdl
@@ -150,6 +150,15 @@ task validate_bam {
             common: true
         }
         outfile_name: "Name for the ValidateSamFile report file"
+        validation_stringency: {
+            description: "Validation stringency for parsing the input BAM.",
+            choices: [
+                "STRICT",
+                "LENIENT",
+                "SILENT"
+            ],
+            tool_default: "STRICT",
+        }
         succeed_on_errors: {
             description: "Succeed the task even if errors *and/or* warnings are detected",
             common: true
@@ -173,6 +182,7 @@ task validate_bam {
         File? reference_fasta
         Array[String] ignore_list = []
         String outfile_name = basename(bam, ".bam") + ".ValidateSamFile.txt"
+        String validation_stringency = "SILENT"
         Boolean succeed_on_errors = false
         Boolean succeed_on_warnings = true
         Boolean summary_mode = false
@@ -203,6 +213,7 @@ task validate_bam {
             ~{reference_arg} \
             ~{mode_arg} \
             ~{stringency_arg} \
+            --VALIDATION_STRINGENCY ~{validation_stringency} \
             ~{sep(" ", prefix("--IGNORE ", ignore_list))} \
             --MAX_OUTPUT ~{max_errors} \
             > ~{outfile_name} \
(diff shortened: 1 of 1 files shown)


tools/picard.wdl
```diff
@@ -283,8 +314,7 @@ task sort {
             -SO ~{sort_order} \
             --CREATE_INDEX true \
             --CREATE_MD5_FILE true \
-            --COMPRESSION_LEVEL 5 \
```
a-frantz, line 286: @adthrasher maybe controversial to remove this? IDK why it's there in the first place which is why I removed it. If there's a reasonable justification we can put it back _with a comment_ (and add this line to other BAM-creating tasks?).

adthrasher: `5` is the default. I'd prefer to parameterize this (eventually) along with other arguments (`CREATE_INDEX`, `CREATE_MD5_FILE`).

tools/picard.wdl
```diff
@@ -193,6 +213,7 @@ task validate_bam {
             ~{reference_arg} \
             ~{mode_arg} \
             ~{stringency_arg} \
+            --VALIDATION_STRINGENCY ~{validation_stringency} \
```
adthrasher, line 216: This is probably the one task that `SILENT` shouldn't be the default.

tools/picard.wdl
a-frantz: We don't need to settle this now, but I think I disagree with at least the back-half of your comment.

It boils down to what our philosophy for the repo is. Are we just exposing "bare metal" or are we including some padding? Are there parameters we can hide so the user doesn't have to think about it?

We're inconsistent in this repository. Some tasks we write pretty "minimally", in that we're really just writing the thinnest WDL wrapper around a CLI that we can. Other tools we've decided we don't like certain aspects of the CLI and have written WDL to obfuscate the "raw" CLI. I think both approaches are fine and have their place. It's not a matter of only using one or the other.

With that flexibility in approach, I think it's reasonable that if we're creating a BAM we might as well index it and calculate an MD5. Both operations might as well be free in comparison to the sort step. I would consider adding BAI and MD5 creation to every step that creates a BAM. It's a bit wasteful spinning up a new task just for BAI _AND/OR_ MD5 creation when that could be handled in the BAM creation task.

Food for thought.

adthrasher: I think it depends. I was envisioning a system that allows streaming output. In that case, there's not need to create an index or MD5, you could stream uncompressed, sorted BAM to a downstream consumer which would save a ton of time on compression and decompression. Streaming would also save the disk storage.

I think what we currently have should be the default, but I have no issue exposing parameters if we've set a reasonable default. For most BAM creation tasks, doing to the MD5 and indexing alongside is probably the right default setting. I just think that we should expose some of that as optional for users to disable in specific cases or environments.

adthrasher (dismissed):

tools/picard.wdl
```diff
@@ -150,15 +150,6 @@ task validate_bam {
             common: true
         }
         outfile_name: "Name for the ValidateSamFile report file"
```
adthrasher, line 152: I was thinking we'd leave the options exposed, but change the default to `STRICT`.

a-frantz: @adthrasher I just added the option back, _but_ I set the default to `LENIENT`. I think the most reasonable default behavior for a Validate task is to *log every potential problem but continue whenever possible*. I also don't think the specifics of this param are significant in the vast majority of cases.

adthrasher (approved):

PR #142, opened 2024-04-04
feat(QC): adds optional optical dup marking

a-frantz:


workflows/qc/quality-check-standard.wdl
```diff
@@ -92,8 +96,10 @@ workflow quality_check {
...
         use_all_cores: "Use all cores? Recommended for cloud environments."
+        optical_distance: "Maximum distance between read coordinates to consider them optical duplicates instead of library duplicates (e.g. PCR duplicates). If `mark_duplicates == false`, this parameter is ignored. If `0`, then _optical_ duplicate marking is disabled and only traditional duplicate marking will be performed. Suggested settings of 100 for unpatterned versions of the Illumina platform (e.g. HiSeq) or 2500 for patterned flowcell models (e.g. NovaSeq). Calculation of distance depends on coordinate data embedded in the read names, typically produced by Illumina sequencing machines. Optical duplicate detection will not work on non-standard names without a custom regex for tile-data extraction. Review the `mark_duplicates` task in `../../tools/picard.wdl` for more information."
```
adthrasher, line 102:
```suggestion
        optical_distance: "Maximum distance between read coordinates to consider them optical duplicates instead of technical duplicates (e.g. PCR duplicates). If `mark_duplicates == false`, this parameter is ignored. If `0`, then _optical_ duplicate marking is disabled and only traditional duplicate marking will be performed. Suggested settings of 100 for unpatterned versions of the Illumina platform (e.g. HiSeq) or 2500 for patterned flowcell models (e.g. NovaSeq). Calculation of distance depends on coordinate data embedded in the read names, typically produced by Illumina sequencing machines. Optical duplicate detection will not work on non-standard names without a custom regex for tile-data extraction. Review the `mark_duplicates` task in `../../tools/picard.wdl` for more information."
```

Maybe this language comes from Picard, but I typically hear PCR duplicates referenced as "technical duplicates". So we may want to change that, but not if what you wrote is consistent with Picard.

adthrasher (approved):

PR #143, opened 2024-04-12
feat: multiqc configs

a-frantz:
A bunch of changes inspired by conversations with David F.

------------------
Eventually we want to add a bunch of lines similar to:
```
custom_plot_config:
  mosdepth-cumcoverage-dist-id:
    title: "Cumulative coverage distribution (whole_genome)"
  mosdepth-cumcoverage-dist-id-1:
    title: "Cumulative coverage distribution (exon)"
  mosdepth-cumcoverage-dist-id-2:
    title: "Cumulative coverage distribution (CDS)"
  mosdepth-cumcoverage-dist-id-3:
    title: "Cumulative coverage distribution (UTR)"
  mosdepth-cumcoverage-dist-id-4:
    title: "Cumulative coverage distribution (whole_genome, markdups)"
  mosdepth-cumcoverage-dist-id-5:
    title: "Cumulative coverage distribution (exon, markdups)"
  mosdepth-cumcoverage-dist-id-6:
    title: "Cumulative coverage distribution (CDS, markdups)"
  mosdepth-cumcoverage-dist-id-7:
    title: "Cumulative coverage distribution (UTR, markdups)"
```
However, this config option is currently bugged in MultiQC.


tools/util.wdl
```diff
@@ -370,47 +370,55 @@ task unpack_tarball {
     }
 }
 
-task make_coverage_regions_beds {
-    # TODO make this customizable
+task make_coverage_regions_bed {
```
adthrasher, line 373: This is certainly out-of-scope, but I was recently extracting `protein_coding` genes from an annotation file. Should we consider generalizing this in the future to extract any given set of features from an annotation GTF file?

adthrasher (approved):

adthrasher (approved):

tools/util.wdl
a-frantz: So this PR "generalized" the extraction to enable pulling out any specific values from the `feature_type` column. If I understand the request, you want to control which column the match is performed on as well as the value matched on? That _should_ be straightforward. We'd want to restrict to columns that make sense to compare on (e.g. _not_ the `start`/`end` columns), but that should be pretty easy.

adthrasher: Right, I was thinking of a more general `extract_features` type task. Much of that logic is here though, so I wouldn't necessarily want to duplicate it. However as written, since it converts to BED, you lose the other annotations. So if you extract `gene` features, you can't subsequently refine to only `protein_coding` genes.

adthrasher: We'd have to parse the annotation field to find the `gene_type` field (for GENCODE). That may not be portable though, so maybe it wouldn't be worth it.

PR #147, opened 2024-04-30
Reorganize ReadGroup struct

adthrasher:
Move the `ReadGroup` struct into a `data_structures` file. Update the workflows to use it. Add tests. Add method to validate a `ReadGroup` struct has fields that a task requires. Also check to ensure that the fields are spec-compliant.


adthrasher:
This is pending the changes in https://github.com/chanzuckerberg/miniwdl/issues/686

data_structures/read_group.wdl
```diff
@@ -0,0 +1,386 @@
...
+        container: 'ghcr.io/stjudecloud/util:1.4.0'
+        maxRetries: 1
+    }
+}
+
+task get_ReadGroups {
+    meta {
+        description: "Gets read group information from a BAM file and writes it out as JSON which is converted to a WDL struct."
+        outputs: {
+            read_groups: "An array of strings containing read group information. If `format_for_star = true`, all found read groups are contained in one string (`read_groups[0]`). If `format_for_star = false`, each found @RG line will be its own entry in output array `read_groups`."
```
a-frantz, line 73: Please update to current behavior

data_structures/read_group.wdl
```diff
@@ -0,0 +1,386 @@
...
+        ReadGroup read_group
+        Array[String] required_fields = ["ID"]
+        Boolean restrictive = true
+    }
+
+    # The SAM spec allows any printable ASCII character in header fields.
+    String sam_spec_pattern = "[\\ -~]+"
+    # We have the opinion that is too permissive for ID and SM.
+    String id_pattern = if restrictive then "id" else sam_spec_pattern
+    String sample_pattern = if restrictive then "sample.?" else sam_spec_pattern
```
a-frantz, line 140: I'm confused by what these patterns are meant to be. Doesn't this say the only acceptable `ID` value is the literal `id`? And Similarly that all `SM` values must start with the literal `sample`? That doesn't seem like what you intended.

data_structures/read_group.wdl
```diff
@@ -0,0 +1,386 @@
...
+            fi
+        fi
+        if [ "~{defined(read_group.ID)}" == "true" ]
+        then
+            if [[ "~{read_group.ID}" =~ ^~{id_pattern}$ ]]
+            then
+                >&2 echo "ID must match pattern ~{id_pattern}"
+                error=1
+            fi
+        fi
```
a-frantz, line 160: IMO it's confusing here to act as if there's a situation where `ID` isn't required. I'd prefer to just directly test `[[ "~{read_group.ID}" =~ ^~{id_pattern}$ ]]` and drop all the other if clauses.

data_structures/read_group.wdl
```diff
@@ -0,0 +1,386 @@
...
+        String check = "passed"
+    }
+
+    runtime {
+        memory: "4 GB"
+        disk: "10 GB"
+        container: 'ghcr.io/stjudecloud/util:1.4.0'
+        maxRetries: 0
+    }
+}
```
a-frantz, line 386:
```suggestion
}

```

data_structures/read_group.wdl
```diff
@@ -0,0 +1,386 @@
+version 1.1
+
+# See the `read_groups` `parameter_meta` for definitions of each field
```
a-frantz, line 3: I think the `read_group` field definitions need to live within this file. Probably as header comments. They are currently defined in `rnaseq-standard-fastq`, which should instead point to this file "for more information".

a-frantz (changes requested):

data_structures/read_group.wdl
adthrasher: Below I test with 
```if [[ "~{read_group.SM}" =~ ^~{sample_pattern}$ ]]```
So for the sample test it tests for exactly the string `sample` plus one optional character. So `sample`, `sample1`, etc. would be excluded, but not `sample_1`. We could get overly complicated, but I was attempting to write a basic check that we could enhance over time if necessary, without tripping up too many users.

File: data_structures/read_group.wdl (hunk 5 above)
Comment (adthrasher) [reply data_structures/read_group.wdl:30]: I've brought them over as a header comment and also documented inline in the `struct` definition. That might be overkill though.

Comment (a-frantz) [reply data_structures/read_group.wdl:30]: I gather it's the exact same information being documented in both places? That sounds like a mistake of only updating one is bound to happen in the future. I think just the header comments. Maybe a comment above the struct definition saying "see header"?

File: data_structures/read_group.wdl (hunk 2 above)
Comment (a-frantz) [reply data_structures/read_group.wdl:167]: That makes more sense but I'm still confused. Don't all these checks need to be negated?
```
        if [[ "~{read_group.ID}" =~ ^~{id_pattern}$ ]]
        then
            >&2 echo "ID must match pattern ~{id_pattern}"
            error=1
        fi
```
The first line will eval to true only when `read_group.ID` is the literal value `id` and then the string `ID must match pattern id` gets printed to STDERR. The ID val _does_ match the pattern.

Personally I think we should restrict to alphanumeric characters (plus some separators?) in ID or at least disbar whitespace. Same for SM?

data_structures/read_group.wdl
adthrasher: My assumption was that the read group fields are unlikely to update in the future.

a-frantz: That is true. But what do we gain from these inline comments? Who are they meant for? I figure anyone who might see these inline comments would have to at least skimmed past the header comments. Is there a case I'm not considering?

data_structures/read_group.wdl
adthrasher: OK, I updated the checks. I always check for `sam_spec_pattern`. If `restrictive = true`, then `ID` disallows the literal string `id` and anything with a `space` and `SM` disallows the literal string `sample` with one or fewer additional characters (`sample.?`) and anything with a `space`.

I have three `String`s `id_pattern`, `sample_pattern`, and `restrictive_pattern` that contain the expressions. So if we decide to add additional patterns, we should be able to do that in the future fairly easily.

a-frantz (approved):

PR #149, opened 2024-05-22
refactor: Update ReadGroup to force uppercase PL. 

adthrasher:
Add sample_override to dnaseq-standard

### DIFF
--- a/data_structures/read_group.wdl
+++ b/data_structures/read_group.wdl
@@ -125,7 +125,7 @@ sam = pysam.AlignmentFile(os.environ["BAM"], "rb")
 out_file = open(os.environ["OUTFILE"], "w")
 header = sam.header.to_dict()["RG"]
 modified_header = []
-for read_group in header:
+for read_group in sorted(header, key=lambda d: d['ID']):
   modified_header.append({k:v.upper() if k=='PL' else v for k,v in read_group.items()})
 json.dump(modified_header, out_file)
 out_file.close()


data_structures/read_group.wdl
```diff
@@ -123,7 +123,11 @@ import json  # lint-check: ignore
 sam = pysam.AlignmentFile(os.environ["BAM"], "rb")
 
 out_file = open(os.environ["OUTFILE"], "w")
-json.dump(sam.header.to_dict()["RG"], out_file)
+header = sam.header.to_dict()["RG"]
+modified_header = []
+for read_group in header:
+  modified_header.append({k:v.upper() if k=='PL' else v for k,v in read_group.items()})
```
a-frantz, line 129: `read_group.items()` is going to return a different order each run of this task. Is that an inconsistency we care about? I believe the order of `@RG` tags in the header _should_ be irrelevant. But we have `sort_star_input.py` to alphabetize these, so there are at least some tools out there where order matters.

adthrasher: That's a good point. However, it's likely a bigger issue with this pipeline. Before, we were just pulling the `RG` records and then assuming that the order matched the ordering of the FASTQ files.

adthrasher: I've added an explicit sort of the `read_group` records in `header`. They're now ordered by the `ID` key. Our `bam-to-fastq` workflow splits BAMs into BAM files of the name format `prefix.RGID.bam`. These are `glob`-ed and are sorted as BASH would (essentially alphabetically). So the sort should come down to the `RG ID` value, which is why we sort by `ID` here.

a-frantz (approved):

PR #153, opened 2024-05-29
feat: GATK MarkDuplicatesSpark

adthrasher:
Adds a tool definition for GATK's MarkDuplicates that is Spark-enabled.

Two options (`remove-all-duplicates` and `remove-sequencing-duplicates`) are intentionally omitted. They cannot be specified (even if `false`) if `duplicate-tagging-policy` is specified.


a-frantz (commented):
Is there no md5 created by the spark version?

adthrasher:
> Is there no md5 created by the spark version?

Nope. No idea why they dropped that. It produces an index, which is presumably single-threaded. Computing an MD5 shouldn't be that hard.

a-frantz (approved):

PR #156, opened 2024-05-30
fix: make index optional

### LINKED ISSUE #155: Picard sort task fails for non-coordinate sorts
In our implementation of the Picard `sort`, we attempt to move the bam index file since Picard uses the wrong name. However, the index file is only created if the sort order is `coordinate`. For other settings, this step will fail.

https://github.com/stjudecloud/workflows/blob/99827132a8d2e6e84bcfe38403dfe821efaffaca/tools/picard.wdl#L321

adthrasher:
For non-coordinate sorts, `picard SortSam` does not output a BAM index, even if `CREATE_INDEX` is `true`. This puts an existence check around the rename of the BAI file and also makes the output bam_index an optional output. Fortunately, it appears that we are not using the generated index in any of our pipelines as something always consumes and transforms the BAM downstream, so it ends up reindexed.

Resolves #155

### DIFF
--- a/tests/tools/test_picard.py
+++ b/tests/tools/test_picard.py
@@ -1,0 +1,7 @@
+import pathlib
+import pytest
+
+@pytest.mark.workflow('picard_sort_queryname')
+def test_picard_sort_queryname(workflow_dir):
+    exists = pathlib.Path(workflow_dir, 'test-output/out/sorted_bam_index/test.bwa_aln_pe.sorted.bam.bai').exists()
+    assert exists is False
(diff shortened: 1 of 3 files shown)


a-frantz (commented):
Can you add a code comment explaining why the index is optional despite the hardcoded `--CREATE_INDEX true`?

> Fortunately, it appears that we are not using the generated index in any of our pipelines as something always consumes and transforms the BAM downstream, so it ends up reindexed.

Another instance where E2E tests would be helpful as a sanity check.

PR #165, opened 2024-07-14
feat: use wdl-format on everything

a-frantz:


workflows/chipseq/chipseq-standard.wdl
```diff
@@ -6,10 +6,14 @@ import "../../tools/ngsderive.wdl"
 import "../../tools/picard.wdl"
 import "../../tools/samtools.wdl"
 import "../../tools/util.wdl"
-import "../general/bam-to-fastqs.wdl" as b2fq
-import "https://raw.githubusercontent.com/stjude/seaseq/2.3/workflows/workflows/mapping.wdl" as seaseq_map
-import "https://raw.githubusercontent.com/stjude/seaseq/3.0/workflows/tasks/samtools.wdl" as seaseq_samtools
-import "https://raw.githubusercontent.com/stjude/seaseq/3.0/workflows/tasks/seaseq_util.wdl" as seaseq_util
+import "../general/bam-to-fastqs.wdl"
+    as b2fq
```
claymcleod, line 10: I definitely do not like this behavior. In my mind, imports should take one line irrespective of how long the line it (at least right now while URLs are the primary thing being imported).

claymcleod (commented):
Everything else I skimmed looked really good!

workflows/chipseq/chipseq-standard.wdl
a-frantz: Breaking each clause onto it's own line was a bit of an experiment. If you feel strongly I have no problem changing this behavior.

Issue #172, opened 2024-07-15
Fragility in workflows around malformed/missing RG records


a-frantz:

a-frantz:
Partially addressed for STAR in #202

Issue #171, opened 2024-07-15
more dynamically allocated mem


a-frantz:
Some of our tasks allocate a large static amount of RAM that is often an over-allocation for many inputs.
One example here: https://github.com/stjudecloud/workflows/blob/main/tools/ngsderive.wdl#L366

DuyLe0710edu:
Hi the team, I am Duy , a sophomore Bucknell Computer Science & Computer Engineering. I am really sorry to contribute to the project this late because I was having a family issue during March. I have experienced in full-stack and some ML/DL research. I also contribute to some project on github and I want to be the GSOCs 2025 for Jude Research Hospital this year. I have experienced with medical application and familiar with hospital settings. I will also put my LinkedIn here: https://www.linkedin.com/in/duy-le-dev/

DuyLe0710edu:
Can you assign me this issues ? I think I can get this, RAM over-allocation is the issue and I have handled one of them before. I came up with a possible solution of we can implement the dynamic memory allocation inside star.wdl because the start using the fixed memory of 50GB allocation so it should be adequate for that. Please correct me if I am wrong.

adthrasher:
@DuyLe0710edu - This issue is lacking some information. To properly implement dynamic memory allocation where we have not previously implemented it, someone will need to either dig in to the underlying tool implementations to understand the memory consumption as a function of the input data size or run a significant amount of real world data through to profile the memory usage. The latter will be difficult for an outside contributor as we cannot share patient data in an uncontrolled manner due to privacy concerns. Without St. Jude data, a contributor would need to gather external data sources and determine that they are likely to perform reasonably similar to St. Jude data.

DuyLe0710edu:
Thank you for the feedback on the dynamic memory allocation implementation. I understand the challenges you've outlined, and here's how I propose it. I think I will make it simple by using multiplication values in memory allocation formulas that intentionally allocate more memory than might be minimally required. If you can provide me the mock data or the format of that I think it could be easier. Do yo recommend any public dataset out there, I can alter the format and try it on. Please assign me this problem.

a-frantz:
[...]

Issue #170, opened 2024-07-15
Tasks that take both pos/name sorted probably need different resources


a-frantz:
Major culprit here is HTSEQ: https://github.com/stjudecloud/workflows/blob/main/tools/htseq.wdl

It has a pretty terrible sort algorithm and eats up resources when the input is position sorted. We've exposed the name sort option but still allocate a large amount of memory and disk. Neither are likely needed.

PriyankaaXD:
Looked into this. `tools/htseq.wdl`'s `count` task already has dynamic memory/disk allocation keyed off `pos_sorted` and `bam_size` (via `modify_memory_gb`/`modify_disk_size_gb`), so the original HTSeq case looks resolved.

Since the title says "tasks" (plural), I audited `tools/*.wdl` for other tasks that take a queryname/coordinate-sorted BAM and found one clear remaining case, confirmed against the source:

**`samtools.wdl` → `merge`**: takes `name_sorted` (queryname vs. coordinate sorted input). `disks` is already dynamic:
```
Int disk_size_gb = ceil(bams_size * 2 + header_size) + 10 + modify_disk_size_gb
```
but `memory` is hardcoded `"4 GB"` with no `modify_memory_gb` escape hatch at all.

I don't want to guess at a size-scaling formula for memory without profiling data (`samtools merge` memory is driven more by open file handles/thread buffers than raw BAM size, so I'd rather not assert a multiplier I can't justify). But adding a `modify_memory_gb` override — mirroring the pattern `disks` and htseq already use — seems like a safe, honest first step that doesn't require guessing.

Mind if I take this and open a PR adding `modify_memory_gb` to `merge`? Happy to dig further into other tasks too if you think more are in scope.

a-frantz:
Thanks for looking into this! I think that for the time being, we can leave `samtools.merge` alone with a static allocation. As you point out, the merge algorithm's memory use is not driven by BAM size; it _should_ use less than 4GB for all files.

There is an argument that we should add a `modify_memory_gb` argument to the task anyway (relevant discussion happening on the Sprocket repo - https://github.com/stjude-rust-labs/sprocket/issues/713 ), although that reasoning isn't tied to sort order and is a tangent off the topic of this issue. If we end up adopting the policy as suggested in the proposed 713 lint, that would be a large update across the whole repo instead of just this samtools task.

[...]

Issue #169, opened 2024-07-15
FastQC `prefix` convention is fragile


a-frantz:
See here: https://github.com/stjudecloud/workflows/blob/main/tools/fastqc.wdl#L59

The above may not work if `prefix` is messed with.

DuyLe0710edu:
I am interested in this problem, @a-frantz can you assign this to me as well ?

DuyLe0710edu:
thanks I will start working on this

DuyLe0710edu:
Hi, I might have a bit issue with setting up the CI tests for my repository and on my local. I though it should be on contribute.md but some how I cannot find the instruction. Can you quickly go through how I can do that, I think my solution is solid now, just wait for testing and pull request.

DuyLe0710edu:
@adthrasher  I wanted to share an update on my progress over the past few days. I’ve spent the last three days educating myself on WDL and have begun diving into the problem areas. I’m pleased to say that I was able to work through three issues. For one of these—a minor issue—I’m ready to submit a pull request, as I believe it should pass the tests without any issues.

For the other two issues, I’m still waiting on your guidance on how to run the IC tests locally. I did check the contribute.md file, but I couldn’t find the necessary instructions there.

Additionally, I’m starting to write **my proposal** and will send it over to your email once it’s ready for your review. I’d really appreciate the chance to connect further, perhaps via a call or over coffee, to discuss these updates and learn more about your insights.

DuyLe0710edu:
please check my pull request on this

a-frantz:
@DuyLe0710edu apologies, this issue didn't contain a clear explanation of the problem. Your PR doesn't address what this issue is about. Although I just investigated locally, and this is a non-issue. The FastQC prefix is fine as it is, and no fix needs to be applied. I was mistaken when I opened this issue.

PR #173, opened 2024-07-16
Clean up various new lints

a-frantz:
Lint directives aren't working for `LineWidth`, but aside from that every lint (even ones not in `sprocket` yet) has been addressed.

Where I felt it was appropriate, existing `TODO` items were converted into issues (#166 , #167 , #168 , #169 , #170 , #171 , #172)


tools/arriba.wdl
```diff
@@ -137,10 +137,14 @@ task arriba {
         File? annotate_fusions
         File? protein_domains
         File? wgs_svs
-        Array[String] interesting_contigs = ["1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12", "13", "14", "15", "16", "17", "18", "19", "20", "21", "22", "X", "Y", "AC_*", "NC_*"]
+        Array[String] interesting_contigs = flatten(
+            [range(23), ["X", "Y", "AC_*", "NC_*"]]
```
a-frantz, line 141: https://github.com/openwdl/wdl/blob/wdl-1.1/SPEC.md#range

This is going to include `0`, so this needs a patch before we can merge

data_structures/read_group.wdl
```diff
@@ -180,7 +199,8 @@ task validate_read_group {
             if [[ ~{read_group.ID} =~ ^~{id_pattern}$ ]] \
             || [[ ~{read_group.ID} =~ ~{restrictive_pattern} ]]
             then
-                >&2 echo "ID (~{read_group.ID}) must not match pattern ~{id_pattern} nor ~{restrictive_pattern}"
+                >&2 echo "ID (~{read_group.ID}) must not match patterns:"
```
adthrasher, line 202: I'm fairly certain that `pattern` should be singular because of the use of `nor`.

adthrasher: We could also remove `nor` and replace it with some other delimiter.

data_structures/read_group.wdl
```diff
@@ -204,7 +224,8 @@ task validate_read_group {
                 if [[ "~{read_group.SM}" =~ ^~{sample_pattern}$ ]] \
                 || [[ "~{read_group.SM}" =~ ~{restrictive_pattern} ]]
                 then
-                    >&2 echo "SM must not match pattern ~{sample_pattern} nor ~{restrictive_pattern}"
+                    >&2 echo "SM must not match patterns:"
+                    >&2 echo "~{sample_pattern} nor ~{restrictive_pattern}"
```
adthrasher, line 228: Same thought here.

tools/cellranger.wdl
```diff
@@ -2,6 +2,7 @@
 ##
 ## This WDL file wrap the [10x Genomics Cell Ranger](https://support.10xgenomics.com/single-cell-gene-expression/software/pipelines/latest/what-is-cell-ranger) tool.
 ## Cell Ranger is a tool for handling scRNA-Seq data.
+#@ except: LineWidth
```
adthrasher, line 5: I assume this means that a lint directive anywhere in the preamble applies to the file? I've been putting them on the first line, but I think it makes sense that it is implemented as anywhere in the preamble.

workflows/general/alignment-post.wdl
```diff
@@ -11,7 +11,8 @@ workflow alignment_post {
         outputs: {
             processed_bam: "Input BAM after being transformed by standard processing",
             bam_index: "BAI index associated with `processed_bam`",
-            bam_checksum:  "STDOUT of the `md5sum` command run on the input BAM that has been redirected to a file"
+            bam_checksum:  "STDOUT of the `md5sum` command run on the input BAM that has been redirected to a file",
+            validate_report: "Validation report produced by `picard ValidateSamFile`. Validation warnings and errors are logged."
```
adthrasher, line 15: Does this need a trailing comma?

workflows/general/samtools-merge.wdl
```diff
@@ -20,7 +20,7 @@ workflow samtools_merge {
     }
     input {
         Array[File] bams
-        String prefix = basename(bams[0], ".bam")  # TODO is this a sane default? Or should be required?
```
adthrasher, line 23: Did you check where this is used to ensure that we're providing something?

tools/cellranger.wdl
a-frantz: My understanding (without having reviewed the code very closely) is that a lint directive anywhere in the preamble will work. I think my preferred style would be to place directly above the version statement (i.e. either the first line if there are no preamble comments, or separated from the preamble comments by a blank).

```
## Cell Ranger blahblahblah

#@ except: LineWidth
version 1.1

task foo {}
```
But this currently sets off the PreambleWhitespace rule.

workflows/general/alignment-post.wdl
a-frantz: Huh. I was actually under the impression that a trailing comma there would cause errors. But I just double checked in both `miniwdl` and `sprocket` and they're ok with or without that comma 🤷

workflows/general/samtools-merge.wdl
a-frantz: Yea we're good on that front. Only used in `dnaseq-core` and there `prefix` is supplied.

data_structures/read_group.wdl
a-frantz: Updated

adthrasher (approved):

PR #180, opened 2024-09-19
ci: use public Sprocket action

adthrasher:
Switch to the public Sprocket action instead of installing Rust and Sprocket on the runner manually.

### DIFF
--- a/.github/workflows/sprocket-check.yaml
+++ b/.github/workflows/sprocket-check.yaml
@@ -1,0 +1,13 @@
+name: Workflows Sprocket Check
+
+on: [push]
+
+jobs:
+    sprocket_check:
+        runs-on: ubuntu-latest
+        steps:
+        - uses: actions/checkout@v2
+        - name: Run sprocket
+          uses: stjude-rust-labs/sprocket-action@v0.3.0
+          with:
+            exclude-paths: template
--- a/.github/workflows/sprocket-check.yaml
+++ b/.github/workflows/sprocket-check.yaml
@@ -8,6 +8,6 @@ jobs:
         steps:
         - uses: actions/checkout@v2
         - name: Run sprocket
-          uses: stjude-rust-labs/sprocket-action@v0.3.0
+          uses: stjude-rust-labs/sprocket-action@v0.4.0
           with:
             exclude-patterns: template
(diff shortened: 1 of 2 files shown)


a-frantz:
Should we run this twice? Once on "check mode" and once on "lint mode"? Also, maybe should be changed upstream (after I approved it 😬 ) but `--deny-warnings` doesn't need to be specified on top of `--deny-notes`. It probably makes more sense to collapse that into one argument with 3 options (fail on `errors`, `warnings`, or `notes`).

adthrasher:
> Should we run this twice? Once on "check mode" and once on "lint mode"? Also, maybe should be changed upstream (after I approved it 😬 ) but `--deny-warnings` doesn't need to be specified on top of `--deny-notes`. It probably makes more sense to collapse that into one argument with 3 options (fail on `errors`, `warnings`, or `notes`).

We could certainly run it twice. For the arguments, GitHub Actions doesn't seem to support `input` typing at the action-level. A workflow can specify types (`choice`, `boolean`, and `environment`), but it doesn't appear that individual actions have that functionality.

a-frantz:
> > Should we run this twice? Once on "check mode" and once on "lint mode"? Also, maybe should be changed upstream (after I approved it 😬 ) but `--deny-warnings` doesn't need to be specified on top of `--deny-notes`. It probably makes more sense to collapse that into one argument with 3 options (fail on `errors`, `warnings`, or `notes`).
> 
> We could certainly run it twice. For the arguments, GitHub Actions doesn't seem to support `input` typing at the action-level. A workflow can specify types (`choice`, `boolean`, and `environment`), but it doesn't appear that individual actions have that functionality.

Well if there's no good way to collapse `deny-warnings` and `deny-notes` I guess that's fine. It's just not the friendliest interface.

a-frantz:
in `sprocket check` the `template` dir isn't being excluded properly. Once that's fixed I'll approve

adthrasher:
> in `sprocket check` the `template` dir isn't being excluded properly. Once that's fixed I'll approve

Yeah, I'm unclear on why that is happening. When I run the same thing locally, it picks up the excluded patterns correctly. I'm working on a change to the action to address it.

a-frantz (approved):

PR #184, opened 2024-11-08
feat: add methylation pipelines

adthrasher:
Add a methylation process workflow that generates unfiltered, normalized Beta values for a sample. Add a second pipeline that consumes an array of unfiltered, normalized beta values for a set of sample. Applies filtering and generates UMAP coordinates.


workflows/methylation/methylation-preprocess.wdl
```diff
@@ -0,0 +1,209 @@
...
+            rgset: "RGSet object",
+            rset: "RatioSet object",
+            annotation: "Annotation object",
+            beta: "Beta values",
+            cn_values: "Copy number values",
+            m_values: "M values",
+            pheno: "Phenotype data",
+            pheno_data: "Phenotype data",
+            probe_names: "Probe names",
+            sample_names: "Sample names",
```
adthrasher, line 19: Some of these need to be removed. Charlie's original script output a lot of intermediate files, which we don't really need.

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,361 @@
...
+    runtime {
+        container: "ghcr.io/stjudecloud/umap:branch-methylation-0.5.7-1"
+        memory: "8 GB"
+        cpu: 1
+        disks: "~{disk_size_gb} GB"
+        maxRetries: 1
+    }
+}
+
+task plot_umap {
```
adthrasher, line 316: This task is primarily for testing purposes. I'm not sure we want to plot the UMAPs here. They'll be hosted on Pecan and use the ProteinPaint scatter plot library.

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,361 @@
...
+
+    Int memory_gb = ceil(size(unfiltered_normalized_beta, "GiB") * 2) + addl_memory_gb
+    Int disk_size_gb = ceil(size(unfiltered_normalized_beta, "GiB") * 2)
+
+    command <<<
+        echo "Combining data"
+        input_files=""
+        for file in ~{sep(" ", unfiltered_normalized_beta)}
+        do
+            #ln -s ~{sep(" ", unfiltered_normalized_beta)} .
```
a-frantz, line 118:
```suggestion
```

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,361 @@
...
+            ln -s $file .
```
a-frantz, line 119: Why do these need to be localized the CWD? Just looking at the Python, I would think that's unnecessary. Can you add a comment explaining the reason?

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,361 @@
...
+
+        python run.py $input_files
+
+    >>>
+
+    output {
+        File combined_beta = combined_file_name
+    }
+
+    #@ except: ContainerValue
```
a-frantz, line 160: IDK if we want this here. In the CI we've specified `--except ContainerValue` so this is redundant in that context. I guess the highlight is annoying in the IDE which could be reason enough. But I'm worried about missing this if/when we finally decide to adhere to `ContainerValue`.

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,361 @@
...
+        memory: "~{memory_gb} GB"
+        cpu: 1
+        disks: "~{disk_size_gb} GB"
+        maxRetries: 1
+    }
+}
+
+task filter_probes {
+    meta {
+        description: "Filter probes based on standard deviation"
```
a-frantz, line 172: Can we add a `help` key that summarizes the logic used to filter?

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,361 @@
...
+
+        python filter.py --num_probes ~{num_probes} beta.csv
+    >>>
+
+    output {
+        File filtered_beta_values = "filtered_beta.csv"
+        File filtered_probes = "filtered_probes.csv"
+    }
+
+    #@ except: ContainerValue
```
a-frantz, line 253: same concern here

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,361 @@
...
+    runtime {
+        container: "quay.io/biocontainers/pandas:2.2.1"
+        memory: "28 GB"
```
a-frantz, line 256: 28 is an odd number. Isn't this going to depend on the size of the input?

workflows/methylation/methylation-cohort.wdl
a-frantz: I think it would be useful (if not necessary) for non-Pecan users of the workflow. It is also quite a lightweight task, so I'm not concerned about wasted compute. Maybe a boolean toggle for skipping this step?

workflows/methylation/methylation-preprocess.wdl
a-frantz: Should we create an `IntermediateFiles` struct like we have for QC? IDK if that would be useful, just an idea. Is there a context where a user would want the intermediate files? Maybe even just for debugging?

workflows/methylation/methylation-preprocess.wdl
```diff
@@ -0,0 +1,209 @@
...
+        R --no-save <<SCRIPT
+            options(error = function() traceback(3))
+
+            library(minfi)
+            library(dplyr)
+            library(data.table)
+            library(IlluminaHumanMethylationEPICmanifest)
+            library(IlluminaHumanMethylationEPICanno.ilm10b4.hg19)
+
+            set.seed(1)
```
a-frantz, line 100: Is there ever a reason for setting another seed?

docker/minfi/1.48.0-1/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM quay.io/biocontainers/bioconductor-minfi:1.48.0--r43hdfd78af_0
+
+RUN R --no-save <<SCRIPT
+    if (!require("BiocManager", quietly = TRUE))
+        install.packages("BiocManager", repos='http://cran.us.r-project.org')
+    BiocManager::install("IlluminaHumanMethylationEPICmanifest")
+    BiocManager::install("IlluminaHumanMethylationEPICanno.ilm10b4.hg19")
```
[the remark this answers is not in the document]
adthrasher: Those two R packages wrap the Illumina manifest files for the EPIC array (https://support.illumina.com/downloads/infinium-methylationepic-v1-0-product-files.html) and the second adds the genomics coordinates (in hg19, because that's what Illumina provides). The source CSV is ~500MB and contains a lot of metadata for each probe that is found on the EPIC array. They're necessary for the steps in `preprocess`, particularly `MapToGenome`.

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,361 @@
...
+        description: "Combine data from multiple samples"
+        outputs: {
+            combined_beta: "Combined beta values for all samples"
+        }
+    }
+
+    parameter_meta {
+        unfiltered_normalized_beta: "Array of unfiltered normalized beta values for each sample"
+        combined_file_name: "Name of the combined file"
+        addl_memory_gb: "Additional memory to allocate for the task (in GB)"
```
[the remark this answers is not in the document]
adthrasher: I thought I copied this from somewhere, but it's obviously not from our standard set. So maybe I didn't.

workflows/methylation/methylation-preprocess.wdl
adthrasher: These arrays have two types of probes (helpfully Infinium type 1 and Infinium type 2). The probes also cover a varying number of CpG sites. The normalization method works by taking `N` probes of each type that each have 1, 2, or 3 CpGs. So it selects 6N probes in total. This selection is "random" if the seed is not fixed, of course. Since we're processing samples individually in this step instead of as a cohort, I want to seed to be consistent so the same set of probes is chosen for each sample.

docker/minfi/1.48.0-1/Dockerfile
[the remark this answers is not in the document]
a-frantz: So this is essentially downloading reference files? Should these be included in the docker image?

workflows/methylation/methylation-preprocess.wdl
a-frantz: Makes sense. Can you add that to the documentation somewhere? Maybe it should go in `meta.help`? Or just an embedded "normal" comment, not sure. Your call.

docker/minfi/1.48.0-1/Dockerfile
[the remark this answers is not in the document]
adthrasher: Sort of, but they're R packages that the underlying `minfi` functions expect to exist. So there isn't a way to provide just the Illumina manifest as a file.

workflows/methylation/methylation-cohort.wdl
adthrasher: I was working off an older version of the CI when I resolved the lint warnings. I can remove these now that my branch is up-to-date.

workflows/methylation/methylation-cohort.wdl
adthrasher: 28 is a bit random. I'll take another look. It won't depend on the size of the input, but rather the width of the matrix, which is a bit harder to calculate upfront. It's the width because I'm reading a single row in to memory at a time.

workflows/methylation/methylation-preprocess.wdl
adthrasher: I'll give it some thought, but I don't think there is any value to most of these files. Certainly not `sample_names` because that will have one entry with the name of the current sample. I can see outputting the M value and CN value as those could be useful.

workflows/methylation/methylation-cohort.wdl
a-frantz: Speaking of the TSV functions, you could try reading into an `Array[Array[String]]` and getting the `length()` of the inner array?

workflows/methylation/methylation-cohort.wdl
a-frantz: IMO if you are going to localize the files, the symlinks needs to be removed before execution ends. Otherwise when debugging the WD is going to be full of broken links. I'd rather just skip the localization entirely.

adthrasher:
@xuqiu5 - Here is the PR for the methylation workflows.

adthrasher:
@claymcleod - I pulled the embedded R and Python scripts in to standalone files that are now built in to our Docker images. This should be easier for Charlie to utilize.

docker/pandas/2.2.1-0/filter.py
```diff
@@ -0,0 +1,47 @@
...
+            probe = line[0]
+            sd = np.std([float(x) for x in line[1:]])
+            data.append([probe, sd])
+
+    sd_df = pd.DataFrame(data, columns=["probe", "sd"]).set_index("probe")
+
+    # Filter probes based on standard deviation
+    filtered_probes = sd_df.sort_values("sd", ascending=False).head(args.num_probes).index
+
+    pd.Series(filtered_probes, index=filtered_probes).to_csv('filtered_probes.csv', index=False)
```
a-frantz, line 37:
```suggestion
    pd.Series(filtered_probes, index=filtered_probes).to_csv("filtered_probes.csv", index=False)
```

docker/pandas/2.2.1-0/filter.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # Filter beta values
+    with open("beta.csv", "r") as f:
```
a-frantz, line 40:
```suggestion
    with open(args.beta, "r") as f:
```
I think?

docker/pandas/2.2.1-0/filter.py
```diff
@@ -0,0 +1,47 @@
...
+        reader = csv.reader(f)
+        header = next(reader)
+        header[0] = "probe"
+        beta_data = [line for line in reader if line[0] in filtered_probes]
+
+    bd = pd.DataFrame(beta_data, columns=header).set_index("probe")
+    bd.to_csv("filtered_beta.csv")
```
a-frantz, line 47: In line 37 you write to this same file. Is line37 supposed to be `unfiltered` instead?

docker/pandas/2.2.1-0/filter.py
a-frantz, line 1: This script looks buggy to me. I'm confused by it at least.

docker/python-plotting/1.0.0/plot_umap.py
```diff
@@ -0,0 +1,13 @@
...
+
+# Read UMAP embedding
+umap = pd.read_csv(sys.argv[1], index_col=0)
+
+# Plot UMAP
+plt.scatter(umap["UMAP1"], umap["UMAP2"])
+plt.xlabel("UMAP1")
+plt.ylabel("UMAP2")
+plt.title("UMAP Embedding")
+plt.savefig("umap.png")
```
a-frantz, line 13: Small nit: can we take 2 args where the second is the output file name? And at that point it probably makes sense to use `argparse`

docker/umap/0.5.7-1/generate_umap.py
a-frantz, line 1: Again, can you add `argparse` for determining the input and output file names?

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,245 @@
+version 1.1
+
+workflow methylation_cohort {
+    meta {
+        description: "Process methylation data for a cohort of samples"
+        outputs: {
+            combined_beta: "Matrix (in CSV format) containing beta values for every (common) probe on the array as rows and all of the input samples as columns.",
+            filtered_beta: "Matrix (in CSV format) containing only beta values for the retained probes (top N highest standard deviation) for all provided samples.",
```
a-frantz, line 8: So the difference between these files is that `combined_beta` contains _every common_ probe, and `filtered_beta` is that same set but filtered to top N hight STDEV? Can that difference be stated explicitly?

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,245 @@
...
+            filtered_probeset: "List of probe names that were retained after filtering to the top N highest standard deviation",
+            umap_embedding: "UMAP embedding for all samples",
+            umap_plot: "UMAP plot for all samples",
+        }
+    }
+
+    parameter_meta {
+        unfiltered_normalized_beta: "Array of unfiltered normalized beta values for each sample"
+        max_length: "Maximum number of beta files to merge before using iteration"
```
a-frantz, line 17: I wonder if this should be hidden somehow... Maybe `group: "Advanced"` or even `"Developer Options"`?

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,245 @@
...
+
+    input {
+        Array[File] unfiltered_normalized_beta
+        Int max_length = 500
+    }
+
+    Int beta_length = length(unfiltered_normalized_beta)
+
+    if (beta_length > max_length){
+        scatter ( merge_num in range((beta_length / max_length) + 1)){
```
a-frantz, line 28:
```suggestion
        scatter (merge_num in range((beta_length / max_length) + 1)){
```
Surprised this isn't firing a lint warning. We should address that.

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,245 @@
...
+            # Get the sublist of beta files
+            scatter ( beta_num in range(max_length)){
```
a-frantz, line 30:
```suggestion
            scatter (beta_num in range(max_length)){
```

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,245 @@
...
+                    if merge_num > 0
+                    then beta_num + (merge_num * max_length)
+                    else beta_num
+                )
+                if (num < beta_length){
+                    File bam_list = unfiltered_normalized_beta[num]
+                }
+            }
+        }
+        scatter (iter in range(length(bam_list))){
```
a-frantz, line 41:
```suggestion
        scatter (iter_index in range(length(bam_list))){
```

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,245 @@
...
+        String combined_file_name = "combined_beta.csv"
+        Int modify_memory_gb = 0
+    }
+
+    Int memory_gb = ceil(size(unfiltered_normalized_beta, "GiB")) + modify_memory_gb
+    Int disk_size_gb = ceil(size(unfiltered_normalized_beta, "GiB") * 2)
+
+    command <<<
+        echo "Combining data"
+
```
a-frantz, line 116:
```suggestion
```
Seems unnecessary to me (and technically adds a failpoint that should be addressed with `set -euo pipefail`, so I'd rather just axe the echo statement)

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    input {
+        File beta_values
+        Int num_probes = 10000
+    }
+
+    Int disk_size_gb = ceil(size(beta_values, "GiB") * 2)
+
+    #@ except: LineWidth
```
a-frantz, line 155:
```suggestion
```
Don't think it's needed?

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,245 @@
...
+    command <<<
+        ln -s ~{beta_values} beta.csv
+
+        python $(which filter.py) --num_probes ~{num_probes} beta.csv
```
a-frantz, line 159:
```suggestion
        python $(which filter.py) --num_probes ~{num_probes} ~{beta_values}
```
With my request to add argparse this can be a bit cleaner (and avoids adding `set -euo pipefail`)

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,245 @@
...
+    >>>
+
+    output {
+        File filtered_beta_values = "filtered_beta.csv"
+        File filtered_probes = "filtered_probes.csv"
```
a-frantz, line 164: Should these use the `prefix` naming convention?

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,245 @@
...
+        filtered_beta_values: "Filtered beta values for all samples"
+    }
+
+    input {
+        File filtered_beta_values
+    }
+
+    Int disk_size_gb = ceil(size(filtered_beta_values, "GiB") * 2)
+
+    #@ except: LineWidth
```
a-frantz, line 194:
```suggestion
```

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,245 @@
...
+    command <<<
+        ln -s ~{filtered_beta_values} beta.csv
+
+        python $(which generate_umap.py)
```
a-frantz, line 198:
```suggestion
        python $(which generate_umap.py) ~{name of arg}
```

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,245 @@
...
+    >>>
+
+    output {
+        File umap = "umap.csv"
```
a-frantz, line 202: Again, should this be configurable?

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,245 @@
...
+    input {
+        File umap
+    }
+
+    command <<<
+        python $(which plot_umap.py) ~{umap}
+    >>>
+
+    output {
+        File umap_plot = "umap.png"
```
a-frantz, line 235: Configurable?

workflows/methylation/methylation-preprocess.wdl
```diff
@@ -0,0 +1,98 @@
+version 1.1
+
+workflow methylation_preprocess {
+    meta {
+        description: "Process raw IDAT files from the Illumina methylation array for a single sample"
+        outputs: {
+            beta_swan_norm_unfiltered: "Normalized beta values for all probes",
+            beta_swan_norm_unfiltered_genomic: "Normalized beta values for all probes that map to the genome",
+            annotation: "Annotation table for probes.  Contains genomic coordinates and sequence and other information about the probes.",
```
a-frantz, line 9:
```suggestion
            annotation: "Annotation table for probes. Contains genomic coordinates and sequence and other information about the probes.",
```

workflows/methylation/methylation-preprocess.wdl
```diff
@@ -0,0 +1,98 @@
...
+            beta_unnorm: "Non-normalized beta values",
+            cn_values: "Copy number values",
+            m_values: "M values",
+            probe_names: "Probe names found on the array",
+            sample_names: "Sample name found on the array",
```
a-frantz, line 14: Is this informative? I don't get why I'd want it output. Especially because this will just be a single sample, right?

workflows/methylation/methylation-preprocess.wdl
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    call process_raw_idats { input:
+        idats,
+    }
+
+    #@ except: LineWidth
+    output {
+        File beta_swan_norm_unfiltered = process_raw_idats.beta_swan_norm_unfiltered
+        File beta_swan_norm_unfiltered_genomic = process_raw_idats.beta_swan_norm_unfiltered_genomic
```
a-frantz, line 33:
```suggestion
    output {
        File beta_swan_norm_unfiltered
            = process_raw_idats.beta_swan_norm_unfiltered
        File beta_swan_norm_unfiltered_genomic
            = process_raw_idats.beta_swan_norm_unfiltered_genomic
```
nit nit

workflows/methylation/methylation-preprocess.wdl
```diff
@@ -0,0 +1,98 @@
+version 1.1
+
+workflow methylation_preprocess {
```
a-frantz, line 3: To prevent code duplication, I think this whole workflow should be axed. What is it doing that isn't handled by the task itself?

workflows/methylation/methylation-preprocess.wdl
```diff
@@ -0,0 +1,98 @@
...
+        description: "Process raw IDAT files from the Illumina methylation array for a single sample"
+        outputs: {
+            beta_swan_norm_unfiltered: "Normalized beta values for all probes",
+            beta_swan_norm_unfiltered_genomic: "Normalized beta values for all probes that map to the genome",
+            annotation: "Annotation object",
+            beta_unnorm: "Beta values",
+            cn_values: "Copy number values",
+            m_values: "M values",
+            probe_names: "Probe names",
+            sample_names: "Sample names",
```
a-frantz, line 54: These descriptions should match the workflow ones

workflows/methylation/methylation-preprocess.wdl
```diff
@@ -0,0 +1,98 @@
...
+        Pair[File, File] idats
+        Int disk_size_gb = 10
+    }
+
+    String idat_base = basename(idats.left)
+    String out_base = sub(idat_base, "(_Grn|_Red).idat", "")
+
+    #@ except: LineWidth
+    command <<<
+        echo "Processing IDAT files"
```
a-frantz, line 73:
```suggestion
```
I don't see the need for this echo

workflows/methylation/methylation-preprocess.wdl
```diff
@@ -0,0 +1,98 @@
...
+        ln -s ~{idats.left} ~{idats.right} .
```
a-frantz, line 74:
```suggestion
        set -euo pipefail

        ln -s ~{idats.left} ~{idats.right} .
```

a-frantz (changes requested):
Lots of tiny nits, but this looks pretty great. I think the script migration is the right call. This seems much more manageable.

docker/minfi/1.48.0-1/methylation-preprocess.R
```diff
@@ -0,0 +1,88 @@
...
+suppressPackageStartupMessages(library("argparse"))
+# create parser object
+parser <- ArgumentParser()
+# specify options
+parser$add_argument("--idat_base", type="character", help="Base path to the idat files")
+parser$add_argument("--out_base", type="character", help="Base path to the output files")
+# parse the arguments
+args <- parser$parse_args()
+
+# These arrays have two types of probes (helpfully named Infinium type 1 and Infinium type 2).
```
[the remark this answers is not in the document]
adthrasher: Fair enough...

docker/minfi/1.48.0-1/methylation-preprocess.R
```diff
@@ -0,0 +1,88 @@
...
+# The probes also cover a varying number of CpG sites.
+# The normalization method works by taking N probes of each type that each have 1, 2, or 3 CpGs.
+# So it selects 6N probes in total. This selection is "random" if the seed is not fixed, of course.
+# Since we're processing samples individually in this step instead of as a cohort,
+# we want the seed to be consistent so the same set of probes is chosen for each sample.
+set.seed(1)
```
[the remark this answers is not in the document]
adthrasher: I also debated that. I think it'd be fine to expose it in the WDL, with a sufficiently descriptive `parameter_meta`.

docker/pandas/2.2.1-0/filter.py
adthrasher: Line 37 writes the list of probe names that are retained. It uses `filtered_probes.csv` rather than `filtered_beta.csv`. 
```
    pd.Series(filtered_probes, index=filtered_probes).to_csv('filtered_probes.csv', index=False)
```

workflows/methylation/methylation-cohort.wdl
adthrasher: Yes, that is the difference between the files. Where would you want that to be stated? It is in the quotes `parameter_meta` entries. Do those need additional text to make clear?

workflows/methylation/methylation-cohort.wdl
adthrasher: You're correct. It was when the code was in-lined and I didn't remove the lint directive.

workflows/methylation/methylation-preprocess.wdl
adthrasher: Originally there was no `methylation-standard.wdl` and I didn't want a task entrypoint as a preprocessing step. Now, I'm not sure exactly how this will be run in production. It is a good question though. How do we handle "workflows" that only have a single step? Do we write a workflow or invoke the `task` directly?

docker/pandas/2.2.1-0/filter.py
a-frantz: O whoops! That's a misread on my part.

workflows/methylation/methylation-cohort.wdl
a-frantz: I guess it's fine. I'm not sure where to add that clarifying text... I understood correctly after all.

workflows/methylation/methylation-preprocess.wdl
a-frantz: IMO we just use the task as the entrypoint. It's easy enough to run tasks using miniwdl. Granted, I don't think that's possible in cromwell? But we've already upgraded to 1.1 which isn't cromwell compatible, so not really a good reason to keep a workflow around.

My main concern with a "wrapper workflow" is that it should have dang-near identical meta entries, which is a source for discrepancies and maintenance headaches.

docker/pandas/2.2.1-0/filter.py
adthrasher: In order to avoid significant memory usage, I'm iterating over the array of beta values for each probe (matrix rows). I compute the standard deviation for each probe and store that in an array. Then I take the array of standard deviations and take the top `num_probes` entries. I then loop back through the array of beta values to apply the filtering and only retain the selected probes.

workflows/methylation/methylation-cohort.wdl
adthrasher: Before going that route, I'd probably just move it to a private declaration. I don't think anyone should change it, but that'd make it much clearer that it probably shouldn't be changed.

workflows/methylation/methylation-preprocess.wdl
adthrasher: In the single-sample context, it probably isn't informative. Especially, since it should match the prefix portion of the file name.

docker/pandas/2.2.1-0/filter.py
a-frantz: Sorry, I was apparently having some trouble yesterday and misread more than one line of this script 😅 

It makes sense to me now on a revisit. Thanks for the explanation! I think you should probably leave a comment that the two iterations over `args.beta` is intentional to reduce memory consumption. I can see a future version of us looking at it and saying "this can be done in one iteration!"

workflows/methylation/methylation-cohort.wdl
a-frantz: Having it as a private declaration sounds good to me. Should we make that change? I think we either need to move it so it can't be modified or need to add something to `meta` clarifying users shouldn't touch it.

workflows/methylation/methylation-preprocess.wdl
```diff
@@ -17,14 +17,17 @@ workflow methylation_preprocess {
 
     parameter_meta {
         idats: "Array of raw IDAT files from the Illumina methlyation array"
```
a-frantz, line 19:
```suggestion
        idats: "Pair of raw IDAT files from the Illumina methylation array"
```

docker/minfi/1.48.0-1/Dockerfile
a-frantz, line 1: Shouldn't this version end in `-0` not `-1`?

docker/umap/0.5.7-1/Dockerfile
a-frantz, line 1: Same Q here about the version number. Should be `-0` I think

docker/minfi/1.48.0-1/Dockerfile
adthrasher: Fair enough, we are programmers after all, counting starts from `0`.

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,250 @@
...
+        call combine_data as final_merge { input:
+            unfiltered_normalized_beta = inner_merge.combined_beta,
+        }
+    }
+
+    if (beta_length <= max_length){
+        call combine_data as simple_merge { input:
+            unfiltered_normalized_beta,
+        }
+    }
```
a-frantz, line 57: What's the deal with the different memory requirements used for these three calls to `combine_data`? Would it be better if `combine_data` dynamically determined memory by the number of inputs or something?

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,250 @@
...
+
+    input {
+        Array[File] unfiltered_normalized_beta
+        String combined_file_name = "combined_beta.csv"
+        Int modify_memory_gb = 0
+    }
+
+    Int memory_gb = ceil(size(unfiltered_normalized_beta, "GiB")) + modify_memory_gb
+    Int disk_size_gb = ceil(size(unfiltered_normalized_beta, "GiB") * 2)
```
a-frantz, line 111: Should we add some headroom for the OS on cloud instances? Just a constant of `+2` maybe?

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,250 @@
...
+        num_probes: "Number of probes to retain after filtering"
+    }
+
+    input {
+        File beta_values
+        String prefix = "filtered"
+        Int num_probes = 10000
+    }
+
+    Int disk_size_gb = ceil(size(beta_values, "GiB") * 2)
```
a-frantz: Same here re: a little wiggle room

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,250 @@
...
+        cpu: 1
+        disks: "~{disk_size_gb} GB"
+        maxRetries: 1
+    }
+}
+
+task filter_probes {
+    meta {
+        description: "Filter probes based on standard deviation"
+        help: "Probes are filtered by calculating the standard deviation of the beta value for each probe across all samples. The top N probes with the highest standard deviation are retained."
```
a-frantz, line 136:
```suggestion
        help: "Probes are filtered by calculating the standard deviation of the beta value for each probe across all samples. The top `num_probes` probes with the highest standard deviation are retained."
```

workflows/methylation/methylation-cohort.wdl
```diff
@@ -2,6 +2,7 @@ version 1.1
 
 workflow methylation_cohort {
     meta {
+        name: "Methylation Cohort"
```
a-frantz, line 5: This is redundant (if we keep the plans to "capital case" to determine `name` automatically)

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,251 @@
...
+            umap_embedding: "UMAP embedding for all samples",
+            umap_plot: "UMAP plot for all samples",
+        }
+    }
+
+    parameter_meta {
+        unfiltered_normalized_beta: "Array of unfiltered normalized beta values for each sample"
+    }
+
+    input {
```
a-frantz, line 20: Should `num_probes` be exposed at the workflow level? It seems reasonable to want to tweak that value.

workflows/methylation/methylation-cohort.wdl
adthrasher: To be honest, I don't understand it myself. I suppose I haven't tried the non-iterative merge. But for the iterative merge, the intermediate merge consistently takes more memory. I would like to have additional array cohorts to do some more testing, but these assignments work for now.

workflows/methylation/methylation-cohort.wdl
```diff
@@ -0,0 +1,251 @@
...
+        filtered_beta_values: "Filtered beta values for all samples"
+        prefix: "Prefix for output file. The extension `.csv` will be appended."
+    }
+
+    input {
+        File filtered_beta_values
+        String prefix = "umap"
+    }
+
+    Int disk_size_gb = ceil(size(filtered_beta_values, "GiB") * 2)
```
a-frantz: wiggle room?

workflows/methylation/methylation-preprocess.wdl
a-frantz: So should we remove it?

workflows/methylation/methylation-preprocess.wdl
```diff
@@ -0,0 +1,64 @@
...
+        Int disk_size_gb = 10
+    }
+
+    String idat_base = basename(idats.left)
+    String out_base = sub(idat_base, "(_Grn|_Red).idat", "")
+
+    command <<<
+        set -euo pipefail
+
+        ln -s ~{idats.left} ~{idats.right} .
```
a-frantz, line 36: New conversation, same line of code so unresolving.

This will break if the `idat_base` isn't used by both files. We're already symlinking here, might as well go all the way and give the links the appropriate name.

workflows/methylation/methylation-cohort.wdl
a-frantz: Ok. Can you leave a comment (or even open an issue) saying as much? "We should do more testing on the memory requirements of `combine_data`"

workflows/methylation/methylation-preprocess.wdl
adthrasher: The challenge there is that `left` and `right` aren't necessarily the green or red channels. A user could swap them around. We could look for that suffix too, but if we're assuming that it is there, then I think it's fair to assume the first part is correct also. These files are generated by the array software, so unless someone mangles them, they should be reliable.

a-frantz: Ah. So it's a bit more complicated than I realized. In that case, if we're going to make this assumption it should be documented in the `meta` somewhere visible. Preferably at both the task level and the workflow(s) level. But that has the problem of duplicating code...

Maybe a `warning` with the details at the task level, and then a `warning: "See task 'foo' for assumptions made"` at the workflow level?

workflows/methylation/methylation-cohort.wdl
```diff
@@ -15,10 +15,12 @@ workflow methylation_cohort {
 
     parameter_meta {
         unfiltered_normalized_beta: "Array of unfiltered normalized beta values for each sample"
+        num_probes: "Number of probes to retain after filtering"
```
a-frantz, line 18: This feels like it could have a little more information. Maybe keep this text as the description, and then add a help key?

workflows/methylation/methylation-cohort.wdl
a-frantz: Honestly I removed the redundant because I thought that was what you were asking for 😅 lol. I don't think it's a big deal either way, but the consistent thing would be to remove it.

a-frantz (approved):

PR #187, opened 2024-11-25
ci: tests

a-frantz:
Lots of lines/files changed here but nothing huge. Most of the changes are related to changing names/paths for our test files to be a little clearer in their contents. There's still some room for improvement there.

I've also added a `tests/workflows/input_json/` directory with some files that aren't hooked up to anything yet.

Includes some minor bug patches. Perhaps controversial, I changed our a longstanding convention for our `star_db` files. These are now expected to be "flat archives" (without a root directory). This makes us robust to the annoying case where the tar archive's filename doesn't match the inner contents.


a-frantz:
This CI test isn't working as expected - https://github.com/stjudecloud/workflows/blob/main/.github/workflows/lint-check.yaml#L43

It should be failing because I'm currently using a `branch-tests` tag but it is passing. Will take a look tomorrow.

adthrasher (approved):

a-frantz:
Last 3 commits broke the CI in an expected way. We should consider the CI green. Here's why:

- `Update lint-check.yaml` fixed the broken lint-check CI and it started catching the usage of `branch-*` images (and OOD `util` images 😬 )
- [`fix: bump to latest util image`](https://github.com/stjudecloud/workflows/pull/187/commits/0e796ea4deb8637f08d100c7469abfea0ab2d66c) fixed the OOD util images, but the `branch-*` tags were still in use so that continued to fail
- [chore: don't use branch image](https://github.com/stjudecloud/workflows/pull/187/commits/22487945b9df56cbefbfc8e1694e8d8ffa94b18b) fixed the lint-check CI but now the STAR tests are failing, as they are trying to pull an image which doesn't exist (and won't exist until this merges into main)

PR #191, opened 2024-12-06
revise: changes needed for wdl-doc compatibility

a-frantz:


tools/picard.wdl
```diff
@@ -819,7 +819,7 @@ task bam_to_fastq {
         prefix: "Prefix for the <type of file> file. The extension `<extension>` will be added."
         paired: {
             description: "Is the data Paired-End (true) or Single-End (false)?",
-            common: true,
+            group: "Common",
```
adthrasher, line 822: I realize I've skipped past a lot of these, but I think I prefer `common` to `Common`. Not enough to hold up the PR though.

workflows/general/alignment-post.wdl
```diff
@@ -8,6 +8,7 @@ import "https://raw.githubusercontent.com/stjude/XenoCP/4.0.0-alpha/wdl/workflow
 
 workflow alignment_post {
     meta {
+        name: "Alignment Post"
```
adthrasher, line 11: I'm not sure I'm a fan of having to specify these since they're mostly a rewritten form of the workflow name in the WDL.

adthrasher (dismissed):

tools/picard.wdl
a-frantz: I do agree that `common` looks better in code format. But for doc rendering, don't we want the titles for each param table capitalized? I suppose we could handle the capitalization within `wdl-doc` by just assuming that no one wants their table titles to be in lowercase.

workflows/general/alignment-post.wdl
a-frantz: Compromise, we can probably auto-create a `name` in `wdl-doc` by taking the workflow name and "title capitalizing" it. Then we leave out the `name` field for most of our workflows and only specify when it's a special case (like the ones that use parentheses).

a-frantz: That compromise sounds sane to me, but I also really don't think it's a big deal to expect users to type this field for each workflow. So I could go either way.

tools/picard.wdl
adthrasher: That was my thought. Take the value of `group` and capitalize the first letter. Maybe that's not the right approach to manipulate the values though.

workflows/general/alignment-post.wdl
adthrasher: I think that's probably reasonable (though you could omit it for now). Split the name on `_` characters and capitalize the resulting words, if `meta.name` does not exist.

a-frantz: I forget the crate name, but the crate we use for the `SnakeCase` and `PascalCase` rules should make this pretty trivial. It's a nifty crate, I was impressed when we first added it.

tools/picard.wdl
a-frantz: > Maybe that's not the right approach to manipulate the values though.

Can you elaborate here?

adthrasher: After I wrote it, I thought maybe we shouldn't manipulate the users' specified values. We'd be making assumptions about how to manipulate those (capitalize).

a-frantz: I think we can do it intelligently without too much overhead/complexity. Here's my thinking for how we handle it:
If a `group` value is all lowercase, assume the user wants it to be capitalized in the docs (at least in the title[s]). If there's any kind of funky capitalization, leave it be and use as is.

PR #192, opened 2024-12-13
feat: add trimmomatic to tools/

mjgattas:


20241213_115526_trim_single_end/wdl/trimmomatic.wdl
```diff
@@ -0,0 +1,50 @@
...
+        # String adapter_file
+        Int leading
+        Int trailing
+        Int window_size
+        Int window_quality
+        Int minlen
+    }
+
+    command <<<
+        cd /usr/local/share/trimmomatic-0.36-5
```
a-frantz, line 34: Are you sure this is necessary? biocontainers _should_ have added a `trimmomatic` executable to the $PATH, as that's what they usually do. It should be unneeded to invoke the full path here.

mjgattas: I was trying to work out access to the executable, but without cd'ing to the exact directory I wasn't able to get it to work

adthrasher: That's likely because it's being called with `java`. There is a utility wrapper script at `/usr/local/bin/trimmomatic` that provides the functionality and is on `PATH`. So you should be able to swap `java -jar trimmomatic.jar` with `trimmomatic` and remove the directory change.

mjgattas: Thank you @adthrasher! Hadn't noticed that distinction

mjgattas:
@a-frantz I think the PR is in enough of a stable feature state that your review/suggestions are welcome. Please let me know how I can shore this up so it's ready for approval/merging

tools/trimmomatic.wdl
```diff
@@ -0,0 +1,151 @@
...
+        Int threads = 8
+        Int? trailing
+        Int window_size
+        Int window_quality
+        Boolean phred64
+    }
+
+    command <<<
+        ls -l /usr/local/share/trimmomatic-0.36-5/adapters/
+        trimmomatic PE -phred33 -threads ~{threads} \
```
a-frantz, line 124: I think these tasks should be combined. You can dynamically select the subcommand with `~{if (defined(<read_two>)) then "PE" else "SE"}`

tools/trimmomatic.wdl
```diff
@@ -0,0 +1,151 @@
...
+        -trimlog trimlog.txt ~{input_file_forward} ~{input_file_reverse} \
+        output_forward.fq.gz output_reverse.fq.gz \
+        ~{if (phred64) then "-phred64" else "-phred33"} \
+        ILLUMINACLIP:~{illumina_clip} \
+        ~{if defined(leading) then "LEADING:~{leading}" else ""} \
+        ~{if defined(trailing) then "TRAILING:~{trailing}" else ""} \
+        SLIDINGWINDOW:~{window_size}:~{window_quality} \
+        MINLEN:~{minlen} \
+        ~{if defined(crop) then "CROP:~{crop}" else ""} \
+        ~{if defined(headcrop) then "HEADCROP:~{headcrop}" else ""} \
```
a-frantz, line 134: style: indent the line continuations by one level (four spaces) and drop this last slash

tools/trimmomatic.wdl
```diff
@@ -0,0 +1,151 @@
...
+        Int minlen
+        Int threads = 8
+        Int? trailing
+        Int window_size
+        Int window_quality
+        Boolean phred64
+    }
+
+    command <<<
+        ls -l /usr/local/share/trimmomatic-0.36-5/adapters/
```
a-frantz, line 123: All our commands that contain more than one Bash statement should start with the line `set -euo pipefail`. This causes the script to fail in some common error cases instead of silently continuing and being hard to debug. (I recommend starting all shell scripts with that line.)

tools/trimmomatic.wdl
```diff
@@ -0,0 +1,151 @@
...
+        ~{if (phred64) then "-phred64" else "-phred33"} \
+        ILLUMINACLIP:~{illumina_clip} \
+        ~{if defined(leading) then "LEADING:~{leading}" else ""} \
+        ~{if defined(trailing) then "TRAILING:~{trailing}" else ""} \
+        SLIDINGWINDOW:~{window_size}:~{window_quality} \
+        MINLEN:~{minlen} \
+        ~{if defined(crop) then "CROP:~{crop}" else ""} \
+        ~{if defined(headcrop) then "HEADCROP:~{headcrop}" else ""} \
+        ; mv output_forward.fq.gz ~{outfile_name_forward} \
+        ; mv output_reverse.fq.gz ~{outfile_name_reverse}
```
a-frantz, line 136: Just use the exact outfile names directly in the trimmomatic command instead of renaming the files

tools/trimmomatic.wdl
a-frantz: Actually, since it looks like you can remove this `ls` and the `mv`s at the bottom, this _will_ just be one Bash statement eventually. In which case `set -euo pipefail` won't do anything. You can leave it out, but still recommend setting that for anything more complicated.

tests/tools/input_json/trimmomatic_trim_single_end.json
```diff
@@ -0,0 +1,11 @@
+{
+    "input_file": "tests/tools/input/subsampled_output_R1_100000.fastq.gz",
+    "leading": 3,
+    "trailing": 3,
+    "threads": 10,
+    "window_size": 4,
+    "window_quality": 15,
+    "minlen": 36,
+    "illumina_clip": "TruSeq3-SE:2:30:10",
```
a-frantz, line 9: Are these the defaults of `trimmomatic`? You should specify the defaults in the `input` block so that users don't need to specify all this (unless they want to override the defaults).

tools/trimmomatic.wdl
```diff
@@ -0,0 +1,151 @@
...
+        description: "Performs single end trimming of reads using the Trimmomatic tool."
+        outputs: {
+            trimmed_reads: "The trimmed single-end reads in FASTQ format."
+        }
+    }
+
+    parameter_meta {
+        input_file: "Input FASTQ format file to run Trimmomatic on."
+        output_name: "Output FASTQ format file of trimmed reads."
+        illumina_clip: "Cut adapter and other illumina-specific sequences from the read."
```
a-frantz, line 14: Can you explain this param more? What syntax does trimmomatic expect?

mjgattas: I will admit I don't fully understand what this adapter is, but there's more information about the Adapter on [the trimmomatic documentation]( http://www.usadellab.org/cms/?page=trimmomatic). It appears that this is an optional argument, so I can modify the wdl task accordingly.

The docs also mention: "Trimming occurs in the order which the steps are specified on the command line. It is recommended in most cases that adapter clipping, if required, is done as early as possible." Given this, I think the adapter cutting should be applied before other parameters, as it currently is in the wdl tasks.

tests/tools/input_json/trimmomatic_trim_single_end.json
mjgattas: Since I'm unfamiliar with the standards for trimming these read files, I'm not sure if these are "defaults" per se, but they are values used in the documentation's examples. Would you like me to make them defaults, or should we consult with end users to see what the most likely defaults would be?

tools/trimmomatic.wdl
```diff
@@ -0,0 +1,77 @@
...
+        Int window_quality
+        Boolean? phred64 = false
+    }
+
+    command <<<
+        ls -l /usr/local/share/trimmomatic-0.36-5/adapters/
+        trimmomatic ~{if defined(read_two) then "PE" else "SE"} -threads ~{threads} \
+        -trimlog trimlog.txt ~{read_one} ~{if defined(read_two) then "~{read_two}" else ""} \
+        ~{outfile_name_one} ~{outfile_name_two} \
+        ~{if defined(phred64) then "-phred64" else "-phred33"} \
```
a-frantz, line 55:
```suggestion
        ~{if (phred64) then "-phred64" else "-phred33"} \
```
`phred64` will always be defined

tools/trimmomatic.wdl
```diff
@@ -0,0 +1,77 @@
...
+                basename(read_one, ".fastq.gz") + ".forward.trimmed.fastq.gz"
+            else
+                basename(read_one, ".fastq.gz") + ".trimmed.fastq.gz"
+        )
+        String outfile_name_two = (
+            if defined(read_two) then
+                basename(read_two, ".fastq.gz") + ".reverse.trimmed.fastq.gz"
+            else
+                ""
+        )
```
mjgattas, line 37: @a-frantz this block of code is giving me trouble, causing this issue:
`(tools/trimmomatic.wdl Ln 34 Col 26) Expected String instead of File?; for basename argument #1 basename(read_two, ".fastq.gz") + ".reverse.trimmed.fastq.gz"`

I'm confused, because a similar passing of a File into basename on line 28 does not cause any errors. Curious if you know how I should handle the optional `outfile_name_two` variable assignment

a-frantz: Is that the error message from `sprocket` or `miniwdl`? Either way it's misleading/incorrect.

The issue is that `read_two` is optional, and must be coerced to a required param before being used in the `basename` func. But this section is going to be rewritten once I get the REGEX to you, so don't worry about it for now

mjgattas:
@a-frantz @adthrasher Currently, this branch is successfully returning single-end and paired-end output using the tool! I think the remaining questions entail:

- Ensuring proper naming conventions for inputs/files throughout the task
- Enabling easiest input of the cut adapter, plus adapter parameters
- Ensuring I incorporate the best tests/test-files back into the workflows repository (I'll need some guidance on the best practices for this)

tools/trimmomatic.wdl
```diff
@@ -0,0 +1,72 @@
...
+        File read_one
+        File? read_two
+        String? illumina_clip
+        String prefix = sub(
+            basename(read_one),
+            "(([_.][rR](?:ead)?[12])((?:[_.-][^_.-]*?)*?))?\\.(fastq|fq)(\\.gz)?$",
+            ""  # Once replacing with capturing groups is supported, replace with group 3
+        )
+        String outfile_name_one = "~{prefix}.R1.trimmed.fastq.gz"
+        String outfile_name_two = "~{prefix}.R2.trimmed.fastq.gz"
```
a-frantz, line 33: Move these to a private declaration (between the `input` block and the `command` block). Also rename them ot `String read_one_name` and `read_two_name`.

tools/trimmomatic.wdl
a-frantz: Delete this `ls` now that you've shared the results of it

tools/trimmomatic.wdl
```diff
@@ -0,0 +1,72 @@
...
+        -trimlog trimlog.txt ~{read_one} ~{if defined(read_two) then "~{read_two}" else ""} \
+        ~{outfile_name_one} ~{if defined(read_two) then "~{outfile_name_two}" else ""} \
+        ~{if defined(phred64) && phred64 then "-phred64" else "-phred33"} \
+        ~{if defined(illumina_clip) then "ILLUMINACLIP:~{illumina_clip}" else ""} \
+        ~{if defined(leading) then "LEADING:~{leading}" else ""} \
+        ~{if defined(trailing) then "TRAILING:~{trailing}" else ""} \
+        SLIDINGWINDOW:~{window_size}:~{window_quality} \
+        MINLEN:~{minlen} \
+        ~{if defined(crop) then "CROP:~{crop}" else ""} \
+        ~{if defined(headcrop) then "HEADCROP:~{headcrop}" else ""}
```
a-frantz, line 57: Increase the indent by one level for this block

a-frantz (commented):
Me and Andrew are discussing what the new `input` block should look like to fully support this funky API. Will get back to you over slack 👍

PR #193, opened 2024-12-16
chore: consistently use .R[1,2] instead of _R[1,2]

a-frantz:


adthrasher (dismissed):

tools/util.wdl
```diff
@@ -780,7 +780,7 @@ task split_fastq {
             stream: true,
         }
         reads_per_file: "Number of reads to include in each output FASTQ file"
-        prefix: "Prefix for the FASTQ file. The extension `.fq.gz` will be added."
+        prefix: "Prefix for the FASTQ file. The extension `.fastq.gz` will be added."
```
a-frantz, line 783: @adthrasher I'm realizing this description isn't exactly right. Should it be:
```suggestion
        prefix: "Prefix for the FASTQ file. The extension `<number>.fastq.gz` will be added."
```
The name is determined by the `split` tool, which is a kind of opaque

a-frantz:
```suggestion
        prefix: "Prefix for the FASTQ files. The extension `<split number>.fastq.gz` will be added."
```

adthrasher:
```suggestion
        prefix: "Prefix for the FASTQ file. A split index and the extension `.fastq.gz` will be added."?
```

a-frantz: That reads to me like "a split index" is a separate file...
```suggestion
        prefix: "Prefix for the FASTQ files. The extension `.fastq.gz` (preceded by a split index) will be added."
```
?

adthrasher: That works for me

adthrasher (approved):

PR #203, opened 2025-01-28
feat: docker re-architecture

a-frantz:

Finally get rid of the terrible `{VERSION}/` directories we were using for Docker versioning. Prefer a `package.json` in each docker dir.

Also adds a `developer_scripts/` directory which currently just has one script. It works on OSX but I believe should be portable with GNU `sed` (I hope at least).

~STILL ON MY WISHLIST: we should be deleting the branch specific tags on merge.~ thanks Andrew for pushing this feat!

- [ ] You have added an entry in any releavant CHANGELOGs (when appropriate).


a-frantz:
@adthrasher just temporarily have this set to use `ci_python` as the base branch to reduce `diff` clutter. I'll wait till you get that working and merged, then rebase and point to `main`

.github/workflows/docker-build.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
           fetch-depth: 0
       - name: set matrix
         id: set-matrix
-        run: echo "images=$(find docker -maxdepth 2 -mindepth 2 | jq --raw-input --slurp --compact-output 'split("\n")[:-1]')" >> $GITHUB_OUTPUT
+        run: echo "images=$(find docker -maxdepth 1 -mindepth 1 | jq --raw-input --slurp --compact-output 'split("\n")[:-1]')" >> $GITHUB_OUTPUT
```
a-frantz, line 21: I think this line is overengineered at this point. Should this be rewritten somehow?

.github/workflows/docker-build.yaml
```diff
@@ -27,29 +27,49 @@ jobs:
...
+        run: |
+          remote_digest=$(docker inspect ${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.TOOL }}:${{ env.IMAGE_TAG }} --format='{{index .RepoDigests 0}}')
+          local_digest=${{ steps.build.outputs.digest }}
+          if [ "$remote_digest" != "$local_digest" ]; then
+            echo "Digests do not match!"
+            echo "Remote: $remote_digest"
+            echo "Local: $local_digest"
+            echo "Should you increment the version number?"
+            exit 1
+          fi
```
a-frantz, line 71: This has not run yet (because it requires being on `main`) so could use some extra scrutiny

a-frantz:
> > Also adds a `developer_scripts/` directory which currently just has one script. It works on OSX but I believe should be portable with GNU `sed` (I hope at least).
> 
> Can you add this to the directory list in the README?

Already did. Looks like the commit and the comment went up at the same time lol

.github/workflows/docker-build.yaml
```diff
@@ -27,29 +27,60 @@ jobs:
...
+          if [ ${{ env.REVISION }} != "null" ]; then
+            echo "IMAGE_TAG=branch-${{ env.GH_REF }}-${{ env.SEMVER }}-${{ env.REVISION }}" >> $GITHUB_ENV
+          else
+            echo "IMAGE_TAG=branch-${{ env.GH_REF }}-${{ env.SEMVER }}" >> $GITHUB_ENV
+          fi
       - name: Log in to the Container registry
         run: echo ${{ secrets.GITHUB_TOKEN }} | docker login ${{ env.REGISTRY }} --username ${{ github.actor }} --password-stdin
       - uses: webfactory/ssh-agent@v0.7.0
         with:
           ssh-private-key: ${{ secrets.STJUDE_SEQERR_READONLY_DEPLOY_KEY }}
```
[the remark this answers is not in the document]
a-frantz: I tried to remove it and got hit with a bunch of errors about the `ssh` socket so put it back 😅 

We are not using/building `SEQERR` anymore,. but I think the `ssh-agent` still needs to be set? I'll be honest I don't really understand the errors I was seeing lol.

.github/workflows/docker-build.yaml
a-frantz: So right now `IMAGE_TAG` gets set dynamically. I think to do what you're suggesting here, we would need to have access to a `MAIN_IMAGE_TAG` and a `BRANCH_IMAGE_TAG`. Then `remote_digest` should get set to the digest of `MAIN_IMAGE_TAG`. If the digests differ, we error and complain?

a-frantz: Pushed something up. Let's see how if it runs correctly!

.github/workflows/docker-build.yaml
kevin-benton: I do something very similar [here](https://github.com/stjudecloud/cicd-workflows/blob/main/.github/workflows/app-nightly-scan.yaml#L19) with the difference being using `ls docker` would drop the `docker/` prefix. Haven't looked farther to see if that is used.

But the `jq` piece is good.

.github/workflows/docker-build.yaml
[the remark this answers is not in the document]
a-frantz: @kevin-benton this convo is out of date now, so tagging you so you see it. We  do not need the `SEQERR` secret anymore, but removing it causes errors I don't understand

.github/workflows/docker-build.yaml
a-frantz: Ok I think we _do_ use the `docker` prefix, so I propose we leave this for now.

.github/workflows/docker-clean.yaml
```diff
@@ -0,0 +1,48 @@
+on:
+    pull_request:
+      types:
+        - closed
+
+env:
+  REGISTRY: ghcr.io
+
+jobs:
+    list-images:
```
a-frantz, line 10: I don't know how to fix this, but this will leave some stragglers if the image tags were updated during the PR process (which will be most of the time).

I think we may want to add a manual workflow that just deletes any tag starting with `branch-` that we run occasionally. That will be the easiest solution at any rate.

.github/workflows/docker-clean.yaml
```diff
@@ -0,0 +1,48 @@
...
+        steps:
+          - name: checkout
+            uses: actions/checkout@v3
+            with:
+                fetch-depth: 0
+          - name: set matrix
+            id: set-matrix
+            run: echo "images=$(find docker -maxdepth 1 -mindepth 1 | jq --raw-input --slurp --compact-output 'split("\n")[:-1]')" >> $GITHUB_OUTPUT
+    if_merged:
+        if: github.event.pull_request.merged == true
```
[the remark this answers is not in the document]
adthrasher: True, since it builds the branch images on creation, we don't necessarily need a merged PR. Any closed PR should be eligible for deletion.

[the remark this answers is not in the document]
a-frantz: Is there a trigger action for just "any branch deletion"?

.github/workflows/docker-clean.yaml
adthrasher: OK, I'm now using the GitHub API to get the IDs of any package with the `branch-<id>` prefix and any images without a tag. All of those are passed through for deletion.

.github/workflows/docker-clean.yaml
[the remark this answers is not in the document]
adthrasher: There is a `delete` action, which you could then check specifically that it is a `branch`.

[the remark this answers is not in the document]
a-frantz: Maybe we should be using that trigger instead. Because the build step is triggered for _every_ push regardless of branch/PR status.

.github/workflows/docker-clean.yaml
a-frantz: Looks great!

kevin-benton (approved):

PR #208, opened 2025-01-30
ci: pytest action waits on build action and remove SEQERR reference

a-frantz:

`pytest` action will now wait for `build` to complete. Also remove the lingering `SEQERR` references in our CI.
~Also, renamed `scripts/` to `docker_scripts/`~ and added it to the README


.github/workflows/pytest.yaml
```diff
@@ -1,6 +1,11 @@
 name: pytest-workflow Check
 
-on: [push]
+# Only trigger, when the build workflow succeeded
```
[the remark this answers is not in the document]
a-frantz: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run

.github/workflows/docker-build.yaml
```diff
@@ -56,28 +56,25 @@ jobs:
...
-        with:
-          ssh-private-key: ${{ secrets.STJUDE_SEQERR_READONLY_DEPLOY_KEY }}
       - uses: docker/setup-qemu-action@v2
       - uses: docker/setup-buildx-action@v2
       - uses: docker/build-push-action@v4
         id: build
         with:
           context: ${{ matrix.image }}
-          build-contexts: scripts=./scripts
+          build-contexts: scripts=./docker_scripts
```
adthrasher, line 65: I actually don't like this change. Part of the reason that we removed embedded scripts was to make it possible for users (such as Charlie) to run them separately. So they're decoupled from the WDL and the Docker images, even though they get included in the docker images.

docker/python-plotting/package.json
```diff
@@ -1,4 +1,4 @@
 {
     "name": "python-plotting",
-    "version": "1.2.0"
+    "version": "1.2.1"
```
adthrasher, line 3: I wonder if these images that are decoupled from an underlying tool version should scrap SemVer and just be integer versioned.

.github/workflows/docker-build.yaml
a-frantz: Ok, that sounds fine to me. I'll revert these changes 👍

docker/python-plotting/package.json
a-frantz: Can you elaborate on your reasoning?

adthrasher: I'm just not convinced that SemVer's normal definitions apply to many of these changes. I don't know that many of our container updates will clearly fit that format. So it may be simpler to just increment an integer release. Even this PR, I'm not really sure it warrants a version bump.

a-frantz: I'm not sure. I'd like to think about this some more. Can we postpone this decision? And merge this in the meantime? (we can continue the discussion over slack)

adthrasher: Yes, I meant to mention that previously. I'm not sure it should've gone in this PR anyway. Fine to continue the discussion later.

adthrasher (approved):

PR #211, opened 2025-01-31
Update docker images to reduce security vulnerabilities

adthrasher:
This resolves 960 security vulnerabilities. Mainly by switching compatible images to an `alpine` base image. Other improvements include updating base OS versions and updating packages. There are 25 remaining security vulnerabilities, all of which are Low/Medium by Snyk ([level documentation](https://docs.snyk.io/manage-risk/prioritize-issues-for-fixing/severity-levels)). All of the remaining vulnerabilities are in the `util` image (which uses Ubuntu 24.04). We do an `apt-get upgrade` in the image and I've skimmed most of the remaining vulnerabilities and they have no fixed version in 24.04. So the only "fix" would be to change to a different image that does not have those vulnerabilities.

One important thing to note, the Python alpine-based images (`python:3.13.1-alpine`) are all using the 3.21 release from 12/05/2024. So it's unlikely that it has *no* vulnerabilities, but rather that those have not been discovered and reported. [Here](https://security.alpinelinux.org/branch/3.20-main) is the list from the prior 3.20 release. The R image (`ghcr.io/r-hub/r-minimal/r-minimal:4.4.2`) uses the older 3.19 release with slightly more [vulnerabilities](https://security.alpinelinux.org/branch/3.19-main).

### DIFF
--- a/workflows/methylation/methylation-cohort.wdl
+++ b/workflows/methylation/methylation-cohort.wdl
@@ -209,7 +209,7 @@ task generate_umap {
     }
 
     runtime {
-        container: "ghcr.io/stjudecloud/umap:0.5.7-3"
+        container: "ghcr.io/stjudecloud/umap:0.5.7-4"
         memory: "8 GB"
         cpu: 1
         disks: "~{disk_size_gb} GB"
@@ -244,7 +244,7 @@ task plot_umap {
     }
 
     runtime {
-        container: "ghcr.io/stjudecloud/python-plotting:1.2.1"
+        container: "ghcr.io/stjudecloud/python-plotting:2.0.0"
         memory: "4 GB"
         cpu: 1
         disks: "4 GB"
--- a/workflows/methylation/methylation-cohort.wdl
+++ b/workflows/methylation/methylation-cohort.wdl
@@ -124,7 +124,7 @@ task combine_data {
     }
 
     runtime {
-        container: "ghcr.io/stjudecloud/pandas:2.2.1-3"
+        container: "ghcr.io/stjudecloud/pandas:branch-update_images-2.2.1-3"
         memory: "~{memory_gb} GB"
         cpu: 1
         disks: "~{disk_size_gb} GB"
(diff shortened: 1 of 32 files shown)


adthrasher:
Tagging @mcrusch and @claymcleod so they're aware. Specifically, since Mike and I discussed security vulnerabilities last week.

claymcleod (approved):
Seems reasonable to me.

workflows/methylation/methylation-cohort.wdl
```diff
@@ -199,7 +199,7 @@ task generate_umap {
     Int disk_size_gb = ceil(size(filtered_beta_values, "GiB") * 2) + 2
 
     command <<<
-        python $(which generate_umap.py) \
+        python /usr/local/bin/generate_umap.py \
```
a-frantz, line 202: Shouldn't the script be on `$PATH`? Can we avoid using the full path? (Asking purely for aesthetic reasons, not a big deal at all if the full path is needed).

workflows/methylation/methylation-cohort.wdl
```diff
@@ -236,15 +236,15 @@ task plot_umap {
     }
 
     command <<<
-        python $(which plot_umap.py) --umap ~{umap} --output-name ~{plot_file}
+        python /usr/local/bin/plot_umap.py --umap ~{umap} --output-name ~{plot_file}
```
a-frantz, line 239: Same Q

a-frantz (dismissed):
Approving even though when you switch the `container` values back to main my review will be dismissed

workflows/methylation/methylation-cohort.wdl
adthrasher: `/usr/local/bin` should be on `$PATH`. `generate_umap.py` is executable but lacks a shebang line, so it can't actually be called directly currently. Calling with `python` doesn't follow `$PATH` resolution.

adthrasher: Personally, I prefer using full paths for fixed location items in Docker containers (as evidenced by this command).

a-frantz: Ah gotchya. Ok. Fine with that. Go ahead and run `update_container_values.sh` and I'll approve.

Issue #189, opened 2024-11-26
bumping a docker image necessitates failing CI before merge


a-frantz:
The process for bumping any of our built docker images involves testing them using a `branch-*` specific image that is built as part of our CI. While testing using these images, the `lint-check.yaml` CI will fail because the tag isn't a valid production value.

We don't want to build the production-ready image as part of branch development, but to get the `lint-check` CI passing we have to reference tags that don't exist yet. So at least the last commit on a PR that bumps one of these docker images is going to be pointing at a non-existent tag (so the `pytest` tests will fail).

Is there a way we can re architect our CI so that the last commit can be green?

a-frantz:
Check this comment for a breakdown of what happened to the CI in a recent PR - https://github.com/stjudecloud/workflows/pull/187#issuecomment-2501211772

a-frantz:
@adthrasher 
> I think you can write an action that runs on `merge` and you could simply add another commit that deletes the `branch-*` portion of the container entries.
> Another option would be to append a step to the docker image builds that updates the tags in the WDL files to whatever was just built.


Andrew Frantz
> Both of those sound possible. The first option would probably use a lot of logic we would need for _deleting_ the `branch-*` images, which we currently don't do. I'm surprised we haven't run into issues with ghcr.io with how many tags we throw up there.

PR #218, opened 2025-02-13
refactor: add container-diff

adthrasher:
Update the digest check for container differences. Since adding the digest check, we've encountered issues and false positives. `container-diff` provides a deeper (hopefully more robust) check.

- [ ] You have added yourself or the appropriate individual as the assignee.

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -81,8 +81,13 @@ jobs:
             echo "Remote: $remote_digest"
             echo "Local: $local_digest"
             echo "Should you increment the version number?"
-            exit 1
+            # exit 1
           fi
+      - name: container diff
+        run: |
+          curl -LO https://storage.googleapis.com/container-diff/latest/container-diff-linux-amd64 && \
+          chmod +x container-diff-linux-amd64 && \
+          ./container-diff-linux-amd64 diff daemon://${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.TOOL }}:${{ env.TAG_TO_CHECK }} daemon://${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ env.TOOL }}:local --type=history --type=file --type=size
       - name: Run Snyk to check Docker image for vulnerabilities
         # Snyk can be used to break the build when it detects vulnerabilities.
         # In this case we want to upload the issues to GitHub Code Scanning


a-frantz:
The new check is failing for the `minfi` image. Any idea why that it is? I don't see any changes to it in the "files changed" tab.

adthrasher:
> The new check is failing for the `minfi` image. Any idea why that it is? I don't see any changes to it in the "files changed" tab.

Yeah, it's correct. This commit (https://github.com/stjudecloud/workflows/commit/5af65ae26e15cbd7852a07bcbe1517fbc5c22319#diff-65053fa5a6e57a96b08ea68284f14b607cf161e2efbcfb602ef6f74c97af0c2b) reformatted the R script, but didn't bump the container revision. I'm not convinced that a reformat needs a revision bump, but the check correctly finds that change.

a-frantz (approved):

a-frantz (changes requested):
oop sorry. You need to update the STAR and `minfi` container values

docker/star/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM debian:10 AS python
+FROM debian:12.5-slim AS python
```
a-frantz, line 1: Has this change been validated in `pytest`?

adthrasher: I didn't do anything special, but the `pytest` check was passing.

a-frantz: I don't see in the git history any tests that were pointing to this branch's built image

adthrasher: Good point. Now I'm wondering if the `pytest` check should automatically change the container entries to use the branch images on PR tests.

a-frantz: That might solve another problem we have as well! I think that would let the "last commit" in a PR pass the check (issue outlined in #189 )

developer_scripts/update_container_tags.sh
```diff
@@ -9,8 +9,8 @@ if [[ "$target" == "main" ]]; then
   search_pattern='container: "ghcr\.io/stjudecloud/([^:]*):branch-[^-]*-(.*)"'
   replacement='container: "ghcr.io/stjudecloud/\1:\2"'
 else
-  search_pattern='container: "ghcr\.io/stjudecloud/([^:]*):(.*)"'
-  replacement='container: "ghcr.io/stjudecloud/\1:branch-'$target'-\2"'
+  search_pattern='container: "ghcr\.io/stjudecloud/([^:]*):(branch-'$target'-)?(.*)"'
```
adthrasher, line 12: @a-frantz - This isn't quite perfect, but will prevent the current branch from being inserted multiple times. It will break if a different branch tag is present, but I think that's OK for now.

a-frantz: Without having testing it locally, this looks fine. Agree that it's ok to break in that case

adthrasher: I actually realized that you're already doing the check I want under `main`. So I just copied the regex from above `branch-[^-]*-(.*)`

a-frantz:
@adthrasher I think that with the CI changes, we should have green CI now. IDK why the `methylation` pytest is failing. But aside from that these changes look good.

adthrasher:
> @adthrasher I think that with the CI changes, we should have green CI now. IDK why the `methylation` pytest is failing. But aside from that these changes look good.

In typical R fashion, the error message is unhelpful.
```
Error in h(simpleError(msg, call)) : 
  error in evaluating the argument 'x' in selecting a method for function 'colnames': object 'CN' not found
Calls: colnames -> .handleSimpleError -> h
Execution halted
```

a-frantz (approved):

PR #230, opened 2025-05-22
chore: add trailing commas

adthrasher:
Enabling the `TrailingCommas` lint now that `miniwdl` supports the syntax. Updating the WDL files to use trailing commas.

### DIFF
--- a/.github/workflows/sprocket-lint.yaml
+++ b/.github/workflows/sprocket-lint.yaml
@@ -14,4 +14,4 @@ jobs:
             exclude-patterns: template
             deny-warnings: true
             deny-notes: true
-            except: TrailingComma,ContainerUri
+            except: ContainerUri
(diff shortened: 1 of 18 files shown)


adthrasher:
@a-frantz - Any thoughts on disabling the `Shellcheck` rule in our CI for now?

a-frantz:
> @a-frantz - Any thoughts on disabling the `Shellcheck` rule in our CI for now?

I'd like us to actually address the shellcheck lints, but I don't have the bandwidth to deal with it in the immediate future. It would be good not only for the robustness of this repo, but also would be informative about ways we might be able to improve the rule in `wdl-lint` (if it has a problematic number of false-positives or something similar to that)

So I guess my answer to this question is dependent on how soon you think one of us can dig into these lints? If it will be a while as it's low on both our priority lists, I'd be ok temporarily disabling it in the interest of a clean CI. But ideally we can work on it sooner rather than later

a-frantz (approved):

Issue #166, opened 2024-07-15
support Single-Ended data


a-frantz:
All of our workflows and (almost) all of our tasks assume that data is Paired-End. SE support would make our workflows and tools more accessible.

a-frantz:
STAR should now work for SE data - #202

a-frantz:
This was discussed is some detail somewhat recently. It was decided that this will not pursued until we actually have a real use case for supporting SE data. Closing issue as not planned.

PR #235, opened 2025-07-08
[WIP] feat: fixes and improvements around RGs

a-frantz:

- [ ] You have added yourself or the appropriate individual as the assignee.


a-frantz:
note to self: CHANGELOGs before merge!

workflows/rnaseq/rnaseq-core.wdl
```diff
@@ -124,9 +124,9 @@ workflow rnaseq_core {
     input {
         File gtf
         File star_db
-        Array[File] read_one_fastqs_gz
+        Array[File]+ read_one_fastqs_gz
         Array[File] read_two_fastqs_gz
```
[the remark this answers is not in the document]
a-frantz: We support Single-End! I think it works at least 🫣

a-frantz:
Something else to note that I'm recording for posterity:

STAR and `rnaseq-core` work for Single-End FASTQs now. However that has not been exposed at the `rnaseq-standard[-fastq]` level, as it requires weird and annoying WDL constructions. Importantly, I'm not aware of any SE RNA-Seq data we have to process, so even if it were enabled we would not be able to functionally test it.

Linking to #166

tools/star.wdl
```diff
@@ -512,7 +512,7 @@ task alignment {
         File star_db_tar_gz
         Array[File]+ read_one_fastqs_gz
         Array[String]+ read_groups
-        Array[File] read_two_fastqs_gz = []
+        Array[File]+? read_two_fastqs_gz
```
adthrasher, line 515: Interesting. Do the engines interpret this correctly? I'm assuming it should read as "optional, but if present, it must not be empty".

workflows/chipseq/CHANGELOG.md
adthrasher, line 1: I don't see these changes in this PR.

tools/star.wdl
a-frantz: Yes, it does seem to work with both miniwdl and sprocket

workflows/chipseq/CHANGELOG.md
a-frantz: They're there? Instead of `util.get_read_groups`, (no validation), we use the tasks from `data_structures/read_group.wdl` that have validation "baked in". `after` clauses are there. And `ngsderive` is gone.

adthrasher: User error. I forgot I was only viewing changes since my last review, which only included the CHANGELOG, not the chipseq workflow.

PR #253, opened 2025-08-14
chore: clean up symlinks after execution

a-frantz:

noticed while working on #251 that we had left work dirs polluted with symlinks. This PR just adds some `rm` statements to clean up after ourselves.

### DIFF
--- a/data_structures/flag_filter.wdl
+++ b/data_structures/flag_filter.wdl
@@ -107,7 +107,7 @@ task validate_string_is_12bit_int {
     >>>
 
     runtime {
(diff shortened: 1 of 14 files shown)


a-frantz:
I think that's possible, but on the other hand, I think _most_ of our tasks that create symlinks are already `rm`ing them and I don't think we've ever had an issue or confusion with any of those.

Currently we are inconsistent, I have a slight preference for cleaning up but would also be fine leaving them. In which case we should go through other tasks that use symlinks and remove the clean up code

adthrasher (approved):

PR #263, opened 2025-09-03
ci: drop pytest in favor of Sprocket test

a-frantz:
~~This PR is for demonstration purposes of the RFC document [here](https://stjude-rust-labs.github.io/rfcs/branches/rfcs/sprocket-test/0001-sprocket-test.html)~~

this PR has morphed from a demonstration into a real PR for swapping our test infra from pytest to Sprocket. Still a WIP as not all the tests are written yet and it has not been hooked into CI yet


a-frantz:
all tests passing 🎉 
@adthrasher for local testing, I `picard::mark_duplicates` uses more memory than available on my MPB, so I've annotated that test with `high_mem`. Running `sprocket dev test -f high_mem` at the root of this branch should be all green. I'm seeing all green on the cluster as well. For executing tests there,  try a `parallelism = 300` to get a good balance of speed and resource utilization. See https://github.com/stjude-rust-labs/sprocket/issues/821 for why just lifting the parallelism throttle completely doesn't work

.github/workflows/build-and-test.yaml
adthrasher, line 1: I haven't gotten further, but my assumption was that we'd call `sprocket dev test` as the second part of this to test the newly built images.

.github/workflows/sprocket-lint.yaml
```diff
@@ -3,12 +3,17 @@ name: Workflows Sprocket Lint
...
-    sprocket_lint:
-        runs-on: ubuntu-latest
-        steps:
-        - uses: actions/checkout@v2
-        - name: Run sprocket
-          uses: stjude-rust-labs/sprocket-action@main
-          with:
-            lint: true
-            except: KnownRules
+  sprocket_lint:
```
adthrasher, line 6: If we're doing `lint` and `format`, I'd probably just shorten this to `sprocket`.
```suggestion
  sprocket:
```

adthrasher: Or maybe lengthen to `sprocket_lint_format`

tools/picard.wdl
```diff
@@ -200,9 +201,6 @@ task validate_bam {
     String mode_arg = if summary_mode
         then "--MODE SUMMARY"
         else ""
-    String stringency_arg = if index_validation_stringency_less_exhaustive
-        then "--INDEX_VALIDATION_STRINGENCY LESS_EXHAUSTIVE"
-        else ""
```
adthrasher, line 205: I don't recall what this parameter does. Why are we removing the argument?

adthrasher:
Note to self: GATK4 tests work locally

adthrasher (approved):

PR #266, opened 2025-09-15
chore: python heredoc cleaning

a-frantz:

ports remaining python heredocs to full scripts


scripts/util/calc_tpm.py
```diff
@@ -0,0 +1,45 @@
+def main(counts_name_path, gene_lengths_path, outfile_path, has_header):
+    counts_file = open(counts_name_path, "r")
+    counts = {}
+    if has_header:
+        counts_file.readline()
```
adthrasher, line 5: Can this be auto-detected?

scripts/util/calc_tpm.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+if __name__ == "__main__":
+    import argparse
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("counts_path")
+    parser.add_argument("gene_lengths_path")
+    parser.add_argument("outfile_path")
+    parser.add_argument("--has_header", action="store_true")
```
adthrasher, line 42: Might want to rename this, since `gene_lengths_path` is expected to always have a header. This is only applied to `counts_path`.

scripts/util/calc_tpm.py
a-frantz: if we are comfortable assuming:
> if row 1 column 2 is not an `int` it has a header

The corner case is if someone provides a sample name like `1234`

scripts/util/calc_tpm.py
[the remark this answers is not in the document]
a-frantz: ah! That's a good thought. I have been assuming that the "util" in `scripts/util/*py` was referring to the image the script belonged to, but that it not how you set up `scripts/methylation/*`. I don't have strong feelings on this architecture question, but I'm leaning towards breaking these out by `tool/*wdl` name?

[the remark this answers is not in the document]
adthrasher: That sounds fine to me. There's also the `data_structures`, but I think `data_structures/*.wdl` is fine. If there is a name conflict, we can revisit.

adthrasher (approved):

PR #282, opened 2025-12-03
feat: aligners & variant callers

adthrasher:
Adds new WDL task implementations:

- bwa-mem2
- minimap2
- ~~hisat2~~
- vg giraffe
- DeepVariant / DeepSomatic
- Strelka
- Manta
- Clair3
- NGSEP

I also created a new evaluation directory under workflows (maybe it should be placed somewhere else) that has workflows for running a sample through multiple analyses in parallel for comparison purposes.


adthrasher:
Lint failures are for EmptyOutputs, which is addressed in #315

stjudecloud-cloudy:
### :white_check_mark: **Snyk checks have passed. No issues have been found so far.**

| Status | Scan Engine |  Critical |  High |  Medium |  Low | Total (0) |
|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/compbio/pr-checks/6de0fd09-de51-4d40-880e-8d16212a8323?refs=d00557fe40819209ba89e4c688aef92898d27de1&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/compbio/pr-checks/6de0fd09-de51-4d40-880e-8d16212a8323/license?refs=d00557fe40819209ba89e4c688aef92898d27de1&source=prChecksComment) |

:computer: Catch issues earlier using the plugins for [VS Code](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner), [JetBrains IDEs](https://plugins.jetbrains.com/plugin/10972-snyk-security), [Visual Studio](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner-vs-2022), and [Eclipse](https://marketplace.eclipse.org/content/snyk-security#external-install-button).

.github/workflows/docker-build.yaml
a-frantz, line 1: lets delete those lines instead of commenting them out

data_structures/read_group.wdl
```diff
@@ -425,3 +425,51 @@ task inner_read_group_to_string {
         maxRetries: 1
     }
 }
+
+task read_group_to_array {
```
a-frantz: can this be consolidated with `inner_read_group_to_string`? A new `Boolean split_on_field` input and mutually exclusive outputs?
```wdl
    output {
        String? stringified_read_group = <>
        Array[String]? read_group_array = <>
    }
```

data_structures/read_group.wdl
a-frantz, line 1: Could we include in this PR the fix in the preamble that borks the syntax highlighting? I think it's incorrectly parsing a JSON codeblock in the preamble.

docker/mako/Dockerfile
a-frantz, line 1: there isn't an official image for this?

tools/bwa.wdl
```diff
@@ -236,14 +236,14 @@ task bwa_mem {
     )
 
     command <<<
-        set -euo pipefail
+        set -xeuo pipefail
```
a-frantz:
```suggestion
        set -euo pipefail
```
unless needed for some reason I'm unaware of

tools/bwamem2.wdl
```diff
@@ -0,0 +1,127 @@
...
+        String read_group
+        File? read_two_fastq_gz
+        String prefix = sub(basename(read_one_fastq_gz), "([_\\.][rR][12])?(\\.subsampled)?\\.(fastq|fq)(\\.gz)?$",
+            "")
+        Boolean smart_pairing = false
+        Boolean skip_mate_rescue = false
+        Int threads = 4
+        Int modify_disk_size_gb = 0
+        Int seed_length = 19
+        Int min_score = 30
```
a-frantz: nit:
```suggestion
        Int seed_length = 19
        Int min_score = 30
        Int ncpu = 4
        Int modify_disk_size_gb = 0
```

tools/clair.wdl
```diff
@@ -0,0 +1,269 @@
...
+
+    output {
+        File pileup_vcf = "~{output_dir}/pileup.vcf.gz"
+        File full_alignment_vcf = "~{output_dir}/full_alignment.vcf.gz"
+        File merged_vcf = "~{output_dir}/merge_output.vcf.gz"
+    }
+
+    requirements {
+        # container: "quay.io/biocontainers/clair3:1.2.0--py310h779eee5_0"
+        container: "hkubal/clair3:v2.0.0"
```
a-frantz: what's the difference between these?

tools/clair.wdl
```diff
@@ -0,0 +1,269 @@
...
+        String model
+        File? bed_regions
+        File? vcf_candidates
+        Array[String] contigs = []
+        String output_dir = "clair3_output"
+        String platform = "ilmn"
+        Boolean all_contigs = false
+        Boolean print_ref_calls = false
+        Boolean gvcf = false
+        Int threads = 4
```
a-frantz:
```suggestion
        Int ncpu = 4
```
consistency with rest of repo

tools/clair.wdl
```diff
@@ -0,0 +1,269 @@
...
+        set -euo pipefail
+
+        ref_fasta=~{basename(reference_fasta, ".gz")}
+        gunzip -c "~{reference_fasta}" > "$ref_fasta" \
+            || ln -sf "~{reference_fasta}" "$ref_fasta"
+        ln -sf "~{reference_fasta_index}" "$ref_fasta.fai"
+
+        # Clair-3 resolves the BAM path...
+        cp "~{bam}" "~{filename}"
+        cp "~{bam_index}" "~{filename}.bai"
```
a-frantz: what does this mean? It doesn't like symlinks?

tools/hisat2.wdl
a-frantz, line 1: I would rather not include something that is immediately deprecated. However there's probably value in keeping this in the history in some way. Maybe pull it out into a branch without a PR associated?

tools/manta.wdl
```diff
@@ -0,0 +1,199 @@
+version 1.2
```
a-frantz: should we bring this up to 1.3 while we're at it?

tools/ngsderive.wdl
a-frantz, line 1: where did the `+ 30` come from?

tools/picard.wdl
```diff
@@ -587,7 +601,7 @@ task collect_alignment_summary_metrics {
     }
 
     Float bam_size = size(bam, "GB")
-    Int disk_size_gb = ceil(bam_size) + 10 + modify_disk_size_gb
+    Int disk_size_gb = ceil(bam_size) + 30 + modify_disk_size_gb
```
a-frantz, line 604: where's the 30 from here?

tools/samtools.wdl
```diff
@@ -664,7 +664,7 @@ task addreplacerg {
     }
 
     Float bam_size = size(bam, "GB")
-    Int disk_size_gb = ceil(bam_size * 2) + 10 + modify_disk_size_gb
+    Int disk_size_gb = ceil(bam_size * 2) + 50 + modify_disk_size_gb
```
a-frantz, line 667: oof, 50?

a-frantz (commented):
I blasted through this quickly just to kick off some questions I have. More thorough review in the works, but wanted to keep the ball rolling.

docker/mako/Dockerfile
adthrasher: Ah, there is. `fg-mako` which my search for `mako` probably didn't return.

tools/bwa.wdl
adthrasher: Stray debugging change.

tools/clair.wdl
adthrasher: If I recall correctly, the biocontainers images were broken. So I switched to the source repo's images. I'll delete the commented out one.

tools/clair.wdl
adthrasher: Yes, I don't recall the specifics now, but the tool breaks on symlinks.

tools/ngsderive.wdl
adthrasher: When running in Azure, these tasks were all unreliable due to disk sizing. We weren't leaving enough overhead in our calculations. I don't recall if I tried `+ 20` or just jumped to `+ 30` though.

data_structures/read_group.wdl
adthrasher: Should be done. Can you verify that it looks correct for you as well?

a-frantz: Hmm. it's now all dark grey text. An improvement over the bright red, but still not right. Try deleting the whole JSON code block maybe? We really don't need it there for any reason, now that [sprocket] tooling has improved in error messaging. Anyone should be able to figure that out with modern tooling.

tools/ngsderive.wdl
a-frantz: O wow, 30 is a beefy amount of disk. I guess if that's what it takes, that's what it takes. Do you suspect that is really all system overhead? I guess it must be, as I don't know where else that much could be coming from

tools/clair.wdl
a-frantz: that's not ideal. I wonder if a hard link would work? I don't think I've ever tried using one of those in a container

data_structures/read_group.wdl
```diff
@@ -417,7 +427,8 @@ task inner_read_group_to_string {
     >>>
 
     output {
-        String stringified_read_group = read_string("out.txt")
+        String? stringified_read_group = read_string("out.txt")
+        Array[String]? read_group_array = read_lines("out.txt")
```
a-frantz:
```suggestion
        String? stringified_read_group = if split_split_on_field then None else read_string("out.txt")
        Array[String]? read_group_array = if split_on_field then read_lines("out.txt") else None
```
and then can this be plugged into the `read_group_to_string` workflow? And in `workflows/` we shouldn't access the `inner_` task

tools/clair.wdl
adthrasher: I'm not sure. I think the `inputs` and `work` are mounted as separate mounts, so a hardlink would fail trying to cross the filesystem boundary.

tools/bwamem2.wdl
```diff
@@ -0,0 +1,127 @@
...
+        File read_one_fastq_gz
+        File reference_index
+        String read_group
+        File? read_two_fastq_gz
+        String prefix = sub(basename(read_one_fastq_gz), "([_\\.][rR][12])?(\\.subsampled)?\\.(fastq|fq)(\\.gz)?$",
+            "")
+        Boolean smart_pairing = false
+        Boolean skip_mate_rescue = false
+        Int seed_length = 19
+        Int min_score = 30
```
a-frantz, line 35: Are these the defaults, or have they been customized? We should document if they differ.

a-frantz: this Q would apply to any of the tool defaults below, so I won't keep asking, but doing a scan through and making sure any tweaks get recorded is good.

cspell.json
a-frantz, line 1: objection to switching to this spell checker instead of maintaining this jumbo list?
https://github.com/crate-ci/typos

I can put up a separate PR that you can merge in, deleting this file if that sounds good to you

tools/clair.wdl
adthrasher: I appreciate the push on this. The upstream tool (clair3) has been updated and now handles symlinks correctly 🎉. Once my verification tasks finish, I'll push an update.

tools/bwamem2.wdl
adthrasher: I'll double-check, but I believe everything is a default. This work was just to validate that the tools ran, so I didn't customize arguments.

adthrasher: Recording the CLI help here. These are all defaults to bwamem2.

```
Launching executable "/usr/local/bin/bwa-mem2.avx2"
-----------------------------
Executing in AVX2 mode!!
-----------------------------
* SA compression enabled with xfactor: 8
Usage: bwa-mem2 mem [options] <idxbase> <in1.fq> [in2.fq]
Options:
  Algorithm options:
    -o STR        Output SAM file name
    -t INT        number of threads [1]
    -k INT        minimum seed length [19]
    -w INT        band width for banded alignment [100]
    -d INT        off-diagonal X-dropoff [100]
    -r FLOAT      look for internal seeds inside a seed longer than {-k} * FLOAT [1.5]
    -y INT        seed occurrence for the 3rd round seeding [20]
    -c INT        skip seeds with more than INT occurrences [500]
    -D FLOAT      drop chains shorter than FLOAT fraction of the longest overlapping chain [0.50]
    -W INT        discard a chain if seeded bases shorter than INT [0]
    -m INT        perform at most INT rounds of mate rescues for each read [50]
    -S            skip mate rescue
    -o            output file name missing
    -P            skip pairing; mate rescue performed unless -S also in use
Scoring options:
   -A INT        score for a sequence match, which scales options -TdBOELU unless overridden [1]
   -B INT        penalty for a mismatch [4]
   -O INT[,INT]  gap open penalties for deletions and insertions [6,6]
   -E INT[,INT]  gap extension penalty; a gap of size k cost '{-O} + {-E}*k' [1,1]
   -L INT[,INT]  penalty for 5'- and 3'-end clipping [5,5]
   -U INT        penalty for an unpaired read pair [17]
Input/output options:
   -p            smart pairing (ignoring in2.fq)
   -R STR        read group header line such as '@RG\tID:foo\tSM:bar' [null]
   -H STR/FILE   insert STR to header if it starts with @; or insert lines in FILE [null]
   -j            treat ALT contigs as part of the primary assembly (i.e. ignore <idxbase>.alt file)
   -5            for split alignment, take the alignment with the smallest coordinate as primary
   -q            don't modify mapQ of supplementary alignments
   -K INT        process INT input bases in each batch regardless of nThreads (for reproducibility) []
   -v INT        verbose level: 1=error, 2=warning, 3=message, 4+=debugging [3]
   -T INT        minimum score to output [30]
   -h INT[,INT]  if there are <INT hits with score >80% of the max score, output all in XA [5,200]
   -a            output all alignments for SE or unpaired PE
   -C            append FASTA/FASTQ comment to SAM output
   -V            output the reference FASTA header in the XR tag
   -Y            use soft clipping for supplementary alignments
   -M            mark shorter split hits as secondary
   -I FLOAT[,FLOAT[,INT[,INT]]]
                 specify the mean, standard deviation (10% of the mean if absent), max
                 (4 sigma from the mean if absent) and min of the insert size distribution.
                 FR orientation only. [inferred]
Note: Please read the man page for detailed description of the command line and options.
```

tools/ngsderive.wdl
adthrasher: Ultimately, we should revisit this later. Once Sprocket has the ability to record disk usage, we'll have better information. The only downside is that it takes the size of `work` at the end of the run, which can undercount if intermediate files are being removed along the way.

a-frantz: > Ultimately, we should revisit this later.

I was going to ask we log it in some way, but we do have issue #88 already (from 3 years ago), so I guess we can resolve this thread.

> The only downside is that it takes the size of work at the end of the run, which can undercount if intermediate files are being removed along the way.

Do you mean the new Sprocket resource tracking you're working on? size of `work` at the end of the task will be a good start, and will be sufficient for most WDL tasks, but there are quite a lot of tasks that involve intermediate files that are cleaned up. IDK how we can really track that well, aside from digging through cloud logs. I'm not sure that would really be in the responsibilities of sprocket though? Worth thinking about and maybe starting a discussion in the sprocket channel

data_structures/read_group.wdl
a-frantz: looks good now 👍

data_structures/read_group.wdl
```diff
@@ -95,10 +85,18 @@ workflow read_group_to_string {
...
     output {
-        String validated_read_group = inner_read_group_to_string.stringified_read_group
+        String validated_read_group = select_first([
+            inner_read_group_to_string.stringified_read_group,
+            "",
+        ])
+        Array[String] validated_read_group_array = select_first([
+            inner_read_group_to_string.read_group_array,
+            [],
+        ])
```
a-frantz: can we bubble the `None`s all the way here? With WDL it is still awkward to use these optional values, but I prefer the explicitness of only one of these being defined at a time. It will cause a loud failure if someone "unwraps" the wrong one by accident, rather than this where the `""` or `[]` will get accepted silently.

a-frantz: worth pointing out if you haven't seen - https://github.com/openwdl/wdl/issues/551

Seems that WDL 1.4 will get an `!` operator for unwraps

docker/ngsep/Dockerfile
a-frantz: nothing off the shelf we could find for this one?

test/fixtures/mutect2/README.md
```diff
@@ -0,0 +1,29 @@
+# Mutect2 fixtures
+
+Inputs and intermediate artifacts used by the per-task tests in `tools/test/mutect2.yaml` for `tools/mutect2.wdl`. Files are downsampled to `chrY` and `chrM` to match the reference at `test/fixtures/reference/GRCh38.chrY_chrM.fa` and the BAMs at `test/fixtures/bams/test.bwa_aln_pe*.chrY_chrM.bam`.
+
+The following list is sorted alphabetically:
+
+## af-only-gnomad.hg38.chrY_chrM.vcf.gz
```
a-frantz: is this VCF suitable for general use? I don't see any reason in this description we can't plop it in the existing `fictures/vcfs` dir and use as a test in non-mutect2 tests (not suggesting you add it to existing tests, but if its in the repo a future PR might want to add it for an extra case)

tools/test/clair.yaml
a-frantz: all of these tests are marked slow, so none will run in CI. If they really are that slow, fair enough. But are you sure none of them are suitable for CI?

tools/test/deepvariant.yaml
a-frantz: same here, all slow

tools/test/manta.yaml
a-frantz: all slow, same Q

tools/test/strelka.yaml
a-frantz: noting all `slow` again

tools/bwa.wdl
```diff
@@ -58,7 +58,7 @@ task bwa_aln {
             n_cores=$(nproc)
         fi
         # -1 because samtools uses one more core than `--threads` specifies
-        (( samtools_cores = n_cores - 1 ))
+        (( samtools_cores = n_cores - 1 || 1 ))
```
a-frantz: is this because samtools errors with `--threads 0`? I would hope samtools would read that as "single thread" considering `--threads 1` is actually 2 threads (if the docs are to be believed, which well. maybe they aren't)

tools/bwamem2.wdl
```diff
@@ -0,0 +1,127 @@
...
+            -t ~{ncpu} \
+            -R "~{read_group}" \
+            -k ~{seed_length} \
+            -T ~{min_score} \
+            ~{if smart_pairing then "-p" else ""} \
+            ~{if skip_mate_rescue then "-S" else ""} \
+            bwa_db/"$PREFIX" \
+            "~{read_one_fastq_gz}" \
+            ~{if defined(read_two_fastq_gz) then "\"~{read_two_fastq_gz}\"" else ""} |
+        samtools view -b -o "~{output_name}" -
```
a-frantz: nit
```suggestion
            ~{if defined(read_two_fastq_gz) then "\"~{read_two_fastq_gz}\"" else ""} \
            | samtools view -b -o "~{output_name}" -
```
i find this easier to read (with the pipe on the left)

tools/bwamem2.wdl
```diff
@@ -0,0 +1,127 @@
...
+    >>>
+
+    output {
+        File alignments = output_name
+    }
+
+    requirements {
+        container: "ghcr.io/stjudecloud/bwamem2:2.3-0"
+        cpu: ncpu
+        memory: "~{4 * ncpu} GB"
```
a-frantz: should this have a `modify_memory_gb` arg? Maybe not necessary, but would be consistent with the rest of the repo

tools/bwamem2.wdl
```diff
@@ -0,0 +1,127 @@
...
+    >>>
+
+    output {
+        File reference_index = bwa_db_out_name
+    }
+
+    requirements {
+        container: "ghcr.io/stjudecloud/bwamem2:2.3-0"
+        cpu: 1
+        memory: "120 GB"
```
a-frantz: 😲 is all this needed?

tools/clair.wdl
```diff
@@ -0,0 +1,290 @@
...
+        File reference_fasta
+        File reference_fasta_index
+        File bam
+        File bam_index
+        File? bed_regions
+        File? vcf_candidates
+        Array[String] contigs = []
+        clair3_model model = clair3_model.ilmn
+        String output_dir = "clair3_output"
+        String platform = "ilmn"
```
a-frantz:
```suggestion
        platform platform = "ilmn"
```
?

tools/clair.wdl
```diff
@@ -0,0 +1,290 @@
+version 1.3
+
+enum platform {
```
a-frantz: Should these be Pascal?

tools/clair.wdl
```diff
@@ -0,0 +1,290 @@
...
+
+    requirements {
+        container: "hkubal/clair3:v2.0.3"
+        cpu: ncpu
+        memory: "64 GB"
+        disks: "~{disk_size_gb} GB"
+    }
+}
+
+task clairs {
```
a-frantz:
```suggestion
task clair_s {
```
? `clairs` looks like a plural of `clair` and is a bit confusing lol

tools/gatk4.wdl
```diff
@@ -1,5 +1,81 @@
-## [Homepage](https://software.broadinstitute.org/gatk)
-version 1.1
+version 1.3
+
+import "picard.wdl"
+
+enum ref_confidence {
```
a-frantz:
```suggestion
enum RefConfidence {
```

tools/gatk4.wdl
```diff
@@ -1,5 +1,81 @@
...
+        vcf_index: "Index file for the VCF."
+    }
+
+    String name
+    Boolean known
+    Boolean training
+    Boolean truth
+    Float prior
+    File vcf
+    File vcf_index
```
a-frantz: just noting that these will get localized in `resource_to_string`. IDK if that can be neatly avoided. These could become `basename()` Strings, but I figure that would be awkward and tricky to track along with the associated files?

tools/gatk4.wdl
```diff
@@ -482,11 +626,620 @@ task mark_duplicates_spark {
...
+        python3 > "interval_count.txt" <<CODE
+        import glob, os
+        # Works around a JES limitation where multiples files with the same name overwrite each other when globbed
+        intervals = sorted(glob.glob("out/*/*.interval_list"))
+        for i, interval in enumerate(intervals):
+            (directory, filename) = os.path.split(interval)
+            newName = os.path.join(directory, str(i + 1) + "." + filename)
+            os.rename(interval, newName)
+        print(len(intervals))
+        CODE
```
a-frantz:
```suggestion
        python3 > "interval_count.txt" <<PYTHON
        import glob, os
        # Works around a JES limitation where multiples files with the same name overwrite each other when globbed
        intervals = sorted(glob.glob("out/*/*.interval_list"))
        for i, interval in enumerate(intervals):
            (directory, filename) = os.path.split(interval)
            newName = os.path.join(directory, str(i + 1) + "." + filename)
            os.rename(interval, newName)
        print(len(intervals))
        PYTHON
```
I think using `PYTHON` constants will cause this to get properly syntax highlighted

tools/samtools.wdl
```diff
@@ -1339,3 +1293,172 @@ task markdup {
...
+    runtime {
+        cpu: ncpu
+        memory: "4 GB"
+        disks: "~{disk_size_gb} GB"
+        container: "quay.io/biocontainers/samtools:1.19.2--h50ea8bc_0"
+        maxRetries: 2
+    }
+}
+
+task sort {
```
a-frantz: we don't already have this one?

a-frantz (commented):
phew 😮‍💨 a lot of code to go through. Still skimmed a good chunk of the code, but wanted to get another round of comments out. This asks a lot of questions again.

docker/ngsep/Dockerfile
adthrasher: Biocontainers has a single old tag (4.0.1: https://quay.io/repository/biocontainers/ngsep?tab=tags) and the project, itself, just provides a Dockerfile (https://github.com/NGSEP/NGSEPcore/blob/master/Dockerfile).

I think this is another tool we can consider just dropping. I only implemented the `SingleSampleVariantsDetector` method, but it is an entire Java framework for next-gen sequencing data.

test/fixtures/mutect2/README.md
adthrasher: It's [gnomAD](https://gnomad.broadinstitute.org/) stripped to just the allele frequency field for hg38. I further stripped it down to `chrY` and `chrM`.  In theory another tool could try to use it, but I think that would be a bad idea. It's a custom Broad reference file used for this specific pipeline


> https://gatk.broadinstitute.org/hc/en-us/articles/360050722212-FAQ-for-Mutect2: 
>Finally, "af-only-gnomad.hg38.vcf" is a copy of the gnomAD VCF stripped of all unnecessary INFO fields.

tools/clair.wdl
adthrasher: We could, but the underlying tool uses snake case for the variants. So it'd have to be something like: 
```wdl
enum platform[String] {
     OntR10DoradoSup4khz = "ont_r10_dorado_sup_4khz"
}
```

tools/bwa.wdl
adthrasher: I think this was a bad attempt to avoid the 0 being treated as an error (`((expr))` evaluating to 0 gets treated as exit code 1).

PR #315, opened 2026-05-18
feat: make FASTQ analysis optional

adthrasher:
This makes the FASTQ portion of the QC pipeline optional. The default remains to run, but this allows a user to disable the analysis that depends on FASTQ files.

~~Also contains an off-topic change to add exceptions for `EmptyOutputs` to tasks with no expected outputs, since that rule was added to Sprocket.~~ This is now in #317.

### DIFF
--- a/workflows/qc/quality-check-standard.wdl
+++ b/workflows/qc/quality-check-standard.wdl
@@ -109,6 +109,10 @@ workflow quality_check_standard {
             warning: "These files can be very large.",
         }
         use_all_cores: "Use all cores? Recommended for cloud environments."
+        run_fastq_analysis: {
+            description: "Create FASTQs from the input BAM and run FASTQ-level analyses?",
+            help: "If false, the pipeline skips SAMtools bam-to-fastq, fqlint, Kraken2, fastp, librarian, and comparative Kraken2. Also disables qualimap_rnaseq (requires a collated BAM from bam_to_fastq).",
+        }
         optical_distance: {
             description: "Maximum distance between read coordinates to consider them optical duplicates instead of library duplicates (e.g. PCR duplicates).",
             help: "If `mark_duplicates == false`, this parameter is ignored. If `0`, then _optical_ duplicate marking is disabled and only traditional duplicate marking will be performed. Suggested settings of 100 for unpatterned versions of the Illumina platform (e.g. HiSeq) or 2500 for patterned flowcell models (e.g. NovaSeq). Review the `mark_duplicates` task in `../../tools/picard.wdl` for more information.",
@@ -117,10 +121,6 @@ workflow quality_check_standard {
         subsample_n_reads: {
             description: "Only process a random sampling of approximately `n` reads. Any `n <= 0` for processing entire input.",
             warning: "Subsampling is done probabalistically so the exact number of reads in the output will have some variation.",
-        }
-        run_fastq_analysis: {
-            description: "Create FASTQs from the input BAM and run FASTQ-level analyses?",
-            help: "If false, the pipeline skips SAMtools bam-to-fastq, fqlint, Kraken2, fastp, librarian, and comparative Kraken2. Also disables qualimap_rnaseq (requires a collated BAM from bam_to_fastq).",
         }
     }
 
@@ -155,9 +155,9 @@ workflow quality_check_standard {
         Boolean store_kraken_sequences = false
         Boolean output_intermediate_files = false
         Boolean use_all_cores = false
+        Boolean run_fastq_analysis = true
         Int optical_distance = 0
         Int subsample_n_reads = -1
-        Boolean run_fastq_analysis = true
     }
 
     call parse_input { input:
--- a/workflows/qc/quality-check-standard.wdl
+++ b/workflows/qc/quality-check-standard.wdl
@@ -118,6 +118,10 @@ workflow quality_check_standard {
             description: "Only process a random sampling of approximately `n` reads. Any `n <= 0` for processing entire input.",
             warning: "Subsampling is done probabalistically so the exact number of reads in the output will have some variation.",
         }
+        run_fastq_analysis: {
+            description: "Create FASTQs from the input BAM and run FASTQ-level analyses?",
+            help: "If false, the pipeline skips SAMtools bam-to-fastq, fqlint, Kraken2, fastp, librarian, and comparative Kraken2. Also disables qualimap_rnaseq (requires a collated BAM from bam_to_fastq).",
+        }
     }
 
     input {
@@ -153,6 +157,7 @@ workflow quality_check_standard {
         Boolean use_all_cores = false
         Int optical_distance = 0
         Int subsample_n_reads = -1
+        Boolean run_fastq_analysis = true
     }
 
     call parse_input { input:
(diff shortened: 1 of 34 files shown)


workflows/qc/quality-check-standard.wdl
```diff
@@ -153,6 +157,7 @@ workflow quality_check_standard {
         Boolean use_all_cores = false
         Int optical_distance = 0
         Int subsample_n_reads = -1
+        Boolean run_fastq_analysis = true
```
adthrasher: I expected this to fail linting, but it didn't.

adthrasher:
Fix for GATK4 tests is in #282

workflows/qc/quality-check-standard.wdl
```diff
@@ -109,6 +109,10 @@ workflow quality_check_standard {
             warning: "These files can be very large.",
         }
         use_all_cores: "Use all cores? Recommended for cloud environments."
+        run_fastq_analysis: {
+            description: "Create FASTQs from the input BAM and run FASTQ-level analyses?",
+            help: "If false, the pipeline skips SAMtools bam-to-fastq, fqlint, Kraken2, fastp, librarian, and comparative Kraken2. Also disables qualimap_rnaseq (requires a collated BAM from bam_to_fastq).",
```
a-frantz: I think the analyses canceled (Kraken, fastp, librarian, and qualimap rnaseq) should receive more emphasis than bam-to-fastq and fqlint. The later are just implementation details. Considering how fast and light fqlint is, I'm not even sure it needs to be mentioned. The heavy computation this skips is bam-to-fastq.

adthrasher: Fair enough. I just listed them in the same order as the workflow (I think).

workflows/qc/quality-check-standard.wdl
```diff
@@ -254,109 +259,111 @@ workflow quality_check_standard {
...
+            }
+        }
+
+        if (run_comparative_kraken) {
+            call samtools.bam_to_fastq as alt_filtered_fastq after quickcheck after comparative_kraken_filter_validator {
+                input:
+                bam = post_subsample_bam,
+                bitwise_filter = comparative_filter,
+                prefix = post_subsample_prefix + ".alt_filtered",
+                # matches default but prevents user from overriding
```
a-frantz:
```suggestion
```

workflows/qc/quality-check-standard.wdl
```diff
@@ -254,109 +259,111 @@ workflow quality_check_standard {
...
+                # If the user wants a collated BAM, they should save the one
+                # from the first bam_to_fastq call.
+                retain_collated_bam = false,
+                # matches default but prevents user from overriding
```
a-frantz:
```suggestion
```

workflows/qc/quality-check-standard.wdl
```diff
@@ -254,109 +259,111 @@ workflow quality_check_standard {
...
+                # Since the only output here is FASTQs, we can disable fast mode.
+                # This discards secondary and supplementary alignments, which should not
+                # be converted to FASTQs. (Is that true?)
```
a-frantz:
```suggestion
                # be converted to FASTQs.
```
yes, it is true AFAIK

workflows/qc/quality-check-standard.wdl
```diff
@@ -254,109 +259,111 @@ workflow quality_check_standard {
...
+                fast_mode = true,
+                paired_end = true,  # matches default but prevents user from overriding
```
a-frantz:
```suggestion
                paired_end = true,  # INVARIANT: this workflow only supports PE data
```

workflows/qc/quality-check-standard.wdl
```diff
@@ -392,18 +399,20 @@ workflow quality_check_standard {
...
+                    bam_to_fastq.collated_bam,
+                    "undefined",
+                ]),
+                prefix = post_subsample_prefix + ".qualimap_rnaseq_results",
+                gtf = select_first([
+                    gtf,
+                    "undefined",
+                ]),
+                name_sorted = true,
+                paired_end = true,  # matches default but prevents user from overriding
```
a-frantz:
```suggestion
                paired_end = true,  # INVARIANT: this workflow only supports PE data
```

a-frantz (changes requested):
as discussed over DM, the comments I request be deleted miss the mark. The fact they match the task default is a coincidence; the inputs have reason to be specified here at the workflow level.

Even though specifying here is technically redundant with relying on the default task values, I'd rather explicitly specify them here with the reasoning. Respecifying here at the workflow level is also a safeguard against the default switching out from under us

workflows/qc/quality-check-standard.wdl
```diff
@@ -254,109 +259,111 @@ workflow quality_check_standard {
...
+                input:
+                bam = post_subsample_bam,
+                bitwise_filter = comparative_filter,
+                prefix = post_subsample_prefix + ".alt_filtered",
+                # matches default but prevents user from overriding
+                # If the user wants a collated BAM, they should save the one
+                # from the first bam_to_fastq call.
+                retain_collated_bam = false,
+                # matches default but prevents user from overriding
+                # Since the only output here is FASTQs, we can disable fast mode.
```
a-frantz:
```suggestion
                # Since the only output here is FASTQs, we can enable fast mode.
```

workflows/qc/quality-check-standard.wdl
```diff
@@ -109,6 +109,10 @@ workflow quality_check_standard {
             warning: "These files can be very large.",
         }
         use_all_cores: "Use all cores? Recommended for cloud environments."
+        run_fastq_analysis: {
+            description: "Create FASTQs from the input BAM and run FASTQ-level analyses?",
+            help: "If false, the pipeline skips FASTQ generation and subsequent FASTQ checks (Kraken2, fastp, librarian, comparative Kraken2, and fqlint). Also disables qualimap_rnaseq (requires a collated BAM from bam_to_fastq).",
```
a-frantz: it would have been cool to see resource utilization comparisons of this on/off, but that isn't easy to gather in the current WDL landscape.

Do you think it's worth mentioning here that this should shave off a big chunk of processing? That would be obvious to anyone who is familiar with the domain, but (a workflow like) QC standard also makes sense to run in the case of someone new to the field. It might read awkwardly without any numbers, but how about:
```suggestion
            help: "If false, the pipeline skips FASTQ generation and subsequent FASTQ checks (Kraken2, fastp, librarian, comparative Kraken2, and fqlint). Also disables qualimap_rnaseq (requires a collated BAM from bam_to_fastq). The steps disabled by this option can have heavy resource utilization.",
```
In a similar vein, would it be appropriate here to comment on _why_ one might want to disable FASTQ analysis in a QC context (aside from cost)? Something about the difference in QCing the sequencing step VS the alignment step (although I guess that division isn't very easy to draw, so maybe too large a can of worms to open here).

a-frantz (dismissed):
a take it or leave it doc suggestion, and another less confident thought I had about potential QC docs (tangent from this PR)

a-frantz (approved):

PR #322, opened 2026-08-03
chore: updates to satisfy ruff and sprocket lint

adthrasher:
Updates to satisfy the new version of Ruff.

Adds `set` to each command block to satisfy `sprocket lint`.

### DIFF
--- a/tools/gatk4.wdl
+++ b/tools/gatk4.wdl
@@ -118,6 +118,8 @@ task base_recalibrator {
     Int java_heap_size = ceil(memory_gb * 0.9)
 
     command <<<
+        set -euo pipefail
+
         # shellcheck disable=SC2102
         gatk \
             --java-options \
@@ -271,6 +273,8 @@ task haplotype_caller {
     Int java_heap_size = ceil(memory_gb * 0.9)
 
     command <<<
+        set -euo pipefail
+
         gatk \
             --java-options \
                 "-Xms6000m -Xmx~{java_heap_size}g -XX:GCTimeLimit=50 -XX:GCHeapFreeLimit=10" \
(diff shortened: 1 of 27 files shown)


adthrasher:
The Sprocket lint appears to be failing for the new `set` rule. We have historically disagreed with putting `set` in simple command blocks, but maybe we should reevaluate.

As for the image bumps, this PR doesn't actually change anything, so I'm inclined not to bump the images and re-release. I think we can just update the script formatting and let that change trickle out as we make substantive changes to the images.

tools/gatk4.wdl
```diff
@@ -189,7 +191,7 @@ task apply_bqsr {
         # shellcheck disable=SC2102
         gatk \
             --java-options \
-                "-XX:GCTimeLimit=50 -XX:GCHeapFreeLimit=10 -Xms3000m -Xmx~{java_heap_size}g" \
+                "-XX:GCTimeLimit=50 -XX:GCHeapFreeLimit=10 -Xms3000m -Xmx~{java_heap_size}g -XX:-UseContainerSupport" \
```
a-frantz, line 194: what's this flag?

a-frantz (approved):
noting that this PR does not include any CHANGELOG entries, and I think that's fine. Just one question about a java flag, otherwise LGTM

tools/gatk4.wdl
adthrasher: I shared this in Slack previously (while you were out): https://stjude.slack.com/archives/C025KBTS9JQ/p1778606767391569

Essentially, it looks like, by default, the JVM detects when it is running in a container and can override your heap space settings. This seems to only really be an issue when using small heap spaces, such as during our CI tests.

adthrasher: It was previously buried in #282, but I promoted it to this PR since it resolves the CI test failures in `gatk4`.

